PR #69, opened 2017-12-05
Adding 360 Lidars support (similar to Velodyne HDL-32E or VLP-16)

parilo:
You may add any number of lidars to the car. Lidars have following settings with specified default values:

\# Number of lasers
Channels = 32

\# Measure distance
Range = 5000

\# Points generated by all lasers per second
PointsPerSecond = 100000

\# Lidar rotation frequency
RotationFrequency = 10

\# Upper laser angle, counts from horizontal,
\# positive values means above horizontal line
UpperFovLimit = 10

\# Lower laser angle, counts from horizontal,
\# negative values means under horizontal line
LowerFovLimit = -30

\# wether to show debug points of laser hits in simulator
ShowDebugPoints = False

\# Position relative to the player.
LidarPositionX = 0
LidarPositionY = 0
LidarPositionZ = 250

\# Rotation relative to the player.
LidarRotationPitch = 0
LidarRotationRoll = 0
LidarRotationYaw = 0

Default values choosed close to Velodyne HDL-32E.
Also one lidar added to ./manual_control.py. You may see bird eye view of lidar points in upper left corner.

Lidar was made using UWorld::LineTraceSingleByChannel method. Serialization of lidar measurement made similarly to camera images.


parilo:
Lidar measurements send in form of packet. Packet is points captured by lidar during last tick. So it is points from segment covered in last tick. On low FPS packets may include points from 360 degrees.

nsubiron:
Hi @parilo,
The code looks great! seems to work as expected. Just noticed one thing, when a laser shot doesn't hit anything you leave the point at {0,0,0}, (you can see the white dot at the centre of the image). Do you anticipate any problem by not adding the misses to the list of points? At `ALidar::ReadPoints` we can skip the points if the ray-trace failed.

parilo:
Hi @nsubiron,
If misses will not be added we will lose organized point cloud (I mean structure of whole points packet). For some algorithms it may be reasonable to have organized one. Also serialization and deserialization code will need to be rewritten because it assumes equals point counts in each channels. I can rewrite code to unorganized point cloud (remove misses) if it is needed. I leaved misses with {0, 0, 0} mainly for possible future use of organized point cloud.

PR #100, opened 2017-12-20
Point cloud from depth

### LINKED ISSUE #86: Extract point cloud from depth map image
In the client side (Python), add the methods necessary to create a point cloud from the depth image received every frame.

marcgpuig:
Fix #86.

Example client that converts a depth image into a point cloud using the intrinsic and extrinsic matrices.
The 3D point cloud is written in [PLY](http://paulbourke.net/dataformats/ply/) format.

77ed4f72ee203b2ee061c1e98086d423


PythonClient/carla/transform.py
```diff
@@ -0,0 +1,153 @@
...
+                    args[0].location.z),
+                Rotation(
+                    args[0].rotation.pitch,
+                    args[0].rotation.yaw,
+                    args[0].rotation.roll)
+            ]
+        self.matrix = numpy.matrix(numpy.identity(4))
+        self.set(*args, **kwargs)
+
+    def set(self, *args):
```
nsubiron: Would something like this be cleaner?
```py
    def set(self, *args):
        """Build the transform matrix given by a Translation, a Rotation and a
        Scale.
        """

        class Pair(object):
            def __init__(self, value_type):
                self.value_type = value_type
                self._value = None

            def set(self, value):
                if not isinstance(value, self.value_type):
                    return False
                if self._value is not None:
                    message = 'cannot make Transform with more than one %r'
                    raise ValueError(message % self.value_type.__name__)
                self._value = value
                return True

            def get(self):
                return self.value_type() if self._value is None else self._value

        # Make sure we get only one of each and not something else.
        pairs = [Pair(Translation), Pair(Rotation), Pair(Scale)]
        for item in args:
            if not any(pair.set(item) for pair in pairs):
                raise TypeError('invalid argument type %r' % type(item).__name__)

        translation, rotation, scale = [pair.get() for pair in pairs]

        # ...
```

PythonClient/carla/transform.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class Transform(object):
+    """A 3D transformation.
+
+    The transformation is applied in the order: scale, rotation, translation.
+    """
+
+    @staticmethod
+    def _from_protobuf_transform(transform):
```
nsubiron: Function no longer needed.

PythonClient/point_cloud_example.py
```diff
@@ -0,0 +1,196 @@
...
+from carla.sensor import Camera
+from carla.settings import CarlaSettings
+from carla.tcp import TCPConnectionError
+from carla.util import print_over_same_line, StopWatch
+from carla.image_converter import depth_to_local_point_cloud, to_rgb_array
+from carla.transform import Transform, Translation, Rotation, Scale
+
+
+def run_carla_client(host, port, far):
+    # Here we will run 3 episodes with 300 frames each.
```
nsubiron: Here we will run 1 episode with 300 frames.

PR #104, opened 2017-12-22
 Travis-ci support

nsubiron:
Adding pylint continuous integration for the "carla" Python API and the examples.

### DIFF
--- a/PythonClient/carla/planner/map.py
+++ b/PythonClient/carla/planner/map.py
@@ -89,12 +89,13 @@ class CarlaMap(object):
         return np.fliplr(self.map_image)
 
     def get_map_lanes(self, height=None):
-        if size is not None:
-            img = Image.fromarray(self.map_image_lanes.astype(np.uint8))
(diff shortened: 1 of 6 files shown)


PythonClient/carla/planner/map.py
```diff
@@ -89,12 +89,13 @@ def get_map(self, height=None):
...
-            img.load()
-            return np.fliplr(np.asarray(img, dtype="int32"))
-        return np.fliplr(self.map_image_lanes)
+        # if size is not None:
+        #     img = Image.fromarray(self.map_image_lanes.astype(np.uint8))
+        #     img = img.resize((size[1], size[0]), Image.ANTIALIAS)
+        #     img.load()
+        #     return np.fliplr(np.asarray(img, dtype="int32"))
+        # return np.fliplr(self.map_image_lanes)
+        raise NotImplementedError
```
nsubiron: This should definitively be fixed! but we can wait to do it in the benchmark branch

PR #105, opened 2017-12-22
Adding 360 Lidars support

### LINKED ISSUE #68: Lidar support
Hi, All.
I've added Lidar support and now cleaning my code and fixing minor problems. Soon I will be able to make cross repo pull request. Will my code be useful for Carla project?

What is supported:
- adding any number of lidars through configuration ini file
- for individual lidars set number of lasers and vertical angles, number of points per second, position, orientation
- currently lidar returns XYZ coordinates in lidar frame
- points are serialized and available in python client and visualized in ./carla_manual_control.py

nsubiron:
Adding the changes of #69.

Resolve #68.

- - -
peek 2018-01-19 17-28

You may add any number of lidars to the car. Lidars have following settings with specified default values:

```
# Number of lasers
Channels = 32

# Measure distance
Range = 5000

# Points generated by all lasers per second
PointsPerSecond = 100000

# Lidar rotation frequency
RotationFrequency = 10

# Upper laser angle, counts from horizontal,
# positive values means above horizontal line
UpperFovLimit = 10

# Lower laser angle, counts from horizontal,
# negative values means under horizontal line
LowerFovLimit = -30

# wether to show debug points of laser hits in simulator
ShowDebugPoints = False

# Position relative to the player.
LidarPositionX = 0
LidarPositionY = 0
LidarPositionZ = 250

# Rotation relative to the player.
LidarRotationPitch = 0
LidarRotationRoll = 0
LidarRotationYaw = 0
```

Default values choosed close to Velodyne HDL-32E.
Also one lidar added to `manual_control.py`. You may see bird eye view of lidar points in upper left corner.

Lidar was made using UWorld::LineTraceSingleByChannel method. Serialization of lidar measurement made similarly to camera images.


nsubiron:
Hi @simmates,

Some of the Lidar code has been refactored in #248, there you can also find the available settings for this Lidar implementation.

If you want to observe the rolling shutter effect you can do so by using a fixed frame rate and increasing the fps, sort of slow motion effect

    ./CarlaUE4.sh -benchmark -fps=500

simmates:
@nsubiron thanks. This is some really important work and will be highly appreciated.
Instead of using higher frame rates: do you know if it is possible to use the substepping system instead and do raycasting in between frames? I don't know if the world will get updated (players positions and rotations) in the physics ticks but could be interesting

nsubiron:
@simmates After investigating a bit it seems to be possible to launch a lidar update on every physics substep. It would require however some changes in the current implementation. I have created a new feature request (#266) for this.

PR #240, opened 2018-02-21
Add route commands

Kuldaen:
#### Description
We have added a RouteCommand enum into the measurements protobuf definition. This exposes the Autopilot's decisions at intersections

  * **Platform(s):** Linux
  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** 4.17, 4.18


nsubiron:
Dear @Kuldaen,

Thanks for contributing to CARLA and sorry for taking so long to reply.

We have been discussing your changes in this pull-request and we have decided not to merge these changes as we do not want to expose this functionality to the client as it is right now.

We plan to move the autopilot functionality as much as possible to the client side, and make it all controllable from the Python API. The current implementation is somehow temporary and we would prefer not to expose it in the server-client protocol.

Thanks again,
The CARLA team

PR #273, opened 2018-03-07
Convert measurements to SI units

### LINKED ISSUE #132: What units are measurements in client given in?
This doesn't seem to be documented anywhere - what units are the acceleration, position, etc, given in when you access player_measurements in the client?

nsubiron:
#### Description

All the units have been converted to SI units

```
                 before         now
---------------------------------------
locations        cm         --> m
speed            km/h       --> m/s
acceleration     (km/h)/s   --> m/s^2
collisions       kg*cm/s    --> kg*m/s
angles           degrees    --> degrees
```

Fixes #132.

#### Possible Drawbacks

This is actually a tricky change, all the client code dependent on units should be carefully checked.


nsubiron:
@felipecode Please cross check that this is not breaking any benchmark code.

PR #299, opened 2018-03-22
Fix race condition

### LINKED ISSUE #297: Race condition between client and server causes disconnect on reset
There is a race condition between the client and server that causes the server 
to reset the connection.

```
    simulator      
    ---------      

read new episode request on port 2000
if (error)
  reset the server
                                  <-- client closes port 2001 and 2002 and
                                      requests new episode on 2000.
send measurements on port 2001    <-- send measurements fails, client closed 2001,
if (error)                            server assumes the client is down.
  reset the server                <-- server is destroyed, diconnects 2000,
                                      client request fails with connection closed.
  
read control on port 2002
if (error)
  reset the server
```

This has been there for a long time, but it didn't show up until recent changes. Recently
we added an extra step in between that gives extra time to the client to win the race.

Possible solution: Read and write errors on ports 2001 and 2002 should not make
the server drop the connection. It is safe to ignore them, if the client really
disconnected it will be detected on the next frame when reading the new episode
request.

nsubiron:
#### Description

Fix the race condition between the client and server that causes the server
to reset the connection.

Now disconnect errors in ports 2001 and 2002 are ignored.

Fixes #297.

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/ServerGameController.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/ServerGameController.cpp
@@ -121,17 +121,18 @@ void FServerGameController::Tick(float DeltaSeconds)
 
   // Send measurements.
   {
-	if(CarlaSettings->bSynchronousMode)
-	{
-	  FlushRenderingCommands();
-	}
+    if (CarlaSettings->bSynchronousMode)
+    {
+      FlushRenderingCommands();
+    }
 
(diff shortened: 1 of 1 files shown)


YashBansod:
I was wondering why this was happening. I built carla from sources and the connection kept disconnecting on start of a new episode (happened almost 1 in 3 times)

nsubiron:
When adding the asynchronous scene captures we put the fence that waits for the images right before sending the measurements, that gives extra time to the client to disconnect during the wait, thus more likely to happen. It's explained with more detail in #297.

PR #316, opened 2018-03-28
Adding the initial version of the example RL environment

### LINKED ISSUE #300: OpenAI-Gym-compliant environment
I'm preparing an OpenAI-Gym-compliant RL environment for Carla. This should enable easier benchmarking of Reinforcement Learning algorithms on Carla. I will be documenting my progress in this issue and create a pull request once it's ready.

tgrel:
### Description

Adding the initial version of the example RL environment for CARLA. This is a simple wrapper around the existing python client that lets you make steps in the environment and receive states. Also a simple test with making a video of a random agent is included. The environment assumes the server is already running and takes its hostname and port number as parameters. I placed the env module inside the PythonClient/carla package (feel free to suggest other locations).

This is the initial version. I'm waiting for your feedback.

Fixes #300 

I tested it by running the server locally on a Ubuntu-16.04 machine and connecting an environment to it.

#### Possible Drawbacks

Users will want to change the reward function, state and possible actions so that it fits their experiments, so this is by no means a complete implementation. It's rather a starting point for modifications. Otherwise we could make all these things configurable but covering all the cases will require a lot of effort.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/316)

### DIFF
--- a/PythonClient/.pylintrc
+++ b/PythonClient/.pylintrc
@@ -1,4 +1,4 @@
 [TYPECHECK]
 ignore=carla_server_pb2.py
-ignored-modules=ConfigParser,numpy,numpy.random,pygame,shutil
+ignored-modules=ConfigParser,numpy,numpy.random,pygame,shutil,cv2
 ignored-classes=_socketobject,EpisodeReady,SceneDescription,Sensor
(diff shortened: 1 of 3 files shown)


marcgpuig:
Try adding the used modules to [`requirements.txt`](https://github.com/carla-simulator/carla/blob/master/PythonClient/requirements.txt) so Travis can check correctly your code.

    gym
    opencv-python

PR #348, opened 2018-04-13
Add "frame number" to measurements and sensor data

### LINKED ISSUE #332: Timestamps for camera frames
I am running Carla 0.8 release on Ubuntu 16.04, rendering on a TitanXp. I am trying to get data out as fast as possible, so I am using asynchronous mode.  I see that I can get the internal game timestamps for the measurements with `measurements.game_timestamp`, but I don't see a corresponding field for the cameras.  The docs state the cameras may be delayed up to two frames from the measurements, so I would like to know the game timestamp when the frame was rendered.  Did I miss this, or is it something you could add in the future?

nsubiron:
#### Description

Adds a _frame number_ field to the measurements message and the sensor data packages. This way it is possible to match sensor data received asynchronously with their corresponding measurements.

```py
measurements, sensor_data = client.read_data()

measurements.frame_number

for name, data in sensor_data.items():
    data.frame_number
```

The `frame_number` comes from Unreal's `GFrameCounter` and it is not restarted on each episode. 

Fixes #332.

  * **Python version(s):** 3.5 and 2.7

#### Drawbacks

The codification of the sensor data changes thus is not compatible with previous versions.


marcgpuig:
Now works fine for me

Issue #286, opened 2018-03-20
How to disable rendering in CARLA server?


GokulNC:
So when the CARLA server is run by using the script `CarlaUE4.sh` to launch it, is there any way to disable the rendering of the game window?

I use a camera in the simulator to get the current FoV from that camera. So just reading the image by `CarlaClient.read_data()` is enough for me to process things. I feel that I can increase my [FPS](https://github.com/carla-simulator/carla/issues/10) by reducing the processing overhead to **render** the game using Unreal Engine.

I tried using the flag `-carla-no-hud` as suggested in the [docs](http://carla.readthedocs.io/en/latest/how_to_run/), but still the rendering is taking place. 

So is there anyway to do this? Or it's not just possible or it doesn't matter not to render?
Thanks :)

marcgpuig:
Hi @GokulNC,

First of all `-carla-no-hud` disables the left-bottom text output from the server:

![cbb50b6e874c0e3a3ec15cc9a5cc4c36](https://user-images.githubusercontent.com/15848712/37659609-bb3b0fd4-2c50-11e8-9d88-c66cf69175d2.png)

Currently is no possible to disable the render and I have the feeling that it will doesn't matter anyways, because regardless you see or not the render, it must be done anyway.

GokulNC:
Thanks @marcgpuig .

Eventhough the computations need to occur, I feel rendering it at a certain FPS (say 15) is quite resource-intensive.

Though there is no way to disable rendering as of now, can this issue be considered as a feature-request?

I will close this issue if this feature is implemented, or if there's an strong assertion that disabling rendering doesn't matter much.

Thanks :)

marcgpuig:
Please try adding these commands:

    -windowed -ResX=8 -ResY=8

So it will run on a `8x8` pixels window. It's a workaround but I hope you can win some extra fps. 😉 

Another solution is work without an actual screen, see our [documentation](http://carla.readthedocs.io/en/latest/carla_headless/).

For sure I'm going to add the *feature-request* tag because it is something that can improve the performance.
But we are currently working on other more critical optimizations, so do not expect big focus on this one.

Since it's more an *Unreal Engine issue*, I'm adding the *good first issue* too, so the community can make a good-and-easy contribution 😄 

Regards!

simmates:
Is it also possible to disable the HUD in the Editor?

simmates:
Ok it is possible in the console with showHUD 0

juaxix:
[...]

PR #469, opened 2018-06-04
Non-rendering mode

juaxix:
Changes to enable or disable rendering of the world (Unreal render) and CARLA HUDs (Unreal HUD rendering) using a property of the server settings.

The changes involve:

  - [x] Clients can use the DisableRendering flag
  - [x] Added the option in : Carla.Settings.ini -> [CARLA/Server] -> DisableRendering 
  - [x] Added the argument for runtime : "-disable-rendering"
  - Tests needed

#### Example
Using the code for manual_control.py, with this modification:

```py
def make_carla_settings(args):
    """Make a CarlaSettings object with the settings we need."""
    settings = CarlaSettings()
    settings.set(
        SynchronousMode=True,
        SendNonPlayerAgentsInfo=True,
        NumberOfVehicles=15,
        NumberOfPedestrians=30,
        WeatherId=random.choice([1, 3, 7, 8, 14]),
        QualityLevel=args.quality_level,
        DisableRendering=True)
    settings.randomize_seeds()
    if args.lidar:
        lidar = sensor.Lidar('Lidar32')
        lidar.set_position(0, 0, 2.5)
        lidar.set_rotation(0, 0, 0)
        lidar.set(
            Channels=32,
            Range=50,
            PointsPerSecond=100000,
            RotationFrequency=10,
            UpperFovLimit=10,
            LowerFovLimit=-30)
        settings.add_sensor(lidar)
    return settings
```

we are removing all the sensors except for the lidar. We can launch the client with a command like this:

> PythonClient$ python manual_control.py -a -l

Screenshot of the result :
example_no_render

  * **Platform(s): Windows and Linux
  * **Python version(s): ** 2.6, 3
  * **Unreal Engine version(s):** 4.18.3

#### Possible Drawbacks

You may want to draw huds?

child of #286.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/469)

### DIFF
--- a/Docs/Example.CarlaSettings.ini
+++ b/Docs/Example.CarlaSettings.ini
@@ -24,7 +24,7 @@ SynchronousMode=true
 ; vehicles, pedestrians and traffic signs. Disabled by default to improve
 ; performance.
 SendNonPlayerAgentsInfo=false
-; Flag to enable or disable the world rendering.
+; Flag to enable or disable the world and hud rendering.
 ; Disabled by default.
 DisableRendering = false
 
--- a/Docs/Example.CarlaSettings.ini
+++ b/Docs/Example.CarlaSettings.ini
@@ -24,6 +24,9 @@ SynchronousMode=true
 ; vehicles, pedestrians and traffic signs. Disabled by default to improve
 ; performance.
 SendNonPlayerAgentsInfo=false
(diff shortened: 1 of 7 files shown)


nsubiron:
Rebased specific commits onto master to remove dependencies with #445.

PR #472, opened 2018-06-04
 Adding unit/acceptance test for map changing

felipecode:
Hello @juaxix,
Thanks for the very useful feature of map changing on client side !
First of all, I added on carla/settings.py the possibility of changing the map name
on the carla settings code. Quite trivial.

Now about the tests.
At first it seemed to be working correctly. I was able to start CARLA on Town01 and execute
tests on Town02. However I decided to make a test to stress this map change since it is a very
important functionality.
For that, I created the acceptance_test called " test_map_changing.py"
Basically, what it does is changing maps from Town01 to Town02 and then Town02 to Town01 ...
repeating this process 8 times. At each time it asserts if the map name received is the same
as the map loaded by the client. 
What happens is the following:

It starts on town 01 , receives town01  --- Correct
It starts on town 02 , receives town02  --- Correct
Then something weird happens:
It starts on town 01 , receives town02 and the connection crashes!

Please accept the pull request and try to reproduce the problem by running the test_map_changing.py
script with nosetests3. 

Thanks again.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/472)

### DIFF
--- a/PythonClient/carla/settings.py
+++ b/PythonClient/carla/settings.py
@@ -39,6 +39,7 @@ class CarlaSettings(object):
         # [CARLA/QualitySettings]
         self.QualityLevel = 'Epic'
         # [CARLA/LevelSettings]
+        self.MapName = 'Town02'
         self.PlayerVehicle = None
         self.NumberOfVehicles = 20
         self.NumberOfPedestrians = 30
@@ -97,6 +98,7 @@ class CarlaSettings(object):
         add_section(S_QUALITY, self, [
             'QualityLevel'])
         add_section(S_LEVEL, self, [
+            'MapName',
             'NumberOfVehicles',
             'NumberOfPedestrians',
             'WeatherId',
--- a/PythonClient/test/acceptance_tests/test_map_changing.py
+++ b/PythonClient/test/acceptance_tests/test_map_changing.py
@@ -1,0 +1,92 @@
+"""
+For now we use the unit tests for this part
+Basically the difference between this folder and the other, the unit tests folder
+is that the acceptance tests is carla based
+
+TODO: Change the name
+
+"""
+
+import unittest
+import random
+import time
+
+from carla.client import make_carla_client
+from carla.sensor import Camera, Lidar
+from carla.settings import CarlaSettings
+
+from carla.util import print_over_same_line
+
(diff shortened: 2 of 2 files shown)


juaxix:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved.

---

*[PythonClient/carla/settings.py, line 42 at r1](https://reviewable.io/reviews/carla-simulator/carla/472#-LEDvV2RC0xdnxTlYdyL:-LEDvV2RC0xdnxTlYdyM:bdsvdc3) ([raw file](https://github.com/carla-simulator/carla/blob/f1ace4da7bfbc6e7c09cd9f898b9a4dd2676094a/PythonClient/carla/settings.py#L42)):*
> ```Python
>         self.QualityLevel = 'Epic'
>         # [CARLA/LevelSettings]
>         self.MapName = 'Town02'
> ```

With this change we should also change the default map in the unreal project settings in Map & Modes section, so you are not loading the Town01 in the first place. But in the future the first level should be the transition level, so...i think it will be ok for the long term

---


*Comments from [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/472#-:-LEDvnkLDPtVUozmzTJR:bnfp4nl)*

felipecode:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[PythonClient/carla/settings.py, line 42 at r1](https://reviewable.io/reviews/carla-simulator/carla/472#-LEDvV2RC0xdnxTlYdyL:-LEDx35T1_unsnFkA3XU:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/f1ace4da7bfbc6e7c09cd9f898b9a4dd2676094a/PythonClient/carla/settings.py#L42)):*
<details><summary><i>Previously, juaxix (juaxix) wrote…</i></summary><blockquote>

With this change we should also change the default map in the unreal project settings in Map & Modes section, so you are not loading the Town01 in the first place. But in the future the first level should be the transition level, so...i think it will be ok for the long term
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/472)*

PR #478, opened 2018-06-06
Change HUD: replace "FPS" by "Simulation Step"

nsubiron:
The fixed time-step mode keeps generating a lot of confusion. I think partly because of using the term FPS for both modes while they're not exactly the same (frames-per-simulation-second vs frames-per-real-second). While we cannot change the command line args because they're handled by UE4, at least we can change the HUD.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/478)


marcgpuig:
Works fine in Windows too

---

Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/478#-:-LEK5z1K7JoWqRaPQMp2:bmeelr3)*

Issue #496, opened 2018-06-13
Issues with two-wheeled vehicles


nsubiron:
Moving the known issues of #477 to its own issue.

Known issues:

- [ ] Control is still not perfect. Though the AI has almost no problem driving the bikes human control is still clumsy.
- [ ] Due to the physics constraint used to make the bikes ride straight, the slightest bump on the road makes the bike jump.
- [ ] Hand and Feet animation are not perfect, some feet might go through the floor and some hands might hover over the handler.
- [ ] Vehicle wheels bone axis must be aligned with world axis to work and most bikes handlers are not world aligned, this causes the wheel and the handler to behave differently while steering in some ocassions.

PR #509, opened 2018-06-15
Update assets for 0.8.4

### LINKED ISSUE #491: No collision info in the 0.8.3 version
Hi, it is so exciting to see a new version of Carla! I just downloaded it and tested on my machine, it runs much faster!

However, I found that there is no collision info on HUD when I run Carla alone. It always shows 0 even if I hit other vehicles or pedestrians. Then I tried with a carla server and `client_example.py`, and there is no collision info on HUD and in the terminal either.

Version: Carla 0.8.3 on Ubuntu 16.04

Alone

Server

-- Update--
It might be related to the vehicle. When I change the setting to another vehicle rather than the default Mustang, the collision info in shown on the HUD.

nsubiron:
Update assets with the latest changes for version 0.8.4.

  * Fixes #491 player vehicle missing collision info.
  * Fixes some issues with two-wheeled vehicles of #496.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/509)

### DIFF
--- a/Util/ContentVersions.txt
+++ b/Util/ContentVersions.txt
@@ -12,3 +12,4 @@
 0.8.1: 1su5wQH7KJphAcirMA8VEDLndGigQ9tzq
 0.8.2: 1llrBkZDvJmxXYh4r3jz0wuwz8Q8jRQeZ
 0.8.3: 1z2XKOk9LoKI-EhxcrpI31W5TeYo2-jjn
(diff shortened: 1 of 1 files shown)


marcgpuig:
Reviewed 1 of 1 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/509)*

PR #549, opened 2018-07-01
Wheels' speed

ManueleSigona:
#### Description

I needed to simulate a wheels' rotational speed sensor for a research project; then I modified all the necessary source files in order to achieve this.
The rotational speed is estimated using current and past wheels' position (in angle), the position is taken from PhysX model's function: GetRotationAngle(), which gives the position in degrees. It's then converted in radians, so that the wheels' rotational speed computed is in rad/s.
The four rotational speeds computed (one for each wheel) are accessible from CARLA client, they're stored in player measurements.
I've made this PR in case it could be of any use to anyone.

Fixes #  

  * **Platform(s):** Lenovo Thinkpad L470 with Ubuntu 18.04
  * **Python version(s):** 3.5.5

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/549)

### DIFF
--- a/Docs/measurements.md
+++ b/Docs/measurements.md
@@ -39,10 +39,10 @@ collision_other            | float       | kg*m/s | General collision intensity 
 intersection_otherlane     | float       |        | Percentage of the vehicle invading other lanes.
 intersection_offroad       | float       |        | Percentage of the vehicle off-road.
 autopilot_control          | Control     |        | Vehicle's autopilot control that would apply this frame.
-wheelFLOmega               | float     | rad/s  | Vehicle's front left wheel' rotational speed.
-wheelFROmega               | float     | rad/s  | Vehicle's front right wheel' rotational speed.
-wheelRLOmega               | float     | rad/s  | Vehicle's rear left wheel' rotational speed.
-wheelRROmega               | float     | rad/s  | Vehicle's rear right wheel' rotational speed.
+wheelFLOmega               | float       | rad/s  | Vehicle's front left wheel' rotational speed.
+wheelFROmega               | float       | rad/s  | Vehicle's front right wheel' rotational speed.
+wheelRLOmega               | float       | rad/s  | Vehicle's rear left wheel' rotational speed.
+wheelRROmega               | float       | rad/s  | Vehicle's rear right wheel' rotational speed.
 
 <h4>Transform</h4>
 
--- a/Docs/measurements.md
+++ b/Docs/measurements.md
@@ -39,6 +39,10 @@ collision_other            | float       | kg*m/s | General collision intensity 
 intersection_otherlane     | float       |        | Percentage of the vehicle invading other lanes.
 intersection_offroad       | float       |        | Percentage of the vehicle off-road.
 autopilot_control          | Control     |        | Vehicle's autopilot control that would apply this frame.
+wheelFLOmega               | float     | rad/s  | Vehicle's front left wheel' rotational speed.
+wheelFROmega               | float     | rad/s  | Vehicle's front right wheel' rotational speed.
+wheelRLOmega               | float     | rad/s  | Vehicle's rear left wheel' rotational speed.
+wheelRROmega               | float     | rad/s  | Vehicle's rear right wheel' rotational speed.
 
 <h4>Transform</h4>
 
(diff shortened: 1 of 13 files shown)


nsubiron:
Great, we could make use of this :) I'll make the review

nsubiron:
The main problem I see is that this only works for 4 wheeled vehicles, for now all the vehicles we have have 4 wheels (yup, even the bikes) but it will be much more robust if any number of wheels could be use, and better if we can filter out the extra wheels on the bikes. I think too it will result in simpler code with less methods repeated.

In the proto message you could use a `repeated` field, and in the carla_server.h a pointer to an array and a size. Now there is a complication here, you need to copy the memory inside CarlaServer because the data is sent asynchronously to the client. You can take a look at how is done with the non-player agents in "CarlaMeasurements" class, they have the same issue.

At some point (soon), all of the measurements are going to be converted to sensors (like the Lidar or the cameras) because we won't have a "player" vehicle anymore. This looks like a great idea to be a sensor on its own that you could attach to any vehicle in the scene. It could be directly implemented like this if you prefer. (Take a look at how the lidar works for instance).

---

Review status: 0 of 13 files reviewed, 4 unresolved discussions (waiting on @ManueleSigona and @nsubiron)

---

*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Game/CarlaPlayerState.cpp, line 52 at r1](https://reviewable.io/reviews/carla-simulator/carla/549#-LGVgY8z5OgjH2m65fa_:-LGVgY8z5OgjH2m65faa:b326yex) ([raw file](https://github.com/carla-simulator/carla/blob/c5391ad4b19ffdca544cb8ad33c39b986a9e565f/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Game/CarlaPlayerState.cpp#L52)):*
> ```cpp
>       WheelFROmega = Other->WheelFROmega;
>       WheelRLOmega = Other->WheelRLOmega;
> 	  WheelRROmega = Other->WheelRROmega;
> ```

Please, do not use tabs, you get wrong indentation in the diff view.

---

*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp, line 128 at r1](https://reviewable.io/reviews/carla-simulator/carla/549#-LGVfmNl1jF6gPwffrzv:-LGVfmNm4G4ociMyQnxS:b-7e0atv) ([raw file](https://github.com/carla-simulator/carla/blob/c5391ad4b19ffdca544cb8ad33c39b986a9e565f/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp#L128)):*
> ```cpp
>   Angle = Wheel->GetRotationAngle() * 0.01745329251f;
>   return Angle;
> }
> ```

I'm pretty sure you could refactor and use a single function instead of duplicating code. Also the check should make sure that there are enough wheels, i.e. if you access wheel at index 2, it should check(Wheels.Num() > 2).

---

*[Util/CarlaServer/include/carla/carla_server.h, line 191 at r1](https://reviewable.io/reviews/carla-simulator/carla/549#-LGVetq13whITqY7Zvd3:-LGVetq25FGyXscIw27q:b1vjpny) ([raw file](https://github.com/carla-simulator/carla/blob/c5391ad4b19ffdca544cb8ad33c39b986a9e565f/Util/CarlaServer/include/carla/carla_server.h#L191)):*
> ```objc
>     float wheelFROmega;
>     float wheelRLOmega;
>     float wheelRROmega;
> ```

Please, follow the convention of other measurements, lower case with underscores.

---

*[Util/Proto/carla_server.proto, line 152 at r1](https://reviewable.io/reviews/carla-simulator/carla/549#-LGVf4elEFerTigBC9pK:-LGVf4elEFerTigBC9pL:b-xh9dn7) ([raw file](https://github.com/carla-simulator/carla/blob/c5391ad4b19ffdca544cb8ad33c39b986a9e565f/Util/Proto/carla_server.proto#L152)):*
> ```protobuf
>     float wheelFROmega = 14;
>     float wheelRLOmega = 15;
>     float wheelRROmega = 16;
> ```

Here the convention should also be all lower case, separating with underscores.

---


*Comments from [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/549#-:-LGVh1IG4mKZf1-6_hWc:b-r7o97z)*

PR #763, opened 2018-09-06
Fix streaming client fails to connect in Windows

### LINKED ISSUE #761: Black screen in manual_control.py in Windows
Version 0.9.0.

The actual problem is that the connection to the streaming port fails (and the failure message is debug only). Somehow using the tcp endpoint as given only works on Linux. I've tried using a resolver instead and seems to work.

```cpp
boost::asio::ip::tcp::resolver resolver(_socket.get_io_service());
auto endpoint_it = resolver.resolve({"127.0.0.1", "2001"});
boost::asio::async_connect(_socket, endpoint_it, _strand.wrap(handle_connect));
```

nsubiron:
#### Description

~~I have change the way the streaming client connects, now it uses a resolver to find the endpoint based on IP and port instead of building the endpoint directly. Not sure why it failed but now it's more robust.~~

Final version hasn't been tested in Windows, please @iFuSiiOnzZ let me know if it fails.

Fixes #761.

  * **Python version(s):** 2.7
- - -
**EDIT**

So the problem wasn't what I thought initially, the quick test I did succeeded for a different reason. I still couldn't test it on Windows though. Anyway, I found a flaw in my original design of the streaming library.

To give a bit of context, this is how it works. When you create a sensor through the RPC channel, this creates a new "stream" in the streaming server and connects the sensor to it so it sends the data through the stream. Each stream has a token used by a streaming client to subscribe to the stream. The token contains the endpoint (ip address + port) for the client to connect and read the data. The problem is that I was using the same endpoint that I use to connect the server (localhost). Somehow this worked on Linux because we only tested in the same computer, but maybe Windows is not accepting this (I'm still not sure if it will accept the new version).

Now I added more flexibility, so the token may not contain any address info (only port), or may contain a different endpoint (say you want to connect through a different IP address, or maybe in the future the stream is on a different computer). If it doesn't have an address (this is by default), a fallback address is used, in this case the same address used by the RPC server.

I also rolled back the resolver part because already have a fixed endpoint, we don't need to resolve anything (at least not at that point).

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/763)

### DIFF
--- a/LibCarla/source/carla/streaming/EndPoint.h
+++ b/LibCarla/source/carla/streaming/EndPoint.h
@@ -10,12 +10,6 @@
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/asio/ip/udp.hpp>
 
-#include <memory>
-#include <mutex>
-#include <unordered_map>
-
-#include <experimental/optional>
-
 namespace carla {
 namespace streaming {
 namespace detail {


M-Hammod:
I tested this in windows and it works fine now. Thank you

Issue #821, opened 2018-09-30
Add save_to_disk methods to the sensor measurements


nsubiron:
Child of #630.

Add `save_to_disk(path)` functions to at least images and lidar measurements. Same as we had in 0.8.x API, but now needs to be implemented in C++.

Blocked by #815.

PR #829, opened 2018-10-02
Improve manual control steering

iFuSiiOnzZ:
#### Description
Currently when the left or right key was pressed the wheels were steering to the maximum angle making the car a bit hard to control. Now, initially the wheels steer 0.2%, which will increase over the time until reaching the maximum steering point. This way we get a more smooth driving experience.

Blocked by #821.

  * **Platform(s):** Windows 10 x64 1803
  * **Python version(s):** 3.7

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/829)

### DIFF
--- a/PythonAPI/manual_control.py
+++ b/PythonAPI/manual_control.py
@@ -24,13 +24,23 @@ Use ARROWS or WASD keys for control.
 STARTING in a moment...
 """
 
+###############################################################################
+
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


nsubiron:
Hey, I'm doing massive changes to the sample scripts, instead of merging this PR I will rebase it onto my branch and continue from here.

Issue #810, opened 2018-09-27
Including LibCarla in a C++ project


harlowja:
Due to how LibCarla (especially the client) builds its own version of boost and rpclib and cstdlib and clang and ... ? I am wondering if this code was made to be included by other C++ applications. It does not really appear that it is possible to use it as is (due to the bundled libraries and such) so I am wondering if there needs to be some kind of documentation saying yes this is possible, or no it isnt.

nsubiron:
We have now an example [Examples/CppClient](https://github.com/carla-simulator/carla/tree/master/Examples/CppClient).

jemishGhoghari:
> We have now an example [Examples/CppClient](https://github.com/carla-simulator/carla/tree/master/Examples/CppClient).

Hello @nsubiron,

How to build this Example on Windows? or Do you have any lead how to do that?

Any Help appreciated!

Thanks

Issue #784, opened 2018-09-18
Control over multiple vehicles in batch mode


kloud1989:
With the latest version (0.9.0), it is now possible to generate vehicles from the client side and control each one of them. But there is no way to control them in a single RPC call. I noticed that the RPC calls issued to control vehicles are high performance cost ops. If there are multiple vehicles to control in one client, the performance is not so good. Maybe control in a batch mode can help? Also it is weird to see changes of vehicles one after another. With batch sync mode, it would be more realistic to see the changes of all vehicles at the same time.

Child of #630.

nsubiron:
This is planned, but not yet implemented.

nsubiron:
Still not implemented, but on next release 0.9.1 apply control/transform functions are non-blocking and thus much cheaper. It's something.

kloud1989:
It seems great. Thanks for your effort @nsubiron ! What I mostly concerned about is the simulation performance. This fix should solve my problem. 

By the way, I'm curious about if sync mode is supported in the next release?

nsubiron:
Sync mode is not yet ready. There is though a method for the client to wait for a simulator update, 
```py
timestamp = world.wait_for_tick()    # block until an update is received
# or
world.on_tick(lambda timestamp: ...) # execute callback on each new update
```
but not yet the other way around. We definitively want to add it, but having a sync mode in a multi-client environment is a bit more complicated that what we had.

kloud1989:
If it is not possible to support sync mode with multiple clients, is it possible to configure the server in single-client mode, and support sync mode with that? In my case, I only need 1 client, but with multiple controllable vehicles. I'm not sure if this suggestion agrees with your roadmap, but without sync mode, it seems not possible to integrate Carla with other traffic simulators like SUMO.

Sakrust:
This is indeed a key feature to use Carla with any other tool.

PR #856, opened 2018-10-10
Client side improvements

### LINKED ISSUE #767: Deadlock when closing the simulator while a client is connected
To reproduce:

1) Launch the simulator (game or editor)
2) Connect a client
3) Stop the simulation

Version 0.9.0 at least both Linux and Windows.

My bet goes to this piece of code where the future.get() may block forever
https://github.com/carla-simulator/carla/blob/a1684ad0ae52344b23eaf0358ed9f40fe845f898/LibCarla/source/carla/rpc/Server.h#L59-L61

### LINKED ISSUE #809: Relase GIL at python API boundaries
From the looks of it the boost python (new) layer is not releasing the GIL for python.

This makes having a responsive UI (using pyglet or pygame) pretty hard to accomplish (especially say when a timeout to the carla server happens).

An easy way to test this is to setup a server, then try to run a client, then after the client connects, kill the server and the app will usually freeze up.

Something like https://stackoverflow.com/a/41168814 is probably needed...

### LINKED ISSUE #843: Streaming client keeps trying to reconnect after destroying a sensor
When you subscribe to a sensor stream from the client side, a new "streaming client" is created that listens to any data generated by that sensor, on disconnect this client continuously tries to reconnect failing to read data if the sensor was destroyed.

To reproduce, just create a sensor from the client side, set the listen callback, and destroy the sensor. It starts printing warnings from that point on.

nsubiron:
#### Description

Requires #829 and all its dependencies.

Includes a bunch of different improvements to the client-side code:

- Related to #810, C++ API has been refactored, hopefully easier to use now, although it still needs to be polished to be linkable to other apps but it's a step forward.
- Fix #809, release GIL whenever possible (this has really improved performance of the Python API, thanks @harlowja for reporting!).
- Make possible to unsubscribe from a sensor (in Python sensors now have a `stop()` method).
- Fix #843 Streaming client keeps trying to reconnect after destroying a sensor.
- Fix #767 Deadlock when closing the simulator while a client is connected (or at least one of them).
- Fix crash on simulator shutdown if a client has connected at some point.
- Related to #784, vehicle control and set transform methods are now sent async which greatly improves performance in the client-side (we still want to add batch mode though).
- Vehicle control is cached and not sent if haven't changed. Now it's possible to call `vehicle.apply_control(...)` continuously and the message will only be sent when necessary.

Plus I redesigned the Python examples:

- `example.py` is gone, now there is a `tutorial.py` in which a simple example on how to use the API is commented step by step.
- `manual_control.py`now has a fancy user interface (for Carla standards :wink:), and in general the driving experience should be smoother. Although you still depend on a simulator running fast, the input is captured at a decent frame rate. It is also possible to change the main view between the different cameras available.

  * **Python version(s):** 2.7

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/856)

### DIFF
--- a/LibCarla/source/carla/client/Sensor.cpp
+++ b/LibCarla/source/carla/client/Sensor.cpp
@@ -9,6 +9,8 @@
 #include "carla/Logging.h"
 #include "carla/client/detail/Client.h"
 
+#include <exception>
+
 namespace carla {
 namespace client {
 
@@ -20,7 +22,11 @@ namespace client {
           GetDisplayId());
     }
     if (_is_listening) {
-      Stop();
+      try {
+        Stop();
+      } catch (const std::exception &e) {
+        log_error("excetion trying to stop sensor:", GetDisplayId(), ':', e.what());
+      }
     }
   }
 
--- a/LibCarla/source/carla/client/detail/ActorState.cpp
+++ b/LibCarla/source/carla/client/detail/ActorState.cpp
@@ -12,10 +12,16 @@ namespace carla {
 namespace client {
 namespace detail {
 
-  std::string ActorState::GetDisplayId() const {
-    using namespace std::string_literals;
-    return "Actor "s + std::to_string(GetId()) + " (" + GetTypeId() + ')';
-  }
+  ActorState::ActorState(rpc::Actor description, Episode episode)
+    : _description(std::move(description)),
+      _episode(std::move(episode)),
+      _display_id([](const auto &desc) {
+        using namespace std::string_literals;
+        return
+            "Actor "s +
+            std::to_string(desc.id) +
+            " (" + desc.description.id + ')';
+      }(_description)) {}
 
 } // namespace detail
 } // namespace client
--- a/LibCarla/source/carla/client/detail/ActorState.h
+++ b/LibCarla/source/carla/client/detail/ActorState.h
@@ -27,7 +27,9 @@ namespace detail {
       return _description.description.id;
     }
 
-    std::string GetDisplayId() const;
+    const std::string &GetDisplayId() const {
+      return _display_id;
+    }
 
     World GetWorld() const {
       return _episode;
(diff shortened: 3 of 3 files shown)


harlowja:
Awesome, will look at soon :)

LibCarla/source/carla/NonCopyable.h
```diff
@@ -8,15 +8,30 @@
 
 namespace carla {
 
-  /// Inherit (privately) to suppress copy-construction and copy-assignment.
+  /// Inherit (privately) to suppress copy/move construction and assignment.
```
harlowja: Why not just use https://www.boost.org/doc/libs/1_63_0/libs/core/doc/html/core/noncopyable.html ?

LibCarla/source/carla/client/Client.h
```diff
@@ -6,29 +6,13 @@
 
 #pragma once
```
harlowja: What about having a client that just follows https://en.cppreference.com/w/cpp/language/pimpl

Then there can be a python client and a c++ client 'implemtnation' and all that users of either interact with is the pimpl layer (which makes including the pimpl layer in a project pretty straightfward).

LibCarla/source/carla/NonCopyable.h
nsubiron: Yeah, this comes back from the time we didn't include boost. I just kept using it for consistency, we could alias it here.

LibCarla/source/carla/client/Client.h
nsubiron: Yes, the pimpl is in the detail::Client, this class is pretty useless by itself, the API depends on all the other classes that are in the "client" folder.

What it would be nice is to have a pure-C interface with the very minimum for communicating with the server, so it would be easier to link anywhere. In practice however the communication is quite complex, a lot of things are cashed to avoid calling the server all the time. It would require some thought doing such interface.

LibCarla/source/carla/client/Sensor.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "carla/client/Sensor.h"
+
+#include "carla/Logging.h"
+#include "carla/client/detail/Client.h"
+
+namespace carla {
+namespace client {
+
+  Sensor::~Sensor() {
+    if (IsAlive()) {
```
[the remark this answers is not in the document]
nsubiron: True, I'll make sure none of them do.

PR #860, opened 2018-10-11
Add link to the CoRL talk

schra:
The talk gives a good overview over Carla. Therefore I propose to somewhere link to that talk.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/860)


nsubiron:
Thanks for the link! we didn't know it was available :)

PR #924, opened 2018-10-30
Drop off png support (for now)

harlowja:
This has various issues that it is causing with other libraries
folks are using with carla; so until it gets a little more attention
we should just avoid png support.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/924)

### DIFF
--- a/Util/BuildTools/Setup.sh
+++ b/Util/BuildTools/Setup.sh
@@ -293,12 +293,12 @@ endif ()
 
 EOL
 
-if [ "${TRAVIS}" == "true" ] ; then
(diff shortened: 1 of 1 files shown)


fabianoboril:
I had to update PythonAPI/Setup.sh as well to remove the PNG dependency.

soulslicer:
> Thanks for this request. We would appreciate this change a lot. In addition, I had to update PythonAPI/Setup.sh as well to remove the PNG dependency.

What did you do exactly?

PR #961, opened 2018-11-12
Osi grpc communication

mshaheerm:
**Feature Request**
_Adding new module osi_grpc_communication_module._

**Description**

- The module provides a standardized interface to connect CARLA with third party tools.
- Module uses the gRPC (google remote procedure call) to send the groundtruth measurement data of the host_vehicle and the non_player_agents (vehicles) with OSI standardization (Open Simulation Interface) at every frame in the episode.
- All the proto_to_python generated files are present with the module. they can be readily used for the compilation and working.
- the origional .proto files are also present for the manipulaiton and advancement.
- osi_grpc.proto is used to generate the .py version of the request with protoc and a special gRPC plugin

 **Where has this been tested?**

- **Platform(s):**
Windows 10 x64
- **Python version(s):**
Python 3.6
- **Unreal Engine version(s):**
UE 4.18.3
- **CARLA version(s):**
0.8.4

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/961)


nsubiron:
Hi @mshaheerm,

Please rebase your changes into `stable` so we can merge your changes only, otherwise we are merging whole 0.8.4 and we don't want that.

PR #981, opened 2018-11-20
Initial version Carla ROS bridge

berndgassmann:
The parent property of an actor via python interface is not yet filled.
Therefore, the parent_id of Actors has to be transferred from the CARLA
server via rpc interface.

Changes in detail:

carla/rpc/Actor.h:
- add parent_id field to the Actor class for rpc transport

TheNewCarlaServer.cpp:
- fill the parent_id field with the appropriate value

client/ActorList:
- added GetActor() function to get an actor by id

client/ActorVariant:
- added actor_list optional parameter to Get() and MakeActor() function
which allows to query for the parent actor in case the actor_list is
available

Checklist:

  - [X] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [ ] Extended the README / documentation, if necessary
  - [X] Code compiles correctly
  - [x] All tests passing with `make check`

  * **Python version(s):** 2.7

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/981)


danil-tolkachev:
Hi @berndgassmann!
I want to test this PR but it seems that I need to build whole CARLA. Maybe is there another way to get precompiled binaries for this PR?

berndgassmann:
> 
> 
> Hi @berndgassmann!
> I want to test this PR but it seems that I need to build whole CARLA. Maybe is there another way to get precompiled binaries for this PR?

Hi,
I don't know. A chance could be the CI result if the package would be built and published there... but I'm not used to the github process yet. So maybe there's another possiblity.

Today I worked on the camera, which will be done tomorrow, I hope.
I fixed the rotation values (carla python interface provides now degree not radians).
But did not yet upload; need to finish and to incorporate the comments from Fabian.

nsubiron:
Thanks! I'll do the review

PR #988, opened 2018-11-21
Fix win32 compilation

iFuSiiOnzZ:
#### Description

Depending on the included headers sometimes the compilation was failing giving errors as ` {instrinsic funtion} is not a member of {class name}` or some Unreal functions as `GetObject` was replaced by the `GetObjectW` macro of Windows. So to fix the problem some of the general Unreal includes and some other unused includes have been removed.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/988)


marcgpuig:
Works for me, in Windows and Linux.
However, I can't use Reviewable right now because some weird abuse detection mechanism I triggered somehow while signing in :worried:

PR #1044, opened 2018-12-13
New driving classes to perform vehicle control from CARLA clients

germanros1987:
#### Description

New driving classes for CARLA 0.9.1, client side. 

* RoamingAgent: agent that can drive at different speeds following waypoints based on PID controllers, attending to other vehicles and traffic signs
* LocalPlanner: functionality to navigate waypoints using PID controllers
* LateralControl and LongitudinalControl

#### Where has this been tested?

  * **Platform(s):** ...
  * **Python version(s):** ...
  * **Unreal Engine version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1044)

PR #1062, opened 2018-12-19
Features/advanced traffic lights navigation

germanros1987:
#### Description

This PR substitute the previous #1044 . It includes a new navigation method that accounts for both type of traffic lights, i.e. European style (Town01, Town02) and US style (new Town03* included in 0.9.2).

  * **Platform(s):** ...
  * **Python version(s):** ...


nsubiron:
Rebased

fabianoboril:
The code looks fine. I would like to see that the automatic_control uses/calls methods from manual_control, rather then copying everything. Alternatively, why not simplifying the code here, and only focusing on the simply agent and remove key control, sensors, etc.

PR #1150, opened 2019-01-18
Fix TCP accept error: Too many open files

### LINKED ISSUE #1119: Tcp accept error: Too many open files
As reported a few times, it seems to be a limit on how many sensors can be added to the simulation (even if they're destroyed before creating a new one).

Adding this issue to investigate this problem.

nsubiron:
#### Description

Fixes #1119.

We were keeping unnecessary references to expired server sessions, thus keeping these sessions alive and their sockets open. I made sure streams and sessions are discarded when the sensor is destroyed. Note also that we need to explicitly remove the stream on end play since the sensor actor is not actually destructed until UE4 garbage collector kicks in.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1150)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
     - get_angular_velocity: for getting the angular velocity
     - add_impulse: for applying an impulse (in world axis)
   * Added support for gnss_sensor
+  * Fix TCP accept error, too many open files while creating and destroying a lot of sensors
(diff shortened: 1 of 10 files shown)


nsubiron:
Closing this one as it introduces a regression, these changes are instead included in #1184.

PR #1173, opened 2019-01-24
Updated the libpng dependency for linux

xksteven:
Updating the Docs for installing on linux. The libpng16-dev no longer exists and is instead replaced by libpng-dev and libpng16-16

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1173)


fabianoboril:
Hm... Ubuntu 16.04 says something different:
https://packages.ubuntu.com/search?keywords=libpng&searchon=names&suite=xenial&section=all
libpng-dev refers to libpng12, not libpng16

xksteven:
@fabianoboril you're right my mistake it is supposed to be libpng16-dev.

PR #1182, opened 2019-01-28
Add more commandline parameters to manual_control.py

fpasch:
- Add --spawnpoint to specify a custom spawn point
- Add --lidarsensor to optionally add a lidar sensor to the vehicle
- Add --frontcamera to optionally add a front camera to the vehicle
- Set role_name of camera-manager sensors to 'view' (to be able to differentiate)

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1182)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
     - add_impulse: for applying an impulse (in world axis)
   * Added support for gnss_sensor
   * OpenDriveActor has been rewritten using the Waypoint API, this has fixed some bugs
+  * Added more commandline parameters to manual_control.py
 
 ## CARLA 0.9.2
 
--- a/PythonAPI/manual_control.py
+++ b/PythonAPI/manual_control.py
@@ -136,7 +136,7 @@ def get_actor_display_name(actor, truncate=250):
 
 
 class World(object):
-    def __init__(self, carla_world, hud, actor_filter):
+    def __init__(self, carla_world, hud, actor_filter, actor_spawnpoint, has_lidar_sensor, has_front_camera):
         self.world = carla_world
         self.hud = hud
         self.player = None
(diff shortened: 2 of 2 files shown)


fpasch:
> > _[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1182)_ status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @fpasch and @nsubiron)
> 
> _[a discussion](https://reviewable.io/reviews/carla-simulator/carla/1182#-LXJHiX18xh7cKZ58pYp:-LXJHiX27JfWM_hEHYqb:b9r2cw0) (no related file):_
> But the script already has front camera and lidar options; toggle front camera with TAB, lidar is sensor number 7 (press key `7`). And what's the point of adding these sensors if they're not capturing the data? Am I missing something?

The script has those sensors, yes. But only for visualization purposes. They disappear when hitting TAB. For and AD stack, you want those sensors to exist all the time.

fpasch:
Having another client-script would mean to have a lot of duplicated code. It would be additional effort to keep both up to date.

For the user, it's visible if they're using the ROS bridge. Then another sensor pops up (and remains existing if he changes the view).

fpasch:
It's not ROS specific. It might be uses by any AD stack.

It depends on the general approach to integrate an AD stack. You have two options:
1. Create your own carla client, spawn whatever you require, connect to the callbacks of the sensors.
2. Use one carla client to spawn your own vehicle (including sensors), use a second client to connect to the sensors (e.g. with the ros-bridge)

From my point of view (ROS AD stack developer), the second approach is the more preferable one because:
- you don't have to care about spawning actors (that's done within the generic manual_control.py)
- you immediately have an optimized view on the controlled vehicle

nsubiron:
But nevertheless you need a second script to connect to the sensors, you can just spawn them in that same script easily
```py
for actor in world.get_actors():
    if actor.attributes.get('role_name') == 'hero':
        lidar = world.spawn_actor(lidar_bp, transform, attach_to=actor)
        lidar.listen(...)
        front_camera = world.spawn_actor(front_camera_bp, transform, attach_to=actor)
        front_camera.listen(...)
        break
```
It seems to me a bit strange creating sensors in one script but not using them.

Sorry to block, I understand your use case, I'm just trying to keep manual control simple, it's already a huge script.

As for using manual control for visualization, that's actually a cool idea, I think it could be quite useful having the option to attach to an existing vehicle, sort of `./manual_control.py --follow="hero"`.

Maybe @felipecode or @germanros1987 can comment? they have better perspective on the use cases.

fpasch:
As discussed with @germanros1987, I'll move the customized client in carla-autoware.

PR #1228, opened 2019-02-08
Expose traffic sign's trigger volumes on Python API

nsubiron:
#### Description

Since the traffic sign information is not yet present in the OpenDrive files, it is not possible to determine which road lanes are affected by traffic lights/signs. As a workaround, I'm exposing the trigger volumes that our UE4 autopilot uses for detecting traffic signs. This way it is possible to deduce in Python which waypoints are affected by traffic signs by testing against their trigger volumes.

But, again, this is a workaround, we want to have this info attached to waypoints in the future since this could be computed offline just based on the OpenDrive file.

Changes:

- Adds new Python actor TrafficSign and re-parents TrafficLight actor
- Adds a `trigger_volume` property to TrafficSign actors
- Hides `GetBoundingBox()` from C++ Actor's interface as most actors do not have bounding box

For this to work, traffic sign blueprints should be updated with

output

I haven't made the changes in the art repo cause they would cause conflicts with other unmerged PRs. But I made I build for testing with the blueprints changed, running `no_rendering_mode.py` will display the trigger volumes 

---> [Get build here](https://drive.google.com/open?id=1vl66AZDkZ1Egd6IYD9gCAcFEFHf0YAwF) <---

#### Drawbacks

@XGodina Some trigger volumes are rotated in editor, unfortunately the rotation info cannot be sent to Python with our current system, we should remove this rotation in editor.

trigger

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1228)


XGodina:
Ok, I am working with it.

PR #1236, opened 2019-02-11
Use absolute import

praveen-palanisamy:
#### Description

- Minor fix to make the import statement absolute. The previous import statement causes issues when used by the scenario_runner. Also, this fix makes the planner script to be consistent with the rest of the Python API codebase.

  * **Platform(s):** Ubuntu
  * **Python version(s):** 3.5, 3.6

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1236)


nsubiron:
Oh it seems this was already fixed in master.

PR #1250, opened 2019-02-13
added steering wheel control example

germanros1987:
#### Description

Added support for Logitech G29 steering wheel and a new manual controller

Fixes #  

  * **Platform(s):** 16.04
  * **Python version(s):** 3.5.5
  * **Unreal Engine version(s): NA

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1250)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## Latest Changes
-
+  * Added manual_control_steeringwheel.py to control agents using Logitech G29 steering wheels (and maybe others).
   * Fixed `manual_control.py` and `no_rendering_mode.py` to prevent crashes when used in "no rendering mode"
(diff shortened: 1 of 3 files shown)


Div99:
Hi, can vehicle control be extended to use Xbox controller too? Using the keyboard is a pain, and using a gaming controller is more intuitive.

damascenodiego:
Hi everyone, I've been playing with Carla for a couple of weeks and one of my colleagues wants to buy a Logitech steering wheel to run some experiments with human drivers.

The problem is that so far our seller has only the Logitech G920 available.

My questions are: 
- Are the CARLA commands mapped to buttons common for both models G29 and G920?
- Have you used any of the buttons specific to G29 (e.g., +/-, enter, 24-point Selection Dial)?

PR #1320, opened 2019-02-28
Waypoint API extended

marcgpuig:
#### Description
Extended the waypoint API with new functions:
- `carla.Waypoint.lane_change`: Returns a `carla.LaneChange`.
- `carla.LaneChange`: A Python enum that tells whether traffic rules allow a lane change.
  + 0: `None`
  + 1: `Right`
  + 2: `Left`
  + 3: `Both`
  + Can be used as a flag i.e.:
    * `LaneChange.Right & LaneChange.None => False`
    * `LaneChange.Right & LaneChange.Left => False`
    * `LaneChange.Right & LaneChange.Both => True`
    * `LaneChange.Right & LaneChange.Right => True`
- `carla.Waypoint.lane_type`: Returns an string that codifies the type of the current lane, i.e.: `"driving", "bidirectional", "shoulder", "sidewalk", "median", "parking", "none", ...`
- `carla.Waypoint.right_lane()`: Returns another `carla.Waypoint` (although the rules allow the change or not) at the closest right neighbor lane. If there is no right lane, it returns `None`.
- `carla.Waypoint.left_lane()`: Returns another `carla.Waypoint` (although the rules allow the change or not) at the closest left neighbor lane. If there is no left lane, it returns `None`. Notice that if the new waypoint is on a different signed `lane_id`, it does mean that it's in the opposite direction.

  * **Platform(s):** Linux
  * **Python version(s):** 2.7 & 3.5

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1320)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
   * Fixed static objects present in the map were marked as "movable"
   * Fixed BP_MultipleFloor, tweaked offset in BaseFloor to adjust meshes between them
   * New traffic signs assets: one-way, no-turn, more speed limits, do not enter, arrow floors, Michigan left, and lane end
+  * Extended the Waypoint API with `lane_change`, `lane_type`, `right_lane()` and `left_lane()`
   * Expose traffic sign's trigger volumes on Python API
   * Updated the Python API to enable the user to acquire a traffic light's pole index and all traffic lights in it's group
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@
   * Fixed static objects present in the map were marked as "movable"
   * Fixed BP_MultipleFloor, tweaked offset in BaseFloor to adjust meshes between them
   * New traffic signs assets: one-way, no-turn, more speed limits, do not enter, arrow floors, Michigan left, and lane end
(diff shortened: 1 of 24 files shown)


aravindSwamy94:
Hello,
           This was a good update from Carla side, I just want to post a simple question, regarding this update.
            With this update, we can predict the entire road information for a given position. Now , I just wanted to know, is there any particular way, where we can get the LaneSection of the warpoint in that road ? just curious to know about this. 

 Thanks in advance

marcgpuig:
Hi @aravindSwamy94.
We want to be transparent about opendrive in our API.
But in the next release, we will provide the waypoint OpenDrive's current `s` parameter so, knowing the current road and lane id of the waypoint, you will be able to parse yourself the `.xodr` file and know everything without our API.
In short, currently not.
Cheers!

aravindSwamy94:
@marcgpuig Thank you so much for your reply, getting the s parameter for the waypoint will really solve this issue. Like we can read the s parameters of lane Sections from .xodr file and correspondingly place our points in that section. 


Is this change available in 0.9.4 release from Carla.

I once again thank you so much for the update.

PR #1364, opened 2019-03-11
Bernatx/recorder

bernatx:
#### Description

The recorder system now will load dynamically the map required for the replayer if it is not already loaded. 

Also, we can specify an absolute path with the filenames, so we can write or read to any folder where the sever has access.

A new API function has been added to the recorder system to allow specify the speed of the playback, so values under 1.0 will be used for slow motion, and values greater than 1.0 for fast motion.

Fixes # 

  * **Platform(s):** Linux Ubuntu 16.04
  * **Python version(s):** 2.7
  * **Unreal Engine version(s):** 4.21.2

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1364)


wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp, line 51 at r1](https://reviewable.io/reviews/carla-simulator/carla/1364#-L_vtucI0Xy286fddRvB:-L_vtucI0Xy286fddRvC:bi19qsu) ([raw file](https://github.com/carla-simulator/carla/blob/39b81d30277b836d41377fd52411cb2b8e523525/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp#L51)):*
> ```cpp
> inline void ACarlaRecorder::SetReplayerSpeed(double TimeFactor)
> {
>   return Replayer.SetSpeed(TimeFactor);
> ```

Is the return statement necessary when the return type is void?

PR #1367, opened 2019-03-11
Adding smoke tests to Jenkins

nsubiron:
#### Description

Added infrastructure to perform smoke tests in every Jenkins build. The idea is to have a minimal set of tests to validate each new build, they should not be an extensive test suite as they're executed each Jenkins build.

I added a few basic tests, but more are necessary, now testing
- version match
- reloading current map
- loading every map
- some blueprints validity
- synchronous mode

Some may be more necessary than other, we can evaluate after a while, but some more tests seem to me necessary to add asap (not adding them due to lack of time, but other people can start adding their own).

#### Possible Drawbacks

Build time gets slightly longer.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1367)


nsubiron:
Not adding it to the CHANGELOG as it does not affect user functionality.

PR #1374, opened 2019-03-12
Updating carla settings to allow user to disable rendering and set se…

wattsavolt:
…rver tieout from the command line

#### Description
The user can now disable rendering and set the timeout value for the carla server from the Unreal command line. Now the carlasettings.ini no longer needs to be used.

  * **Platform(s):** Linux
  * **Python version(s):** 2.7

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1374)


hh0rva1h:
@nsubiron The command-line option -no-rendering does not work as of Carla 0.9.9.2, doing it using a init file works though, there has to be a bug somewhere.
Furthermore those options are not properly documented, the respective documentation file has been removed.

PR #1395, opened 2019-03-15
Fix discrepancy between timestamps

### LINKED ISSUE #1381: Timestamp of SensorData diverges
Since #1348 , Carla SensorData contains a timestamp.

Unfortunately this diverges from the timestamp reported by world.on_tick(). The longer you run Carla Server, the greater the difference gets.

It seems that GetWorld().GetTimeSeconds() reports the wrong time.

Tested with parameters: ```-benchmark -fps=20```

nsubiron:
#### Description

Make sensors use timestamp from UCarlaEpisode instead of UWorld. The latter stores the value in a float accumulating greater error.

Fixes #1381.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1395)

### DIFF
--- a/LibCarla/source/carla/sensor/SensorData.h
+++ b/LibCarla/source/carla/sensor/SensorData.h
@@ -40,7 +40,7 @@ namespace sensor {
       return _frame_number;
     }
 
-    /// Time the data was generated.
+    /// Simulation-time when the data was generated.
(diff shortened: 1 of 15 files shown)


nsubiron:
I guess we can skip the CHANGELOG as it was a regression within the same version.

PR #1418, opened 2019-03-20
Add HD maps

johannesquast:
#### Description

Add HD maps for Towns

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1418)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto
+*.pcd filter=lfs diff=lfs merge=lfs -text
--- a/HDMaps/Readme.md
+++ b/HDMaps/Readme.md
@@ -1,0 +1,10 @@
+CARLA HD maps
+=============
+
+This folder contains HD Maps for the various towns provided with CARLA.
+
+The PointCloud files are using the same coordinate system as CARLA.
+This is a left-handed coordinate system with x and y as ground plane and z axis up.
+
+Please be aware that PointCloud tools (especially the PointCloud Viewer "pcl_viewer") is assuming
+a right-handed coordinate system, so it displays everything in a mirrored fashion.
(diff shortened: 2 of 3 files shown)


nsubiron:
Please, don't add LFS files to this repo, that adds an extra dependency to our users and slows down downloads significantly. HD maps should probably go to the [carla-content repo](https://bitbucket.org/carla-simulator/carla-content).

johannesquast:
@germanros1987 @nsubiron Each map is around 100MB size and German asked me to put them in  {CARLA_ROOT}/HDMaps/TownX.pcd. (please see https://github.com/carla-simulator/carla-autoware/pull/19).
With LFS the size of the repo to download is only increases slightly (and not by 1 GB (10 towns with 100MB each).
How to proceed?

nsubiron:
We explored the possibility of using GitHub's LFS in the past for storing the assets, but we encountered a lot of problems, specially pricing, that's why we chose BitBucket for the Carla assets

- There is a quota on GitHub of 1GB/month bandwidth for LFS files, see [About storage and bandwidth usage](https://help.github.com/en/articles/about-storage-and-bandwidth-usage). As far as I understood, if we store 1GB of LFS files, we can only make a single download a month, after that they disable LFS until next month. Forks seem to count against the parent repo quota too. We can buy extra quota, but last time we checked, it was prohibitively expensive because of the download quota.

- On top of that, every time you modify a LFS file you store a new copy of it, we would reach the 1GB storage limit of GitHub if we modify a map once. For instance, we have about 4GB of assets compressed, the repository is already about 30GB big (and it's less than a year old).

- Then, even if we are willing to pay for the extra quota, users need to install LFS to download Carla, now this is not need it, the assets are downloaded as zip from AWS. With old versions of git (default in Ubuntu 16.04), if you forget to do a `git lfs clone` the clone is extremely slow.

Since we have all these problems already solved for the carla-content repo, I would put all the LFS files there. If necessary, we can make in Jenkins a separate package (or repo) with the HD maps and upload it automatically to AWS.

johannesquast:
Hi @germanros1987 and @nsubiron,
I'll do a PR at bitbucket with all the maps once I've created them completely.

PR #1451, opened 2019-03-27
Reworked OpenDrive parser

marcgpuig:
#### Description

Since we had a limited OpenDrive parsing, recently appeared the need to expand it.

- Now the whole OpenDrive is parsed and not only what we initially need.
- The internal structure is more similar to OpenDrive itself, so it is more expandable and queries are easy.

#### Python API Changes

Our idea always is not to change how the Python API is used once we add new functionalities, but in this case, this was not possible.

###### Additions:

* ~~**carla.Waypoint.id**: Unique identifier for a Waypoint based on the hash of it's `road_id`, `section_id`, `lane_id`, and `s`.~~ (already in master, added section_id)
* **carla.Waypoint.section_id**: Refers to OpenDrive's [`laneSection`](http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf#page=21) number used as an id.
* ~~**carla.Waypoint.s**: The [`s`](http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf#page=15) parameter of OpenDrive, it defines the position along the reference line, measured in meters from the beginning of the track, calculated in the XY-plane.~~ (already in master)
* **carla.LaneType**: Define the type of lane. The most common are, _NONE_, _Driving_, _Shoulder_, _Sidewalk_, _Bidirectional_, and _Parking_ although OpenDrive use [more of them](http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf#page=93). These enums can be used as flags for queries.
* **_Extended_ carla.LaneMarking**:
  + **type**  Defines the [types](http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf#page=92).
  + **color**: Defines the [color](http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.4H.pdf#page=92).
  + **lane_change** contains the lane change information of a single road mark, indicating from which side of it you are allowed to cross.
  + **width**: Width of the road mark.
* **carla.Waypoint.right_lane_marking** / **carla.Waypoint.left_lane_marking**: Returns a `carla.LaneMarking` for the right or left Waypoint's lane marking.

###### Changes:

* **carla.LaneMarking** is not an enum anymore, but an object. The enum is now **carla.LaneMarking.type**.
* **carla.Waypoint.lane_type** doesn't work with strings anymore, now uses an enum `carla.LaneType`.
* **carla.Map.get_waypoint(location, project_to_road=True, lane_type=carla.LaneType.Driving)**: you can use it like before and will work as expected, but now accepts a `carla.LaneType` that can be used as a flag to filter lane types, for instance:
`carla.Map.get_waypoint(location, project_to_road=True, lane_type=(carla.LaneType.Driving | carla.LaneType.Sidewalk)` will search for the nearest waypoint on the center of a _Driving_ or _Sidewalk_ lane.

###### More

* Allow "carla.Map" object to be constructed from a XODR file (now it can be constructed without connecting to the simulator).
* New functionality FileSystem::ListFolder.
* Const and non-const versions of iterator::make_map_{values,keys}_iterator, now they also work on references instead of copies.
* Add additional repo for testing resources (OpenDrive files), automatically downloaded when running unit tests.

#### To Do

- [x] Update CHANGELOG
- [x] Update python_api.md
- [ ] Fix LaneCrossingCalculator
- [x] ~~Test OpenDriveActor~~ (done in a different branch)
- [x] Fix uses of `lane_type == 'driving'` (replace with `lane_type == carla.LaneType.Driving`)
- [x] Fix uses of `waypoint.next(precision)[0]`

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1451)


nsubiron:
:+1: well done @nsubiron.

PR #1489, opened 2019-04-03
Spawn actors in batch

nsubiron:
#### Description

Added methods to spawn actor in batch mode, reducing the number of accidents related to vehicles spawning on top of each other.

  * Added new `carla.command.SpawnActor` to spawn actors in batch
  * Added method `client.apply_batch_sync` that waits for server response
  * Updated `spawn_npc.py` to spawn vehicles in batch

The `SpawnActor` command also accepts queuing commands to be applied to the newly spawned actor.
```py
batch.append(
    SpawnActor(blueprint, transform).then(
        SetAutopilot(FutureActor, True)).then(
            ApplyVelocity(FutureActor, Vector3D(z=5))))
```

 `spawn_npc.py` has been updated to spawn all the vehicles in batch, but now it does not support spawning more vehicles than spawn points. This is probably a good thing due to the amount of accidents that it was producing.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1489)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@
   * Fixed vehicles missing the route if autopilot enabled too late
   * Enhanced stop triggers options
   * Fixed division by zero in is_within_distance_ahead()
+  * Added new carla.command.SpawnActor to spawn actors in batch
+  * Added method `client.apply_batch_sync` that waits for server response
+  * Updated `spawn_npc.py` to spawn vehicles in batch
 
 ## CARLA 0.9.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@
   * Fixed division by zero in is_within_distance_ahead()
   * Added new carla.command.SpawnActor to spawn actors in batch
   * Added method `client.apply_batch_sync` that waits for server response
+  * Added optional argument "actor_ids" to world.get_actors to request only the actors with the ids provided
   * Updated `spawn_npc.py` to spawn vehicles in batch
 
 ## CARLA 0.9.4
--- a/Docs/python_api.md
+++ b/Docs/python_api.md
@@ -21,6 +21,7 @@
 - `show_recorder_collisions(string filename, char category1, char category2)`
 - `show_recorder_actors_blocked(string filename, float min_time, float min_distance)`
 - `apply_batch(commands, do_tick=False)`
+- `apply_batch_sync(commands, do_tick=False) -> list(carla.command.Response)`
 
 ## `carla.World`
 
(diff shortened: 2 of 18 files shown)


wattsavolt:
___
*[PythonAPI/examples/spawn_npc.py, line 95 at r1](https://reviewable.io/reviews/carla-simulator/carla/1489#-Lbciu8W0Ee8JLzJFPVJ:-LbcikN-9ygeFuTFSFS0:b-8jszbc) ([raw file](https://github.com/carla-simulator/carla/blob/e6debe3c0faadbfaeb127bff91a328d7a93edb36/PythonAPI/examples/spawn_npc.py#L95)):*
> ```Python
>         batch = []
>         for n, transform in enumerate(spawn_points):
>             if n >= args.number_of_vehicles:
> ```

This conditional check seems unnecessary because args.number_of_vehicles is assigned to the number of spawn points above if the original value was greater.

tinmodeHuang:
@nsubiron  I haven't found the call `carla.command.FutureActor` in the documentation, introducing it to me, please

nsubiron:
@tinmodeHuang It's just a place-holder value to make the code more readable, it'll be replaced by the actor id of the newly created actor.

tinmodeHuang:
thanks! and otherwise could you explain it to me that the statement `light_state = vls.Position | vls.LowBeam | vls.LowBeam`?

doterop:
That is just setting the light configuration of the vehicles to turn them on

tinmodeHuang:
the call `print('light_state:',light_state)` outputs as follows:
```
(python37) E:\CARLA_0.9.9.4\WindowsNoEditor\PythonAPI\examples>python spawn_npc.py --car-lights-on
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
light_state: 3
```
so what the 3 means

doterop:
One thing that I notice is that one `vls.LowBeam` is duplicated and didn't need it, so the final set should be:
`light_state = vls.Position | vls.LowBeam`

`VehicleLightState` is an enum used as a flag to configure the lights of the vehicles. In this case, the position and low beam lights are set to be turned on. The 3 value that you see is just the final result of the mask.

tinmodeHuang:
I can't agree more with you, and then `light_state = vls.Position | vls.LowBeam | vls.HighBeam` means that there are seven kinds of states for lights of a vehicle.
`light_state: 7`

doterop:
In that case, you are setting those 3 types of lights, but the 7 is the result of the mask not the number of lights.

tinmodeHuang:
all states lists as follows:
```turn on: None, Position, LowBeam, HighBeam, Position + LowBeam, Position + HighBeam and what is the other?```

**Comment by doterop** [comment]
LightState is an enum whose values are used as flags, so its values are:

```
NONE          = 0         = 0   
Position      = 0x1       = 1   
LowBeam       = 0x1 <<  1 = 2   
HighBeam      = 0x1 <<  2 = 4   
Brake         = 0x1 <<  3 = 8   
RightBlinker  = 0x1 <<  4 = 16  
LeftBlinker   = 0x1 <<  5 = 32  
Reverse       = 0x1 <<  6 = 64  
Fog           = 0x1 <<  7 = 128 
Interior      = 0x1 <<  8 = 256 
Special1      = 0x1 <<  9 = 512 
Special2      = 0x1 << 10 = 1024
```

If you want to turn on only position lights: `light_state = vls.Position` and light_state is 1 

If you want to turn on only low beam lights: `light_state = vls.LowBeam`  and light_state is 2

If you want to turn on position and low beam lights: `light_state = vls.Position | vls.LowBeam` and light state is 3 because the mask of those light types is: light_state = 1 | 2 = 3. Is an add bitwise operation.

**Comment by tinmodeHuang** [comment]
get it, thank you very much, but I'm sorry for having the last problem, a block of statements in spawn_npc.py as follows:
```python
                batch.append(SpawnActor(blueprint, transform)
                    .then(SetAutopilot(FutureActor, True, traffic_manager.get_port()))
                    .then(SetVehicleLightState(FutureActor, light_state)))

        for response in client.apply_batch_sync(batch, synchronous_master):
            if response.error:
                logging.error(response.error)
            else:
                vehicles_list.append(response.actor_id)
                print('\n',response.actor_id)
```
it outputs as follows:
```bash
(python37) E:\CARLA_0.9.10.1\WindowsNoEditor\PythonAPI\examples>python spawn_actor.py --car-lights-on
{'number_of_vehicles': 10, 'number_of_walkers': 50, 'safe': False, 'filterv': 'vehicle.*', 'filterw': 'walker.pedestrian.*', 'tm_port': 8000, 'sync': False, 'hybrid': False, 'seed': None, 'car_lights_on': True}

 96

 97

 98

 99

 100

 101

 102

 103

 104

 105
```
I wanna know about how to get the actor `vehicle.tesla.model3` I want.
maybe I should clear my head to use other way being convenient to spawning all actor.

doterop:
I suggest you to use `spawn_actor` instead to control one specific actor.

tinmodeHuang:
> I suggest you to use `spawn_actor` instead to control one specific actor.
fair enough

PR #1515, opened 2019-04-09
Update content

nsubiron:
Updated the content link to download latest assets.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1515)


daohu527:
@nsubiron 
I have downloaded the package several times with chrome. Every time I fail and download again from the beginning. Is this my network problem? Because the link is Google Cloud, but I am in China? I almost downloaded 1.1G and it failed. 

Is there any other way to download it? Thx!

PR #1537, opened 2019-04-15
Bugfixes about recorder:

bernatx:
#### Description

There are some bug fixes in the recorder system:

1) Fixing the camera following when a new map is loaded by the recorder. 

The wrong variable was used in that case ('FollowingId' instead of the 'ThisFollowingId')

2) Fixing positions at start of replayer if another replayer was play. 

The second time the replayer is called, the actors were interpolated with its previous position and the new one, but at the beginning the previous position was where the previous replayer leaves the actor, so the actors were interpolated from its current position and the new position they need to go, and the effect was that the actors were sliding fast to its new position for a short time (only in 1 frame).

3) The API function 'show_recorder_file_info' was showing the wrong parent id of an actor.

It was showing the same id than the child actor because was using the wrong field name ('DatabaseId' instead of 'DatabaseIdParent').

Fixes #  

  * **Platform(s):** Ubuntu
  * **Python version(s):** 2.7 and 3.7

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1537)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Latest
 
+  * Recorder fixes:
+    - Actors at start of playback could interpolate positions from its current position instead than the recorded position, making some fast sliding effect during 1 frame.
+    - Camera following in playback was not working if a new map was needed to load.
+    - API function 'show_recorder_file_info' was showing the wrong parent id.
   * New recorder features:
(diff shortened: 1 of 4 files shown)


bernatx:
The changes to this PR will be applied in PR#1568, so we close this PR.

PR #1549, opened 2019-04-17
Improving visual quality of the rgb scene capture

wattsavolt:
#### Description

Added temporal aliasing to screen captures with post processing. This gets rid of the jaggies during motion. Set the target gamma to 1.4. I found this value to be the best. It minimises the brightness difference between the screen capture and the main camera. There is still a small difference on some monitors. I disabled auto exposure because it was disabled on the main camera.  

Fixes #  

  * **Platform(s): Ubuntu 16.04
  * **Python version(s): 3.5
  * **Unreal Engine version(s): 4.22.0

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1549)


wattsavolt:
PR updated to address save_to_disk easy. Alpha values now set to 255.

nsubiron:
Have you measured what's the cost of setting the alpha channel to max? for a 1920x1080 image for instance? That's executed in the render thread, every millisecond there reduces our FPS, unless that's very small I would do it in the client-side when the image is received, there is executed asynchronously.

analog-cbarber:
Isn't FPixel reader used by all the camera sensors? I have added a sensor in my fork that uses all four channels. Won't setting alpha to max break sensors like that?

nsubiron:
@analog-cbarber Yes, that would set alpha to max for all the camera-based sensors. In carla we don't have at the moment any sensor using the alpha channel, but I see this could be a problem.

Ideally we would like to tell Unreal to generate a buffer without alpha, but I don't think that's possible with these cameras.

analog-cbarber:
You can indeed set the alpha channel in Post Processing materials in 4.21.

I have created a "reflectance" material for use in Lidar modeling. It computes reflectance separately for the RGB channels and uses the alpha channel to report the proportion of reflectance due to diffuse vs specular reflectivity so that we can apply different distance-based attenuation for specular vs diffuse light. Maxxing the alpha channel would make everything look like it was 100% diffuse for my sensor.

Since I am working on a fork, I can always just undo the change, but I was hoping to eventually contribute this back to CARLA so it would be nice if this was something that was supported.

nsubiron:
Then I think setting alpha to max should be done in the client side in the `ImageSerializer::Deserialize`, then if there is a sensor that uses the alpha channel it can implement its own serializer (it can reuse  `ImageSerializer::Serialize` but overload the deserialize function). That'd have zero overhead because all the dispatch is done at compile-time.

wattsavolt:
The cost was relatively small but I will move to client side to ensure it works for any new sensors.

wattsavolt:
Now pixels changed on client side as requested. As Nestor said above, the Deserialize function will have to be overridden by new sensors that need the alpha value

wattsavolt:
___
*[LibCarla/source/carla/sensor/s11n/ImageSerializer.cpp, line 21 at r7](https://reviewable.io/reviews/carla-simulator/carla/1549#-LeNNQhBCnuCqS1Od-wS:-LeX6noiEMI4629ZPQHC:b6x1eqy) ([raw file](https://github.com/carla-simulator/carla/blob/b44cacb8684dd39469a855fd3e4fffcfa4ee1b12/LibCarla/source/carla/sensor/s11n/ImageSerializer.cpp#L21)):*

Not that makes any difference to the result, but wouldn't have been easier doing this :wink:
```cpp
  SharedPtr<SensorData> ImageSerializer::Deserialize(RawData data) {
    auto image = SharedPtr<data::Image>(new data::Image{std::move(data)});
    // Set alpha of each pixel in the buffer to max to make it 100% opaque
    for (auto &pixel : *image) {
        pixel.a = 255u;
    }
    return image;
  }
```
</blockquote></details>

Does data::Image constructor remove the header info from the data it's passed?

wattsavolt:
___
*[LibCarla/source/carla/sensor/s11n/ImageSerializer.cpp, line 21 at r7](https://reviewable.io/reviews/carla-simulator/carla/1549#-LeNNQhBCnuCqS1Od-wS:-LeX8X4mDiJlMioy-B2Z:b-y9krue) ([raw file](https://github.com/carla-simulator/carla/blob/b44cacb8684dd39469a855fd3e4fffcfa4ee1b12/LibCarla/source/carla/sensor/s11n/ImageSerializer.cpp#L21)):*
<details><summary><i>Previously, wattsavolt (Aidan Clear) wrote…</i></summary><blockquote>

Does data::Image constructor remove the header info from the data it's passed?
</blockquote></details>

Just looked at the code, the data::Array class that it derives takes the offset into account so I'll change to your version. Thanks.

marcgpuig:
We noticed that somehow when you are using our Scene Capture camera, the results on the client and the editor doesn't match on low values of gamma, maybe there is more post process going on on background. Also, we must choose which value of gamma we should use. 1.4 gets the nearest result to the Editor, but the "standard" is 2.2 and maybe what should be fixed is the light values.
Opinions @wattsavolt @TheNihilisticRobot @XGodina @nsubiron?

wattsavolt:
CHANGELOG line duplication issue fixed

wattsavolt:
*[a discussion](https://reviewable.io/reviews/carla-simulator/carla/1549#-LfZJOjr4wCIHMIqqPWW:-Lf_3tiD5SVWJ2wZnMP4:bh9tk5k) (no related file):*
<details><summary><i>Previously, TheNihilisticRobot wrote…</i></summary><blockquote>

I don't remember the specific value but I remember to have setted the Gamma when I was working on the weather blueprint to the value I thought that matched the visual results between the capture camera and the editor camera the most, making it possible to work for both cameras simultaneously.
If we set now for a more standard value we should tweak the weather a bit but focusing on the camera we need to give the best results (presumably the screen capture one).

</blockquote></details>

1.4 was the value that made the screen capture look closest to the game camera for me. On one of my monitors, it was exactly the same. On the other, the screen capture was still a little brighter. I've set it back to 2.2 because that probably works best for everyone.

---

wattsavolt:
___
*[CHANGELOG.md, line 18 at r8](https://reviewable.io/reviews/carla-simulator/carla/1549#-LfZ4Nwo--h537t4sKGA:-LftJhLg0w66RHD0Dnf7:bt2vr0b) ([raw file](https://github.com/carla-simulator/carla/blob/fa6be64bd757fd9b4edeceee524dc42a45947e18/CHANGELOG.md#L18)):*

This line is duplicated, comes from another PR
</blockquote></details>

Fixed

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp, line 168 at r11](https://reviewable.io/reviews/carla-simulator/carla/1549#-LftFvNeAH6XlNUQoPNi:-LftKZ3r7ZkTwPFPR1ds:b-e34vp4) ([raw file](https://github.com/carla-simulator/carla/blob/0dd6aca8c0dc0189cf4aea37d0d0c86b5d904b92/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp#L168)):*
<details><summary><i>Previously, marcgpuig (Marc Garcia Puig) wrote…</i></summary><blockquote>

please, remove the unnecessary spaces
</blockquote></details>

lol you really picking at straws :D

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp, line 168 at r11](https://reviewable.io/reviews/carla-simulator/carla/1549#-LftFvNeAH6XlNUQoPNi:-Lftb7u_AW7RQn1D85UC:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/0dd6aca8c0dc0189cf4aea37d0d0c86b5d904b92/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp#L168)):*
<details><summary><i>Previously, wattsavolt (Aidan Clear) wrote…</i></summary><blockquote>

lol you really picking at straws :D
</blockquote></details>

Done.

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp, line 17 at r11](https://reviewable.io/reviews/carla-simulator/carla/1549#-LftEZLH6LtOXZBRC5XU:-LftbMCdCiP09fAqlwR1:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/0dd6aca8c0dc0189cf4aea37d0d0c86b5d904b92/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp#L17)):*
<details><summary><i>Previously, marcgpuig (Marc Garcia Puig) wrote…</i></summary><blockquote>

After the last changes, this isn't needed anymore right?
</blockquote></details>

Done.

wattsavolt:
*[a discussion](https://reviewable.io/reviews/carla-simulator/carla/1549#-LfxSo4Y2vByMdHRRJub:-LfxzSkfCy50OYl8mNFi:b-p1ak6u) (no related file):*

https://youtu.be/VEpaXIVygM0
</blockquote></details>

Sorry about that. I've set the values back to default. I am creating another PR for changing motion blur settings from python api. 

---

wattsavolt:
*[a discussion](https://reviewable.io/reviews/carla-simulator/carla/1549#-LfxSo4Y2vByMdHRRJub:-LfxzyoV5vQglGSOIch4:b-896fix) (no related file):*
<details><summary><i>Previously, wattsavolt (Aidan Clear) wrote…</i></summary><blockquote>

Sorry about that. I've set the values back to default. I am creating another PR for changing motion blur settings from python api. 
</blockquote></details>

Done.

---

PR #1552, opened 2019-04-18
RPC server optimizations 

nsubiron:
#### Description

Several optimizations and small refactoring, most remarkable the server now won't wait for game-thread tasks if the client call them asynchronously, this greatly increases the load of messages that the server can parse.

- Added metadata to messages, now only contains info on whether the call was async, but later we can reuse this for client ids and permissions.
- Removed dynamic allocations when possible (removed `std::function` and `std::shared_ptr` from function bindings).
- Refactored `AsioThreadPool` and test `ThreadPool` into a generic `carla::ThreadPool`, it also uses the same tricks to avoid dynamic allocations.
- Moved methods in "Overload.h" to its own `carla::Functional` class.

Client and server interfaces remain unchanged :+1: 

#### Possible Drawbacks

Now we uncovered another issue, if you send requests to the server too fast some of the messages might get lost; maybe an issue of rpclib but we should investigate further.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1552)


wattsavolt:
___
*[LibCarla/source/carla/Functional.h, line 67 at r1](https://reviewable.io/reviews/carla-simulator/carla/1552#-Ldhd8_f63ZV-dd-4lHk:-LdibaLJ6muwoH5oGUSS:b78nrvg) ([raw file](https://github.com/carla-simulator/carla/blob/4872e161ec1fcf17dd8d74b934439b8041b7d27d/LibCarla/source/carla/Functional.h#L67)):*

?
</blockquote></details>

lol sorry accidentally quoted those lines

PR #1588, opened 2019-04-29
Fix dead-lock when loading a new map in synchronous mode.

### LINKED ISSUE #1412: Loading multiple maps in sequence
The following code excerpt fails:

```
import carla
client = carla.Client('localhost', int(2000))
client.set_timeout(25.0)
world = client.load_world('Town01')
settings = world.get_settings()
settings.synchronous_mode = True
world.apply_settings(settings)

world = client.load_world('Town02')
```

Basically it seems that you need to call world.tick() before loading another
map. I think map loading should not need to wait for client synchronization.

nsubiron:
#### Description

It seems we forgot to reset the settings after a new world is loaded, the intended behaviour was to reload the map with default settings because a fresh new map is created, and in fact the new world had the settings reloaded but the engine was not properly notified thus creating inconsistent state.

The following works now (also without ticking in between).
```py
client.set_timeout(2.0)

world = client.get_world()

while True:
    print('new world: enable sync mode.')
    world.apply_settings(carla.WorldSettings(synchronous_mode=True))

    for _ in range(0, 10):
        time.sleep(1)
        world.tick()
        world.wait_for_tick()
        print('tick!')

    world = client.reload_world()
```

Fixes #1412.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1588)


pawel-ziecina:
What do you think about extending tests in https://github.com/carla-simulator/carla/blob/master/PythonAPI/test/smoke/test_sync.py with related with issue cases?

nsubiron:
@pawel-ziecina good call, I'll do that.

felipecode:
:lgtm:

PR #1639, opened 2019-05-14
Set vulkan as default graphics api for linux. Changed make files to allow user to start …

wattsavolt:
…unreal editor with vulkan or opengl

#### Description
Since Unreal 4.21, Unreal picks Vulkan instead of OpenGL as the default graphics API for Linux. If Vulkan is not installed on the user's system or the graphics driver or gpu does not support it, then Unreal will detect this automatically and use OpenGl 4.5 instead. Vulkan is a lower level graphics API and performs better than OpenGL. Users who enable it should see a significant increase in fps when running the simulator. 

To use Vulkan, install on your system along with latest graphics driver using the following tutorial:
https://linuxconfig.org/install-and-test-vulkan-on-linux
Currently 415 is the latest driver so replace 396 with 415.
vulkaninfo | less command can be used to check that vulkan is working on your system. You should be able to see details about your graphics card.

This PR updates the Unreal project settings to add Vulkan to the list of RHIs. Unreal should pick Vulkan instead of OpenGL by default once vulkan is installed on your system.

Note: This change only effects Linux. DirectX 11/12 is used on Windows and Unreal does not support Vulkan on Windows currently.

  * **Python version(s):** irrelevant
  * **Unreal Engine version(s):** 4.22.1
  * **GPUs:** GTX 1080 and RTX 2070 

#### Possible Drawbacks

GPUs below NVidia 900 series may be better off using OpenGL. If this is the case, OpenGL can be enabled when launching the unreal editor with the CARLA make script by doing the following:
make launch ARGS=--use-opengl

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1639)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Vulkan support: Changed project settings to make vulkan default but opengl can be selected using the make script
   * Upgraded to Unreal Engine 4.22
   * Recorder fixes:
     - Actors at start of playback could interpolate positions from its current position instead than the recorded position, making some fast sliding effect during 1 frame.
--- a/Docs/configuring_the_simulation.md
+++ b/Docs/configuring_the_simulation.md
@@ -81,7 +81,7 @@ environment variable `DISPLAY` to empty
 
 ```sh
 # Linux
-DISPLAY= ./CarlaUE4.sh
+DISPLAY= ./CarlaUE4.sh --use-opengl
 ```
 
 This launches the simulator without simulator window, of course you can still
--- a/Docs/configuring_the_simulation.md
+++ b/Docs/configuring_the_simulation.md
@@ -76,12 +76,12 @@ environment variable `DISPLAY` to empty
     **DISPLAY= only works with OpenGL**<br>
     Vulkan is now the default graphics API used by Unreal Engine and CARLA on
     Linux. Unreal Engine currently crashes when Vulkan is used when running off-screen.
-    Therefore the --use-opengl flag must be added to force the engine to use OpenGL instead.
+    Therefore the -opengl flag must be added to force the engine to use OpenGL instead.
     We hope that this is issue is addressed by Epic in the near future.
 
 ```sh
 # Linux
-DISPLAY= ./CarlaUE4.sh --use-opengl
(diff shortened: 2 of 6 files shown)


nsubiron:
Is vulkan default in the release version too?

wattsavolt:
You mean the release version of Unreal? Vulkan is default for Linux since the release version of Unreal 4.21. However our project settings had excluded Vulkan from the list of supported graphics APIs so it would never be used.

wattsavolt:
Looks like Unreal chooses the RHI implementation at the start of the application. Vulkan will be chosen by default by the release version but the user can also do ./CarlaUE4.sh -opengl to run with OpenGL instead.

nsubiron:
I've been testing the build with Vulkan enabled.

To install Vulkan in Ubuntu, the following seems to be enough

    sudo apt-get install libvulkan1

The performance is impressively better, specially with small images (twice FPS is some cases)
- [benchmark_opengl.log](https://github.com/carla-simulator/carla/files/3191780/benchmark_opengl.log)
- [benchmark_vulkan.log](https://github.com/carla-simulator/carla/files/3191782/benchmark_vulkan.log)

with big images, the fact that the current method makes 2 copies of the buffer (instead of one with OpenGL) probably makes the difference.

Unfortunately, I'm not able to run the build without display. Setting `DISPLAY=` ends up in crash, running from a terminal without display it complains as if vulkan is not installed. This is sad, it's a blocker for us.

wattsavolt:
I tried running the release. The -vulkan flag is not necessary because vulkan is default and without the DISPLAY=, it works the same as using the -opengl flag.

wattsavolt:
___
*[Util/BuildTools/BuildCarlaUE4.sh, line 59 at r4](https://reviewable.io/reviews/carla-simulator/carla/1639#-LgXn-QKDjl1tahk_8ng:-LgXodMzDQLtMEWIldSf:b-z12bsd) ([raw file](https://github.com/carla-simulator/carla/blob/0ac426f889ea47a1ff71707221154ccf6c963e74/Util/BuildTools/BuildCarlaUE4.sh#L59)):*

I would call the flag just `-opengl`, for consistency with UE4
</blockquote></details>

I used the -- for consistency with the other arguments. Should it be --opengl?

PR #1692, opened 2019-05-29
fix calculation of forward vector

xmyqsh:
Reference: https://i0.wp.com/slideplayer.com/4241728/14/images/34/Roll+Pitch+Yaw+The+rotation+matrix+for+the+following+operations%3A+X+Y+Z.jpg
and
https://wiki.unrealengine.com/Blueprint_Rotating_Movement_Component

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1692)

### DIFF
--- a/LibCarla/source/carla/geom/Math.cpp
+++ b/LibCarla/source/carla/geom/Math.cpp
@@ -115,7 +115,7 @@ namespace geom {
     const float sp = std::sin(ToRadians(rotation.pitch));
     const float cy = std::cos(ToRadians(rotation.yaw));
     const float sy = std::sin(ToRadians(rotation.yaw));
-    return {cy * cp, sy * cp, sp};
+    return {cy * cp, sy * cp, -sp};
   }
 
 } // namespace geom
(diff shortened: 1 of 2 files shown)


xmyqsh:
@nsubiron There is no left-hand side or right-hand side in respect of transformation or rotation by matrix or quaternion.

There is only have left-hand side or right-hand side system when we talk about Euler.
https://i0.wp.com/slideplayer.com/4241728/14/images/34/Roll+Pitch+Yaw+The+rotation+matrix+for+the+following+operations%3A+X+Y+Z.jpg
This link's formulation applys all rotation matrix calculation from Euler with intrinsic sequence ZYX, no matter in left hand side or right hand side system.

https://wiki.unrealengine.com/Blueprint_Rotating_Movement_Component
This link tells us that the Euler angle in the UE4 who use the left-hand side system is [-roll, -pitch, yaw]

The purpose of this PR is to clarify the meaning of the code.
The principle behind these lines of code applies to all related works.

nsubiron:
Sorry, I don't see how this makes the code clearer, if anything the opposite cause we're already using this convention in other parts of the code.

Issue #1125, opened 2019-01-11
Error when importing tensorflow before carla


pawel-ziecina:
Hi,
I noticed serialization (?) error when importing tensorflow before carla.

Code to reproduce:
```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import tensorflow as tf
import carla
print(carla.Location())
```

Output:
```
Traceback (most recent call last):
  File "./foo.py", line 7, in <module>
    print(carla.Location())
RuntimeError: std::bad_cast
```

When move carla import before tensorflow it works ok.

OS: Ubuntu 18.04
carla: 0.9.2 official build
tensorflow: tensorflow-gpu==1.12.0

coreper:
I have a similar problem with Pytorch 1.0. I have to reinstall pytorch 0.4 to solve this problem.

kk2491:
I am facing the same issue when I try to do the below lines of code 

```
>>> import carla
>>> 
>>> client = carla.Client('localhost', 2000)
>>> client.set_timeout(10.0) 
>>> 
>>> world = client.get_world()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: std::bad_cast
>>> 
>>> world = client.get_world()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: std::bad_cast
>>> 
```

Thank you

pawel-ziecina:
Still observing `std::bad_cast` error with below snippet on 0.9.4 official build with tensorflow-gpu==1.12.0

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

import tensorflow as tf
import carla
print(carla.Location())
```
@kk2491 could not reproduce your issue:

```python
import carla
client = carla.Client('localhost', 2000)
client.set_timeout(10.0) 
world = client.get_world()
```

kk2491:
I was able to fix that issue, but I don't remember what exactly the problem was.

Can you try this.. please check if there is any previous version of Carla installed in your system? I remember like I had to delete the older version which I built from source.

Let me know if it works.

Thank you,
KK

pawel-ziecina:
I haven't got any reference to other carla release in my system. Also above snippet was checked on other workstations.

@kk2491 Just check my snippet and let me know if it fails or not. I think it is bug not related with environment.

If needed my venv preparation:

```shell
cd <carla_release_dir>
python3 -m venv .venv
source .venv/bin/activate
[...]

PR #1730, opened 2019-06-08
update python __str__ wrapper

xmyqsh:
Solve the problem of `RuntimeError: std::bad_cast` described here: https://github.com/carla-simulator/carla/issues/1125#issuecomment-500085292

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1730)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
   * API extension: waypoint's `junction_id` that returns de OpenDrive identifier of the current junction
   * API extension: add gamma value as attribute to RGB camera
   * API change: deprecated waypoint's `is_intersection`, now is `is_junction`
+  * API update: solve the problem of RuntimeError: std::bad_cast described here: #1125 (comment)
   * Removed deprecated code and content
   * New recorder features:
     - Added optional parameter to show more details about a recorder file (related to `show_recorder_file_info.py`)
--- a/PythonAPI/carla/source/libcarla/Blueprint.cpp
+++ b/PythonAPI/carla/source/libcarla/Blueprint.cpp
@@ -17,12 +17,11 @@ namespace sensor {
 namespace data {
 
   std::ostream &operator<<(std::ostream &out, const Color &color) {
-    std::ostringstream oss;
-    oss << "Color(" << std::to_string(int(color.r))
-        << ',' << std::to_string(int(color.g))
-        << ',' << std::to_string(int(color.b))
-        << ',' << std::to_string(int(color.a)) << ')';
-    return out << oss.str();
+    out << "Color(" << std::to_string(color.r)
+        << ',' << std::to_string(color.g)
+        << ',' << std::to_string(color.b)
+        << ',' << std::to_string(color.a) << ')';
+    return out;
   }
 
 } // namespace data
--- a/PythonAPI/carla/source/libcarla/Blueprint.cpp
+++ b/PythonAPI/carla/source/libcarla/Blueprint.cpp
@@ -17,11 +17,12 @@ namespace sensor {
 namespace data {
 
   std::ostream &operator<<(std::ostream &out, const Color &color) {
-    out << "Color(" << int(color.r)
-        << ',' << int(color.g)
-        << ',' << int(color.b)
-        << ',' << int(color.a) << ')';
-    return out;
+    std::ostringstream oss;
+    oss << "Color(" << std::to_string(int(color.r))
+        << ',' << std::to_string(int(color.g))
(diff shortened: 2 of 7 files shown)


xmyqsh:
@germanros1987 Please review and approve this PR for 
https://github.com/carla-simulator/scenario_runner/commit/d2858373eeffe24eb59ec82b1f36aa513095888b

xmyqsh:
@nsubiron 
I seems that this line
```
.L.str.1:
        .asciz  "%d"
```
in assembly language which is generated by `std::to_string` in necessary.

And the `std::ostringstream` provides lots of condition checking into the assembly language.

-------
The following is the regression test: 
Case 1: the original code snip
```
 19   std::ostream &operator<<(std::ostream &out, const Color &color) {
 20     out << "Color(" << int(color.r)
 21         << ',' << int(color.g)
 22         << ',' << int(color.b)
 23         << ',' << int(color.a) << ')';
 24     return out;
 25   }
```
```
>>> import networkx
>>> import sys, os, glob
>>> sys.path.append(glob.glob('../carla/dist/carla-0.9.5-py2.7-linux-x86_64.egg')[0])
>>> import carla
>>> print carla.Color()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: std::bad_cast
>>>
```
Case 2: with `std::ostringstream` without `std::to_string`
```
   19   std::ostream &operator<<(std::ostream &out, const Color &color) {
~  20     std::ostringstream oss;
+  21     oss << "Color(" << int(color.r)
   22         << ',' << int(color.g)
   23         << ',' << int(color.b)
   24         << ',' << int(color.a) << ')';
~  25     return out << oss.str();
   26   }
```
```
>>> import networkx
>>> import sys, os, glob
>>> sys.path.append(glob.glob('../carla/dist/carla-0.9.5-py2.7-linux-x86_64.egg')[0])
>>> import carla
>>> print carla.Color()
Color(
>>>
```
Case 3: with `std::ostringstream` with `std::to_string`
```
   19   std::ostream &operator<<(std::ostream &out, const Color &color) {
~  20     std::ostringstream oss;
~  21     oss << "Color(" << std::to_string(int(color.r))
~  22         << ',' << std::to_string(int(color.g))
~  23         << ',' << std::to_string(int(color.b))
~  24         << ',' << std::to_string(int(color.a)) << ')';
+  25     return out << oss.str();
   26   }
```
```
>>> import networkx
>>> import sys, os, glob
>>> sys.path.append(glob.glob('../carla/dist/carla-0.9.5-py2.7-linux-x86_64.egg')[0])
>>> import carla
>>> print carla.Color()
Color(0,0,0,255)
>>>
```

xmyqsh:
@nsubiron
I have got another solution. I will double check it and make a new PR.

nsubiron:
@xmyqsh I'm also investigating this now, for me `networkx` works fine, but importing `tensorflow` breaks the ostream. [This SO question](https://stackoverflow.com/q/4362402) seems the same problem, it could be what they suggest, when the dynamic library is loaded is not properly initializing the streams to format numbers.

nsubiron:
So if the problem is parsing numbers, we can use `std::to_string` when parsing numbers, only with that should do it.

xmyqsh:
@nsubiron You are right. It is a problem with parsing numbers on stream which appears in Boost.Python in 2012 as described in the [SO question](https://stackoverflow.com/questions/4362402/what-is-wrong-with-c-streams-when-using-boost-python). And this problem has been fixed in Boost.Python later in 2012 as `Neil G` said finally. Although we don't know what is the problem exactly and how Boost.Python fixed it.

Now, let's move on our carla PythonAPI. It works well when stands alone. But, when importing some package or loading some dynamic library, eg. networkx, tensorflow, before it, the old parsing numbers on stream problem appears again which same as in 2012.

As far as I know tensorflow us SWIG and ctypes instead of Boost.Python. So this problem cannot be fixed even we update the tensorflow to latest version. This is a very lower level feature which loaded dynamically that only Boost.Python uses it and others do not support.

The solution should be: 1. make others support this feature(not possible); 2. enhance the parsing numbers on stream patch in Boost.Python(do not know it is possible or not)

So, the best solution is to support the function of parsing numbers on stream by ourselves explicitly.
And `std::string` should be the minimum part necessary and best solution currently. Although it also generates about 4x more code https://godbolt.org/z/OqPuzu.

nsubiron:
Great, yes it's a good compromise. In any case, printing is inherently slow, it's not critical to have this part a bit less optimal.

PR #1736, opened 2019-06-11
Caching rendered towns to make next executions faster in no rendering mode

manishthani:
#### Description

*No rendering mode* can save and load cached towns images in order to make the rendering of maps faster. 
In the first execution of the `no_rendering_mode.py` script for a specific town, the execution will be slow since it needs to render the selected town image, and when rendering finishes it will also save the rendered image inside `./PythonAPI/examples/cache/no_rendering_mode/` directory. Therefore, in further executions of the script, it will **not** render the map again, instead it will just read the previously rendered image of the town, saving users a lot of time when running the script. 

Note that *No rendering mode* will update the corresponding cached town image if its OpenDrive file content changes.

Finally, refactored some code to make it more consistent with other other `PythonAPI` example scripts  .

  * **Platform(s):** Ubuntu 16.04LTS
  * **Python version(s):**  2.7.12 and 3.5.2

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1736)


wattsavolt:
___
*[PythonAPI/examples/no_rendering_mode.py, line 1004 at r1](https://reviewable.io/reviews/carla-simulator/carla/1736#-LhfXtV1CGVQZu82v-tW:-LhfXtV1CGVQZu82v-tX:b-xdzk7u) ([raw file](https://github.com/carla-simulator/carla/blob/64b763b00566b5fde2654727f3f3015f57210417/PythonAPI/examples/no_rendering_mode.py#L1004)):*
> ```Python
>         ]
> 
>         info_text = info_text
> ```

Seems like a waste of an assignment?

PR #1745, opened 2019-06-13
Bernatx/navigation

bernatx:
#### Description

We have added support to spawn pedestrians that will walk at random points continually. We don't have a full API for them yet, but we will.

Using the example script we can spawn 100  pedestrians with the following command:

./spawn_pedestrians.py -n 100

  * **Platform(s):** Ubuntu 16.4, Windows 10
  * **Python version(s):** 3.7

#### Possible Drawbacks

Currently for navigation calculations (path finding) every scenario need a binary file generated by hand with the recast library. We hope to automatize this soon.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1745)


wattsavolt:
___
*[PythonAPI/examples/spawn_pedestrians.py, line 143 at r2](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhK11J30L2kHeupXABf:-LhKTN861UA6TkXGK_FS:bj570j5) ([raw file](https://github.com/carla-simulator/carla/blob/bd9dbe437b476d5682748cc37101e92ae8bdf93b/PythonAPI/examples/spawn_pedestrians.py#L143)):*
<details><summary><i>Previously, manishthani wrote…</i></summary><blockquote>

Maybe this try catch is not needed?
</blockquote></details>

Looks like it prevents keyboard interrupt exceptions from being thrown

wattsavolt:
___
*[LibCarla/source/carla/client/detail/WalkerNavigation.h, line 85 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhjCcr84I6RXaj-41dR:-LhjCcr84I6RXaj-41dS:bmv6526) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/LibCarla/source/carla/client/detail/WalkerNavigation.h#L85)):*
> ```objc
>   private:
> 
>     Client &_client;
> ```

Not good to use references as class members. Could this be changed to a pointer instead?

wattsavolt:
___
*[PythonAPI/examples/spawn_npc.py, line 174 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhjEXL3B0vazdqSDewj:-LhjEXL3B0vazdqSDewk:b-b60n61) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/PythonAPI/examples/spawn_npc.py#L174)):*
> ```Python
>             all_actors[i].set_max_speed(1 + random.random())    # max speed between 1 and 2
> 
>         print('spawned %d walkers, press Ctrl+C to exit.' % len(info))
> ```

Do we have to be compatible with python 2.5? If we are targeting python 2.7 and higher, then it's cleaner to use .format on strings instead of %. print('spawned {} walkers, press Ctrl+C to exit.'.format(info))

wattsavolt:
___
*[LibCarla/source/carla/client/detail/WalkerNavigation.h, line 85 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhjCcr84I6RXaj-41dR:-LhjK09fAeQAXYosv2ee:be0i2y7) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/LibCarla/source/carla/client/detail/WalkerNavigation.h#L85)):*

I think that in this case because a client is a must for this object to work, we can use a reference and force the creation of this object with a client object. Then we don't need to check for null client object. Is not that an advantage?
</blockquote></details>

It's just rule of thumb not to use references as members usually but if its better in this case, then all good :)

wattsavolt:
___
*[PythonAPI/examples/spawn_npc.py, line 174 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhjEXL3B0vazdqSDewj:-LhjKQjp8f61Y5_-4LSQ:bjjdqds) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/PythonAPI/examples/spawn_npc.py#L174)):*

For simplicity I'm still using that way because it is used in all .py scripts we have so far. Compatibility with old python versions is always a plus :)
</blockquote></details>

It's only a small thing. Just asking because I see this in all the python files. I think its better to take advantage of new features in the language. I don't think people should be using less than 2.7.

wattsavolt:
___
*[LibCarla/source/carla/client/detail/WalkerNavigation.h, line 85 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-LhjCcr84I6RXaj-41dR:-LhjPJTv3Y2ffz3eJI2L:b6mq1c) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/LibCarla/source/carla/client/detail/WalkerNavigation.h#L85)):*

I agree with Bernat, pointers and references are basically the same, both are equally dangerous in terms that the object could be destroyed and we wouldn't notice, the advantage of references is that they cannot be set to null.

In this case, the Client object belongs to the Simulator object, the Simulator should never be destroyed as long as any object in the API is still alive. I think that's always the case, but... this is a potential source of crashes. Maybe we can reconsider and shared_ptr the client, but we need to check the whole client part to avoid circular dependencies.
</blockquote></details>

I remember looking up this a while ago and found a link on stackoverflow discussing the pointer vs reference member issue. 
https://stackoverflow.com/questions/892133/should-i-prefer-pointers-or-references-in-member-data 

Yeah shared_ptr would be best if circular dependency could be avoided

analog-cbarber:
> Do we have to be compatible with python 2.5? If we are targeting python 2.7 and higher, then it's cleaner to use .format on strings instead of %. print('spawned {} walkers, press Ctrl+C to exit.'.format(info))

I really don't think the string formatting technique really matters. I don't see why using format is "cleaner" than the old % technique, although it is more flexible if you want to use named arguments instead of positional formatting. I pretty much just use python 3.6 for my projects and still use % from time to time.

I do like the new f-strings introduced in python 3.6, e.g.:

~~~python
print(f'spawned {len(info)} walkers, press Ctrl+C to exit')
~~~

but you will just have to wait until you drop support for 2.7 for that. (BTW, everyone should stop using 2.7 ASAP! Many important libraries will be dropping support in the next 12 months)

wattsavolt:
It doesn't matter regarding performance but I just think it's neater cosmetically and especially in more complex use cases. If a type being printed changes from a float to decimal for example, the code will not have to be changed for the format to work. To each his own I guess.

I agree that support for python 2.X should be discontinued.

wattsavolt:
___
*[PythonAPI/examples/spawn_npc.py, line 165 at r7](https://reviewable.io/reviews/carla-simulator/carla/1745#-Lhq1IZA4L-Et5QBonVk:-Lhq1IZA4L-Et5QBonVl:br5ijnl) ([raw file](https://github.com/carla-simulator/carla/blob/1b7ae626707add1401602a9e68995d92393f283f/PythonAPI/examples/spawn_npc.py#L165)):*
> ```Python
> 
>         # initialize each controller and set target to walk to (list is [controler, actor, controller, actor ...])
>         for i in range(0, len(all_id), 2):
> ```

When I try to run the script, I'm getting an exception because the code is entering the finally clause before the all_id list is initialised. 

Traceback (most recent call last):
  File "./spawn_npc.py", line 194, in <module>
    main()
  File "./spawn_npc.py", line 185, in main
    for i in range(0, len(all_id), 2):
UnboundLocalError: local variable 'all_id' referenced before assignment

This list probably should be declared at the beginning of the try

BrandonHaynes:
I am seeing errors of the following form in this branch when launching via `CarlaUE4.sh`:
```
Failed to load Navigation Mesh file '../../../CarlaUE4/Content/Carla/Maps/Nav/Town03.bin
```

The errors are not present when launching the Carla binary directly.  `NavigationMesh.cpp` may need a fallback-directory guard similar to the one in `OpenDrive.cpp`.

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/CarlaServer.cpp, line 598 at r8](https://reviewable.io/reviews/carla-simulator/carla/1745#-LiEA3ds3cMjSVFRh-uO:-LiEA3ds3cMjSVFRh-uP:b-ck6uix) ([raw file](https://github.com/carla-simulator/carla/blob/3e377798aafe065bc5c698acb993bed4fadcca3b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/CarlaServer.cpp#L598)):*
> ```cpp
>     {
>       RESPOND_ERROR("unable to get bounding box: actor view is invalid");
>     }
> ```

This function is returning the half height of the character's bounding capsule. Might be better to reference the capsule half height instead of the bounding box in the messages.

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/CarlaServer.cpp, line 590 at r8](https://reviewable.io/reviews/carla-simulator/carla/1745#-LiEBGNRAp7-9G3nETRF:-LiEBGNRAp7-9G3nETRG:bge9jpg) ([raw file](https://github.com/carla-simulator/carla/blob/3e377798aafe065bc5c698acb993bed4fadcca3b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Server/CarlaServer.cpp#L590)):*
> ```cpp
>   };
> 
>   BIND_SYNC(get_walker_base_offset) << [this](
> ```

Not sure if this is the most descriptive name. What about 'get_walker_half_height'?

BrandonHaynes:
NB: I am unable to reliably get all walkers to navigate using this branch and the `spawn_npc.py` script.  Only a random subset actually seem to move anywhere.

ThierryDubois:
> NB: I am unable to reliably get all walkers to navigate using this branch and the `spawn_npc.py` script. Only a random subset actually seem to move anywhere.

Maybe it is unrelated, but from the latest commit I deleted the lines that take care of spawning vehicles and it seems to work for me. I just spawn vehicles in a separate script.

*I deleted from Batch = [ ] to the apply batch line.

BrandonHaynes:
> What do you mean by a 'random subset move'? 

After successfully spawning n walkers (which we agree may be less than the number requested), some of them remain stationary and never actually move anywhere.  The set and number of these frozen walkers appears to vary between runs of the script.

I can open a separate issue if you'd to not pollute this pull request.

BrandonHaynes:
The stationary walker issue resolved when I updated to the latest commit in this branch.  Apologies for the noise.  

I did have a related issue when using the walker's current location as a parameter to `controller.start()` (i.e., `controller.start(walker.get_location())`) -- this seems like it should work but does not.  I can work around it, but thought I'd mention it.

BrandonHaynes:
Another NB: I'm seeing an exception thrown in [`Navigation.cpp` ](https://github.com/carla-simulator/carla/blob/bernatx/navigation/LibCarla/source/carla/nav/Navigation.cpp#L636) after the simulation has run for about 15 minutes with many walkers.

PR #1758, opened 2019-06-17
Set up RootComponent in ASensor

nsubiron:
#### Description

Set up the `RootComponent` in `ASensor` so we don't need to set it up on each individual sensor. Forgetting to set up the root component results in difficult to track bugs (e.g. sensor not having a transform), this way we make sure every sensor has a root component. Moreover, they will be displayed in editor (though for the moment all of them look like cameras).

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1758)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/GnssSensor.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/GnssSensor.cpp
@@ -5,7 +5,6 @@
 
 #include "Carla.h"
 #include "Carla/Sensor/GnssSensor.h"
-#include "StaticMeshResources.h"
 
 FActorDefinition AGnssSensor::GetSensorDefinition()
 {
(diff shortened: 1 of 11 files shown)


wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp, line 59 at r1](https://reviewable.io/reviews/carla-simulator/carla/1758#-LhdzsLw4VGIR6j0x61o:-LhdzsLw4VGIR6j0x61p:bdw6dy4) ([raw file](https://github.com/carla-simulator/carla/blob/96b8cc1265c819a3561a5b2bd8163c661dc44d63/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp#L59)):*
> ```cpp
>   RootComponent = MeshComp;
> 
>   DrawFrustum = CreateDefaultSubobject<UDrawFrustumComponent>(TEXT("DrawFrust"));
> ```

Is the DrawFrustum not necessary? I reference this in the PR for enabling texture streaming but I can change it.

wattsavolt:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp, line 59 at r1](https://reviewable.io/reviews/carla-simulator/carla/1758#-LhdzsLw4VGIR6j0x61o:-Lheb8E70c34GlYrR5Hk:b-dk7hkt) ([raw file](https://github.com/carla-simulator/carla/blob/96b8cc1265c819a3561a5b2bd8163c661dc44d63/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/SceneCaptureSensor.cpp#L59)):*

Oh I see, I thought it wasn't used. If it's just for the location we can probably use the actor's location, it should be the same.
</blockquote></details>

Yeah I can change to use capture component location instead.

PR #1858, opened 2019-07-08
Simplified Import Assets Pipeline

manishthani:
#### Description
Simplified pipeline for importing props and maps and also about defining props in the Carla Blueprint Library. Updated documentation about these 2 features (See `export_import_dist.md` to know about the steps for importing maps or props).

  * **Platform(s):** Ubuntu 16.04LTS
  * **Python version(s):** 2.7 and 3.5

#### Possible Drawbacks
Map can be wrongly imported due to how external software export the meshes in terms of units, since Unreal uses centimetres and external tools tend to export in meters.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1858)


Docs/asset_packages_for_dist.md
```diff
@@ -109,7 +109,7 @@ Import
 │   │   ├── LaneMarking1_Spec.png
 │   │   ├── MapToImport01.fbx
 │   │   └── MapToImport01.xodr
-|   └── MapToImport02
+|   ├── MapToImport02
```
nsubiron: :+1:

PR #1914, opened 2019-07-20
doc fix: load_map -> load_world

togaen:
The documentation for loading a map with the python API references a function that doesn't exist. It seems like it should be `load_world` instead of `load_map`.

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1914)


marcgpuig:
Thanks for your PR :)

PR #1935, opened 2019-07-25
Add Responsibility Sensitive Safety (RSS) to CARLA Client library 

johannesquast:
#### Description

Adding a build variant containing a RSS Sensor, RSS result visualization, an RSS Restrictor implementation and an manual control example. The imlementation is using the AD RSS library (https://github.com/intel/ad-rss-lib).

Currently only some road structures are supported. All vehicles must be on the same road segment. Intersection handling is missing.
Integration into LibCarla and PythonAPI are optional and available through additional extended make targets.

  * **Python version(s):** Python 2.7 and Python 3.5

#### Possible Drawbacks

Additional build variant required as AD RSS library is (currently) only supported on Linux.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1935)


manishthani:
Diving into the code, we have noticed some things that would be nice if they are changed. Here is a list of what we have seen so far:

- Manual control: Would it be better to add the RSS sensor into the existing `manual_control.py`? We think that this way it is possible to avoid duplicated code as much as we can and make it more maintainable in the future. 
Also, we encountered 4 bugs: The first one is that the car gets blocked even when trying to go reverse when the sensor detects a vehicle in front (Brake input value is maximum). The second one is that rays tend to blink frequently when drawing them. The third one is that it does not work with intersections, even if you mentioned so, there should be a way to make it work. The last one is that RSS is technically a sensor, so it should not block user interaction if it detects a vehicle in front that is very close to ego vehicle, unless specified by the user with a flag. We understand that this is a 1st version of the RSS sensor, so if it is not possible to fix them now, these bugs reports are just to keep them in mind for future improvements.

- Compile option of RSS: We think that the ad-rss-lib should be always included and compiled inside Carla, so this way we won't need any special flag only for this sensor and would make things more readable in code ( no separated code, no extra `#ifdef` macros and namespace declarations). Furthermore, it would complicate Jenkins and Travis builds, because it would need to build and test Carla with and without the RSS option during pull requests and when merging to master. 

- Documentation: It would be great to change comments from `/*` and `*/` to `///` in header files and keep `//` in cpp files. By doing that, you could let Doxygen to detect it as a comment and generate documentation automatically. That's the way we are generating documentation in Carla. Would appreciate a lot to have more comments and documentation, but that is not mandatory :)

- Uncrustify: There are some files that need to be prettified (such as `RssCheck.h`). Please run uncrustify for all files.

Please let us know if you have some doubts about it.
Thank you very much for your effort and looking forward to integrating this new feature!
**Pd**: Carla team will be on holidays for the next 3 weeks, so you can take your time on this :)

johannesquast:
> Manual control: Would it be better to add the RSS sensor into the existing manual_control.py? We think that this way it is possible to avoid duplicated code as much as we can and make it more maintainable in the future.

As RSS will remain a build variant, using only one _manual_control.py_ would imply heavy variant handling.
As there is already a _manual_control_steeringwheel.py_ and parts / classes from _manual_control.py_ are existing as copies in various states/versions in various places (e.g. _World_, _HUD_, _FadingText_, ... in _automatic_control.py_ or in _no_rendering_mode.py_), it would be worth thinking about refactoring the "examples" and extract some kind of "high level" Python API.

> Also, we encountered 4 bugs: The first one is that the car gets blocked even when trying to go reverse when the sensor detects a vehicle in front (Brake input value is maximum).

The RssRestrictor now does not intervene when in reverse gear.

> The second one is that rays tend to blink frequently when drawing them.

The RSS Sensor might not process with drawing FPS. The visualization via "DebugHelper" is set to last approximately one frame. If there are no new results for the current frame there is no visualization.

> The third one is that it does not work with intersections, even if you mentioned so, there should be a way to make it work.

There is no easy way to get the road structure including intersections from CARLA to convert it into RSS world model. There is ongoing work to get this data from the OpenDRIVE maps, but this will not be part of this branch / the initial RSS submission.

> The last one is that RSS is technically a sensor, so it should not block user interaction if it detects a vehicle in front that is very close to ego vehicle, unless specified by the user with a flag. 

The example _manual_control_rss.py_ uses the RSS Sensor and the RSS Restrictor. As this example is meant to showcase RSS, the restrictor should be enabled by default.

> Compile option of RSS: We think that the ad-rss-lib should be always included and compiled inside Carla

RSS is currently only available for Linux. We don't have the resources (technical and personal) to support Windows build, so the build variant needs to stand.

> Documentation: It would be great to change comments from /* and */ to /// in header files and keep // in cpp files

Done.

> Uncrustify: There are some files that need to be prettified

Done.

Kingsman1995:
hello, when i install the carla-feature-rss, an error just happened :+1: 
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Util/BuildTools/Linux.mk:125: recipe for target 'setup' failed
make: *** [setup] Error 128
Could u please tell me how to fix it? i've already tried git init command, but still not work.

jackbart94:
Sorry if it's been a while. I have tested it again and analyzed the code more in depth, and I still have some concerns. I've added only two small comments related to code, since most of the issues can be found in different files.

1. The visualization of the RSS is truly amazing, yet the frequent blinking makes it so hard to look at for more than 5 seconds. Is it in any way possible to avoid the lines appearing and disappearing so often?

2. Give users the possibility of turning RSS Restrictor on or off. I don't know if it's a lot of work, but maybe people want to use the RSS only as a sensor. I know this is supposed to showcase the whole system but a small change now would make a difference.

3. Overall, comments. Code needs to be commented more, and English has to be checked since there are several errors. Shouldn't take long.

4. We use "_", not "m" before private variables, it would be nice if you could change them.

5. Use "namespace cc = carla::client", and also for others, as it improves readability.

6. Include pedestrians in this, if possible.

7. Either do the TODOs, or remove them and leave a comment saying how that could be improved, otherwise this will always be considered as not finished, even for a first implementation.

8. The whole method to check for the obstacle and ego vehicle's road and lane seems odd (RssCheck.cpp, checkObjects), and this is visually seen when testing (for example, car braking because of nearby cars that however are behind it, or also considering the distance of cars in the opposite road as possible threats).

PR #1951, opened 2019-07-31
Implementation of the new traffic manager

jackbart94:
#### Description

A new implementation for the traffic manager has been developed. There are several modules, all of them containing a description of the functionalities they provide.

  * **Python version(s):** 2.7.12 and 3.5.2

#### Possible Drawbacks

Since it is not using the trigger boxes of the autopilot, some changes have to be made on different maps in order for it to avoid running the following traffic lights on a red semaphore:
- Town03, BP_TrafficLight01, the bounding box for the semaphore has to be moved out of the junction.
- Town04, AmericanLights_Town04_03, the bounding box for the semaphore has to be moved out of the junction.

For a similar reason, in Town03, the junction on SM_RoadTown03_116 is correctly managed only for the autopilot, and should be modified to give it a better regulation.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1951)

### DIFF
--- a/LibCarla/source/carla/traffic_manager/README.md
+++ b/LibCarla/source/carla/traffic_manager/README.md
@@ -1,0 +1,35 @@
+# Traffic Manager
+
+## Installing Traffic Manager
+
+First of all, make sure that CARLA has already been installed, following the instructions on the documentation, and that you have built the different modules:
+
+```
+
+make launch
+make PythonAPI
+make package
+make LibCarla
+
+```
+
+Then, after downloading the scenario runner repository and extracting it, on the CMakeLists.txt file in ${TRAFFIC_MANAGER}/source you must set an environment variable, LIBCARLA_LOCATION, to where your build of LibCarla resides. After this, you can do the following to build the traffic manager:
+
+```
+
+mkdir build
+cd build
+cmake..
+make
+
+```
+
+## Using Traffic Manager
+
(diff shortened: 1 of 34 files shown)


manishthani:
Diving into the code, I have noticed some things that would be nice if they are changed. Here is a list of what I have seen so far:
- Passing vectors by copy: I can understand that maybe in some cases you need to make a full copy of an array, but in most cases it is way more efficient to pass containers and other data structures by reference. 
- Const correctness: Class methods such as getters are not marked as `const` ( by doing so, it means that those functions will not modify the members of that class). Furthermore, there are no `const` parameters in any class method. This applies to all the methods that do not modify the members of a class. 
- `.hpp` to `.h`: All header files in Carla have the extension `.h`. I do not know the reason why it could be better or not, but for consistency reasons with Carla, it would be better to change it to that extension. 
- Documentation: Really appreciate that the code is quite well documented. The only thing that would be great is to change `/*` and `*/` to `///`. By doing that, you could let Doxygen could detect it as a comment and generate documentation automatically. That's the way we are generating documentation in Carla.
- Uncrustify: Please use the uncrustify tool to make the code look pretty for all the `.cpp` and `.h` files. I think that the command inside Carla is `make pretty -f filename`.
- White space trailing: Use any tool to remove unnecessary white spaces automatically when saving a file (such as the white spaces that exceed after writing a line of code or the ones between lines).
 
Please let me know if you have some doubts about it.
Thank you very much ! :)

xmyqsh:
@jackbart94 Are you caffe fans?

PR #1984, opened 2019-08-08
Add german / international keyboard support to manual_control.py

cassfalg:
manual_control.py uses the backquote key ("`") to change the used sensor. That is an issue on German keyboard layouts (and potentially a number of other international layouts) since there is no backquote key in this layout (it's not a single keypress, you need to press a key and another key to create a backquote character). I propose to use the "n" key instead, which is not used so far and which I assume is produced by a single key press on a lot if not all layouts.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/1984)


cassfalg:
Is this merge request missing something from my side? It's been a while, and it's a really small change, probably like a 60 second review...?

nsubiron:
Personally I would keep both "`" and "n" to change the sensor.

cassfalg:
> Personally I would keep both "`" and "n" to change the sensor.

I don't really care what the actual key is, honestly, or how many there are -- as long as it is available on my keyboard. Do you want me to make a patch on top that adds in the old key as well?

nsubiron:
It was just a suggestion, that key is conveniently placed in English keyboards. In any case this PR looks good to me.

jackbart94:
> > Personally I would keep both "`" and "n" to change the sensor.
> 
> I don't really care what the actual key is, honestly, or how many there are -- as long as it is available on my keyboard. Do you want me to make a patch on top that adds in the old key as well?

If you could do that, I'll test it and merge very quickly - sorry for the delay!

cassfalg:
> > > Personally I would keep both "`" and "n" to change the sensor.
> > 
> > 
> > I don't really care what the actual key is, honestly, or how many there are -- as long as it is available on my keyboard. Do you want me to make a patch on top that adds in the old key as well?
> 
> If you could do that, I'll test it and merge very quickly - sorry for the delay!

Here you go. I hope.

PR #2035, opened 2019-08-26
Fixing -no-rendering command line argument

jit-rc:
#### Description

-no-rendering in command line did not work before

Fixes #  -no-rendering command line argument

  * **Platform(s):** 18.04
  * **Python version(s):** 3.6.8
  * **Unreal Engine version(s):** 4.22.2

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2035)


marcgpuig:
Hi @jit-rc, thanks for your PR.
@lassepe is right, this functionality is not intended to be used anymore, but if there is a mismatch we must fix it anyway.
But why is not consistent? In the same file, you can see the other options and they that start with `-`.

jit-rc:
All the command-line line arguments passed during running of the server ./CarlaUE4.sh, are single `-` and that was also mentioned in the documentation when `no-rendering` was officially supported. In the code however `FParse::Param` adds a `-` making the argument `--` but `FParse::Value` does not add anything keeping the arguments with single `-`.  Thats why it looks consistent in code but outcome of what argument to be passed in is inconsistent.

Issue #2038, opened 2019-08-26
Synchronous_mode gets stuck in endless loop


jitrc:
SynchronizeFrame sometime does not get the correct GFrame Counter
and thus keeps waiting  in world.tick()

![Screenshot from 2019-08-26 01-04-22](https://user-images.githubusercontent.com/660504/63675682-33d70300-c79e-11e9-9aae-0016552fb82d.png)

Lot of additional messages added to debug the situation, that is what is in the image.

jit-rc:
This is our intermediate solution 
```c++
void FCarlaEngine::OnPostTick(UWorld *, ELevelTick, float DeltaSeconds)
{
  uint32_t RunSomeCount = 0u;
  do
  {
    Server.RunSome(10u);
    // TODO: fix bug and remove
    // To solve for bug of Client not receiving new Frame and stuck forever in SyncronizeFrame
    // We are repeatedly sending same frame, only in SynchronousMode, and waiting for next tick
    // It only happens in Sync mode and faster than real-time speeds 
    // and we need to send Frames repeatedly fast enough to maintain average performance 
    RunSomeCount++;
    if (bSynchronousMode && RunSomeCount % 100 == 0)
     WorldObserver.BroadcastTick(*CurrentEpisode, DeltaSeconds);
  }
  while (bSynchronousMode && !Server.TickCueReceived());
}

jit-rc:
Seems like the culprit is `session 0: connection too slow: message discarded` , data streaming is dropped instead of being queued.

nsubiron:
> session 0: connection too slow: message discarded

I see, that causes a dead-lock between client and server, but not sure how to fix it.

Definitively not very nice the way the simulator drops messages when the connection is slow, but having a queue also has its own problems.

Nevertheless, there is always a possibility some message is lost, in that case the client ends up waiting forever. I guess the best would be adding a timeout to the image queue in Python, say if the image hasn't arrived in X seconds, skip it and go on to the next tick (and make sure you ignore images with older frame numbers in case the image eventually arrives).

jit-rc:
In our case we are not even using image and operating in no-rendering mode, that's how we are getting those high FPS and hitting this issue. We are only waiting for new Frame no. to be avaialbe WorldObserver.BroadcastTick(in Server) as SynchronizeFrame (in Client) waits.

```cpp
static void SynchronizeFrame(uint64_t frame, const Episode &episode) {
[...]

PR #2070, opened 2019-09-04
Improve sync mode performance

jit-rc:
#### Description

High FPS synchronous_mode

 * TCP no_delay option enabled to allow high fps, otherwise sometimes get stuck in 20 FPS, due to 40ms network delay for small packaets
 * Avoids Deadlock: https://github.com/carla-simulator/carla/issues/2038
 * Max FPS benchmarking client utilities in Cpp and Pyhton
 * Small productivity improvements: run server without capturing mouse in Viewport

Fixes #  Avoids Deadlock in high speed synchronous_mode

  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** 4.22.2

#### Possible Drawbacks

Current mechanism of avoiding deadlocks is hacking, but did not affect maximum FPS possible for the server and client. Still it might spam the network more and use more CPU.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2070)

### DIFF
--- a/Examples/CppClient/Makefile
+++ b/Examples/CppClient/Makefile
@@ -21,8 +21,9 @@ clean:
 run: build
 	$(call log,Running C++ Client...)
 	@$(BINDIR)/cpp_client $(ARGS)
+	@$(BINDIR)/benchmark_max_fps $(ARGS)
 
-build: $(BINDIR)/cpp_client
+build: $(BINDIR)/cpp_client $(BINDIR)/benchmark_max_fps
 
 $(BINDIR)/cpp_client: build_libcarla
 	$(call log,Compiling C++ Client...)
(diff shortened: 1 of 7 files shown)


nsubiron:
Rather than adding this workaround I would address the source of the problem. Specially I see problematic removing the callbacks call in the client-side, that could increase a lot the amount of frames that are dropped on async mode.

So the problem is that the function that writes data to the buffer simply drops the data if the socket is busy; this is bad in general but specially critical in sync mode because it causes a dead lock. In the past we didn't add a queue of messages because if the client is slow the queue would grow indefinitely and we don't want that, so dropping messages was "good enough" (there was no sync mode). But we could add a fixed-size queue (I don't know, 2 or 3 messages), that fixes the dead lock in sync mode and also would improve the rate messages are dropped in async mode.

Other points are good, but since they're not dependent on each other perhaps you could split the PR, this way we can discuss and approve changes separately.

jit-rc:
@nsubiron I didn't expect this PR to be merge, posted it to keep this option open and available to other users who might be face the same problem, till we the queue implemented. 

You can also manually take other improvements like `TCP no_delay`

nsubiron:
Alright I see, yes it's good to have this somewhere :smile:

PR #2112, opened 2019-09-17
Updated TrafficManager

pravinblaze:
#### Description

This is an implementation of client side multi-vehicle controller with capabilities of following the road, avoiding collisions with other vehicles, respecting traffic rules and changing lanes when appropriate to simulate realistic traffic.

P.S. See carla/TrafficManager/README.md for details on how to build and use.

  * **Unreal Engine version(s):** 4.22.1

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2112)


jackbart94:
We've started the in-depth review, please review and modify the in-depth comment that we made. Adding to that, we also have a few more general suggestions that can improve the code quality:

1. Other than explaining what each .h file does, please comment variables (using ///) since it's important to know what every variable is in charge of.
2. We initially suggested `typdef` (our bad!) but we actually think that `using` is better.
3. There is a way to batch-spawn vehicles instead of doing it one by one (same with delete)
4. Revise all the variables and data structures. We noticed that some of them might be local instead of members (see comment on `planner_frame_a` for example). Also, make sure there are no leftover variables from the previous design (we've found some and pointed them out)
5. Use carla/Logging.h instead of normal prints for messages/debugging.
6. Provide an explanation to tell us why some complex data structures have been used and why are they more efficient than others (one example could be the unordered_map with boolean keys)
7. Sometimes it's useful to have comments in the .cpp files as well, comment briefly what's being done especially when a lot of operations are involved.
8. Some variable names should be more specific.
9. Please, make sure to check carefully everything that is already developed inside the CARLA library. This is important for several reasons, one because you can mimic the implementation and understand more the coding standards, and two because many methods are already there and you could just use them instead of recreating those.
10. Some variables (such as `localization_frame`) are shared pointers of a vector. Is it possible that they could just be vectors?

jackbart94:
Adding to the already present comments, as general rules:
1) Check if all includes are actually used in the class.
2) If floats, put f after the number, if unsigned int put u.
3) Use auto keyword only when needed, don't abuse of it. There's an article with some ground rules in the review to help you out.
4) Why did you create SimpleWaypoint.h instead of using the Waypoint class? For example, setLeftWaypoint isn't actually checking that the Waypoint being set is really on the left. If possible, use the Waypoint class.

pravinblaze:
___
*[TrafficManager/source/pipeline/CollisionStage.cpp, line 287 at r4](https://reviewable.io/reviews/carla-simulator/carla/2112#-LozuKve8d3JfzXVRxJP:-LpRjTr9CfbbRDfQ2Ckq:bgytv13) ([raw file](https://github.com/carla-simulator/carla/blob/35a55068ec3440963cce7024812006410dfa69c8/TrafficManager/source/pipeline/CollisionStage.cpp#L287)):*
<details><summary><i>Previously, manishthani wrote…</i></summary><blockquote>

Ok, waiting for your response.
</blockquote></details>

We saw Transform::TransformPoint. I see that the method is designed to receive a reference to an input vector, modify it in-place by rotating it along the orientation of the transform object and shifting its base onto the transform object's location vector.

We could use it to place the boundary corner vectors (ex: x_boundary_vector + y_boundary_vector) on to the vehicle's position vector using Transform::TransformPoint but I'm not sure if we should do it because we will need to create a transform object with the vehicle's position & y,r,p = 0, create names for all the four corner vectors, transform and then use them. It's just easier to do one addition. Maybe you were thinking about a different way to use Transform::TransformPoint that can help us?

pravinblaze:
___
*[TrafficManager/source/pipeline/CollisionStage.cpp, line 289 at r4](https://reviewable.io/reviews/carla-simulator/carla/2112#-Lozse6KCFuHcZ4_iigX:-LpRnaA95i-X1nq45h1T:brhvk9y) ([raw file](https://github.com/carla-simulator/carla/blob/35a55068ec3440963cce7024812006410dfa69c8/TrafficManager/source/pipeline/CollisionStage.cpp#L289)):*
<details><summary><i>Previously, manishthani wrote…</i></summary><blockquote>

> We will look into this
Ok, waiting for your response.
</blockquote></details>

We saw that FDebugShapeDrawer::Draw can draw a carla::rpc::DebugShape. carla::rpc::DebugShape only has a few simple shapes like line, arrow, box, etc. We need to draw a more complex tracing the curved path the vehicle is going to take.

Also, I see that DebugShapeDrawer.h is in carla/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Util/ and not present in the after you build LibCarla. DebugShape.h is available in LibCarla.

pravinblaze:
___
*[TrafficManager/source/pipeline/CollisionStage.cpp, line 259 at r4](https://reviewable.io/reviews/carla-simulator/carla/2112#-LozkORd-ORvSCKZZ7Cu:-LpRjE8u6yk9CXrHBu-Q:b-o7bm0u) ([raw file](https://github.com/carla-simulator/carla/blob/35a55068ec3440963cce7024812006410dfa69c8/TrafficManager/source/pipeline/CollisionStage.cpp#L259)):*
<details><summary><i>Previously, manishthani wrote…</i></summary><blockquote>

Ok, waiting for your response.
</blockquote></details>

Done. We changed the vector ordering around. Have a look, the current logic should be more intuitive.

pravinblaze:
> Adding to the already present comments, as general rules:
> 
> 1. Check if all includes are actually used in the class.
> 2. If float, put f after the number, if unsigned int put u.
> 3. Use auto keyword only when needed, don't abuse of it. There's an article with some ground rules in the review to help you out.
> 4. Why did you create SimpleWaypoint.h instead of using the Waypoint class? For example, setLeftWaypoint isn't actually checking that the Waypoint being set is really on the left. If possible, use the Waypoint class.

Hey @jackbart94, we wanted to create a client-side, discrete cache of Carla's continuous map. Since asking for next, left and right waypoints on carla::client::Waypoint is expensive, we wanted a container that can hold Carla's waypoint and connect them up to form a representation of the map. This way we can quickly access the map without communicating with Carla. The class InMemoryMap is responsible for placing all the right connections. Maybe we should consider having a way to ensure that no one else can modify the map once it is created.

PR #2148, opened 2019-10-01
Local planner PID controller

felipecode:
#### Description

We found out that the previous control was overly spiky.
image
The new parameters
image
Thanks to @ykarmesh for this finding and contributions.
The new agent is much easier to imitate through behavior cloning.

  * **Platform(s):** Linux, CARLA 0.96
  * **Python version(s):** latest
  * **Unreal Engine version(s):** latest

#### Possible Drawbacks

I don't know the extension in which the basic agent is used but this might create problems.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2148)


ZhangYouRong:
thanks a lot that realy help

PR #2182, opened 2019-10-23
Return HitDistance instead of HitRadius

### LINKED ISSUE #1952: obstacle sensor distance problem
Dear Carla Developers, @bernatx  @nsubiron @felipecode @TheNihilisticRobot ..
Some friends have checked the codes of ObstacleSensor.cpp, the problem that this sensor is not showing the correct distance to detected obstacle is that, instead of distance, it brings Radius values. and you could see the related line of code here in this link: https://github.com/irgendwie/carla/commit/ae6ba4de83e031c65ea96d773c6e223f93752b3b
After we changed this to the code and re-compiled Carla, it is fortunately working correctly. However, I want to ask you is it possible that you correct this code and re-compile Carla again or publish a new built release base on this?
because I have written all  of of my codes and run it in a released version of carla9.5 before where everything is working in my project except the distance of obstacleSensor.(I was running through ./CarlaUe4.sh or compiled version) but when I made this changes and run Carla from source code(by opening in Unreal Editor), every thing in my project disturbed, I mean the quality of map, Fps that my project worked with was 20Fps but in this way(run from source code and by opening UnrealEditor) I could not define sample rate and it changes so the codes are not working in new sample-rate, coordination also changed here compared to the time I was running my program with ./CarlaUE4.sh!!! (actors or vehicles coordinate that I set before)
That is why I need a compiled version again or do you know why every thing that I set before is changing when I run it in this way(with UnrealEditor issue#22 or from source code)?

Thank you for your help,

SK398497:
#### Description

Fixes #1952 

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2182)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   * Exposed in the API: camera, exposure, depth of field, tone mapper and color attributes for the RGB sensor
   * Fixed materials and semantic segmentation issues regarding importing assets
   * Added TrafficManager to replace autopilot in managing the NPC vehicles
+  * Fixed ObstacleSensor to return HitDistance instead of HitRadius
 
 ## CARLA 0.9.6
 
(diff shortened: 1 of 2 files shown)


SK398497:
Hi @bernatx, as I do not own the repository from which the commit is being made, I have asked @irgendwie to make the change to his fix-obstacle branch.

marcgpuig:
We have some problems with Jenkins trying to access to this branch, don't know why. Can you try to update it? Also, adding a small fix description to the `CHANGELOG.md` will be useful, thanks :)

irgendwie:
@marcgpuig 
- Branch was rebased onto the current master
- A fix description was added to the CHANGELOG.md file

marcgpuig:
@irgendwie Thanks for your effort! We'r trying to fix jenkins right now, sorry for taking so long.

PR #2194, opened 2019-10-28
Forward header offset directly to the Array constructor

### LINKED ISSUE #2193: [0.9.6] Windows - assertion error
When using the C++ library, I get the already reported assertion error:

Assertion failed: (_data.size() - _offset) % sizeof(T) == 0u,
\carla\LibCarla\cmake..\source\carla/sensor/data/Array.h, line 137

I tried doing the same thing as https://github.com/carla-simulator/carla/issues/1778, but i still get the assertion error (although i build the client and server with the newest master - which means the "fix" should already be there).

I only try to do client.GetWorld() and the assertion error occurs.

After some rounds of debugging, I found out the following:

In LibCarla/source/carla/sensor/data/RawEpisodeState.h, we have:

```
explicit RawEpisodeState(RawData data)
     : Super(std::move(data)) {
    Super::SetOffset(Serializer::header_offset);
}
```

The call of the constructor of the base class will trigger the following code in LibCarla/source/carla/sensor/data/Array.h:

```
explicit Array(size_t offset, RawData data)
    : SensorData(data),
    _data(std::move(data)) {
    SetOffset(offset);
}

explicit Array(RawData data)
     : Array(0u, std::move(data)) {}

void SetOffset(size_t offset) {
    _offset = offset;
    DEBUG_ASSERT(_data.size() >= _offset);
    DEBUG_ASSERT((_data.size() - _offset) % sizeof(T) == 0u);
    DEBUG_ASSERT(begin() <= end());
}
```

So first, the second constructor is called which calls the first one with an offset of 0. The first constructor calls SetOffset and this will break because the offset is wrong.

So, when i changed LibCarla/source/carla/sensor/data/RawEpisodeState.h to look like:

```
explicit RawEpisodeState(RawData data)
      : Super(Serializer::header_offset, std::move(data)) {
}
```

it worked as expected on my machine.

Is this the correct solution for the problem?

Vaan5:
#### Description
In LibCarla/source/carla/sensor/data/RawEpisodeState.h, we have:

```
explicit RawEpisodeState(RawData data)
     : Super(std::move(data)) {
    Super::SetOffset(Serializer::header_offset);
}
```

The call of the constructor of the base class will trigger the following code in LibCarla/source/carla/sensor/data/Array.h:

```
explicit Array(size_t offset, RawData data)
    : SensorData(data),
    _data(std::move(data)) {
    SetOffset(offset);
}

explicit Array(RawData data)
     : Array(0u, std::move(data)) {}

void SetOffset(size_t offset) {
    _offset = offset;
    DEBUG_ASSERT(_data.size() >= _offset);
    DEBUG_ASSERT((_data.size() - _offset) % sizeof(T) == 0u);
    DEBUG_ASSERT(begin() <= end());
}
```

So first, the second constructor is called which calls the first one with an offset of 0. The first constructor calls SetOffset and this will break because the offset is wrong.

Because of that, LibCarla/source/carla/sensor/data/RawEpisodeState.h was changed like:

```
explicit RawEpisodeState(RawData data)
      : Super(Serializer::header_offset, std::move(data)) {
}
```

Fixes #2193 

  * **Platform(s):** Windows 10 64 bit 
  * **Python version(s):** 3.7

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2194)


nsubiron:
Same error happens in `LidarMeasurement` constructor, a bit more involved to fix there because the offset is variable.

germanros1987:
@bernatx is this going to be ever merged? Does it make sense? Are we dumping this?

PR #2199, opened 2019-10-30
Moved GNSS sensor from client to server

doterop:
#### Description

Moved the GNSS sensor from client to server side.

  * **Platform(s):** Linux
  * **Python version(s):** 3.5.2

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2199)


xmyqsh:
Good job! Can you elaborate on the pros and cons?

marcgpuig:
@xmyqsh we decided to port as much as we can to server so the client is as light as possible.
- **Pros:**
  - Fewer calculations on the client-side.
  - Exactly the same output for different clients listening to the same sensor.
- **Cons:**
  - More calculations on the server-side.
  - More networking traffic.

xmyqsh:
@marcgpuig Do you think it is proper to port `LaneInvasionSensor` to the server-side?
The following is my solution that I just figure out:
Create a `LaneInvasionSensor.*` under `Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor`,
and `#include <carla/road/...>` related functions to calculate the result just as the client-side solution does. All of the others are similar to the normal server-side sensor where the Server-side sensor Guide described.

nsubiron:
> Exactly the same output for different clients listening to the same sensor.

Unfortunately two clients cannot listen to the same sensor...

Issue #2081, opened 2019-09-06
Obstacle sensor distance remains constant


ndeshp2s:
Hello all,

I am trying to use the obstacle sensor in Carla 0.9.6. However, the sensor always shows a constant reading event after the obstacle is moved. How do i solve this?

Thank you.

dkarunakaran:
@ndeshp2s I am having the same problem. Did you solve it?

marcgpuig:
Did #2182 fixed this problem?

PR #2228, opened 2019-11-15
fix the obstacle sensor returned hit distance problem

### LINKED ISSUE #1816: distance of each vehicle's side to other obstacles
what is the best way to get the distance of each vehicle's side to other obstacles?
As I saw with getLocation() method I could just get the central location of the actor in the world and I could not get the coordinate of different points of the vehicle.
I have also tried other.obstacle sensor but with distance attribute which is supposed to get distance to other obstacle, I always get a constant value of 50 which is weird. does this sensor actually work?

 I would be thankful for your help

xmyqsh:
#### Description

Fixes #1816 #2081  

  * **Python version(s):** 2.7

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2228)


marcgpuig:
Hi @xmyqsh, thanks for the PR
Looks like I've just added a commit to your master while resolving the conflicts, that was really intrusive by my part, sorry :)

marcgpuig:
I'm closing this since #2182 has been merged into master

PR #2243, opened 2019-11-25
Feat: Adding the option to load a specific map while testing the spaw…

germanros1987:
…n_npc script

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2243)

### DIFF
--- a/PythonAPI/examples/spawn_npc.py
+++ b/PythonAPI/examples/spawn_npc.py
@@ -27,6 +27,7 @@ import logging
 import random
 import textwrap
 
+
 def list_options(client):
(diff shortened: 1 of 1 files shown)


jackbart94:
I've left a few comments, code is taken from the pre-existent config.py script. This way:
- Allows user to see available maps
- Allows to load new map
- Allows to restart existing map
- Avoids reset every time (as Nestor suggested)

PR #2292, opened 2019-12-11
Added IMU and Radar visualization to manual_control.py

marcgpuig:
#### Description

Added support for realtime IMU and Radar data visualization in `manual_control.py`.
- IMU is always enabled at the left panel
  + This means that now, the compass is rotated but it now correctly represents the OpenDrive's North. 
image
- Radar's visualization is disabled by default and it can be enabled pressing the `g` key. This has not been assigned to any keyboard number as you would expect, this is because I found the visualization easier to understand and validate using other cameras:
radar_pr
The red intensity means negative speed, and the blue one means positive speed in sensor coordinates. Depth, azimuth, and elevation are used to calculate the 3d position of each detection.

  * **Python version(s):** 3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2292)


nsubiron:
> The red intensity means negative speed, and the blue one means positive speed in sensor coordinates.

Ahh such a missed opportunity, [redshift](https://en.wikipedia.org/wiki/Redshift#Doppler_effect) works the other way around :smile:

PR #2319, opened 2019-12-28
Reference PR for CARLA without TM. 

felipecode:
Traffic manager is currently broke
for ports different than 2000. Even for the port 2000 I am experience Segfaults.
So this PR is for a version that has most of stuff from 0.9.7 but without TM integrated

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2319)


germanros1987:
Can we close this @felipecode ?

bernatx:
I think we can close this PR because there are no changes, and the new Traffic Manager is being merged with the port thing fixed, now it is multiclient.

Issue #2335, opened 2020-01-07
Low quality settings persist after switching back to Epic


BenAscent:
After running Carla in low quality using the following command, any subsequent run of the simulator retains the low quality level even if Epic quality is specified.

```
./CarlaUE4.sh -quality-level=Low
```

It looks to happen because the `~/.config/Epic/CarlaUE4/Saved/Config/LinuxNoEditor/GameUserSettings.ini` file is saved with the following content:

```
[ScalabilityGroups]
sg.ResolutionQuality=100.000000
sg.ViewDistanceQuality=3
sg.AntiAliasingQuality=3
sg.ShadowQuality=0
sg.PostProcessQuality=0
sg.TextureQuality=0
sg.EffectsQuality=0
sg.FoliageQuality=0
```

But when Carla is started again in Epic quality, those settings are not replaced by their initial values (quality of 3 for all sg), making the low quality mode persisting after the first use, unless the ini file mentioned above is manually cleared.

My suspicion is that the settings from this file are applied after the code that handles `quality-level`, but I haven't tried to verify that.

This occurs with version 0.9.7 on Linux (Ubuntu 18.04).

This issue is particularly annoying due to #2336

germanros1987:
We are working on fixing this very annoying issue.

BenAscent:
@germanros1987 Thank you!

Note that it seems that calling
```
Scalability::ToggleTemporaryQualityLevels(true);
```
Before applying quality level settings seem to solve the problem. I tried to put it in `ApplyQualityLevelPostRestart()` just before the switch `statement`.

This function seems to prevent saving the settings change into the user file. You'll need to include `Scalability.h`

PR #2352, opened 2020-01-10
Fix Low / Epic quality levels transition

### LINKED ISSUE #2336: Low-quality mode is extremely bright
When running in low quality mode, there is an exposure problem which makes the game very bright. See screenshot below.

**Epic quality**
image

**Low quality**
image

This happens in 0.9.7 and was not happening in 0.9.6.

It seems to be due to the scalability level for post-process that is set to 0 in low quality:
```
sg.PostProcessQuality=0
```
More specifically, it's this specific value that causes the problem:
```
r.EyeAdaptationQuality=0
```
(as defined in `[PostProcessQuality@0]` of BaseScalability.ini).

Switching `r.EyeAdaptationQuality` to 1 or 2 solves the problem.

We also found out that using the apparently new `BP_WeatherSafe` as weather class instead of `BP_Weather` prevents the problem to occur.

This problem occurs on both Windows and Linux (Ubuntu 18.04). GPUs tested include nVidia 1070 and 2070.

bernatx:
#### Description

This PR will fix the transition from Epic / Low quality settings. 

Two problems were found in version Carla 0.9.7:

First, Unreal was writing the last quality settings to a file in /Saved/ folder and was reading and applying those settings without taking care of the desired quality settings that the user could specify.
We have fixed this thanks to the help of BenAscent who pointed the way to disable the writing of those settings.

Second, when enabling low-quality settings, the autoexposure was disabled but a value of Exposure Compensation of 4.5 was applied what made the image brighter than usual. We fixed that forcing autoexposure to be enabled also at low-quality level.

Fixes #2336, #2335

  * **Platform(s):** Ubuntu and Windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

If you had the problem of bright image in low-quality settings, probably you will need to remove the /Saved/ folder because there Unreal has written the last quality settings. Just clean that folder and run again with any Epic or Low quality settings.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2352)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Settings/CarlaSettingsDelegate.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Settings/CarlaSettingsDelegate.cpp
@@ -14,6 +14,8 @@
 #include "InstancedFoliageActor.h"
 #include "Kismet/GameplayStatics.h"
 #include "Landscape.h"
+#include "Scalability.h"
+
 
 static constexpr float CARLA_SETTINGS_MAX_SCALE_SIZE = 50.0f;
 
@@ -71,10 +73,14 @@ void UCarlaSettingsDelegate::ApplyQualityLevelPostRestart()
   UWorld *InWorld = CarlaSettings->GetWorld();
 
   const EQualityLevel QualityLevel = CarlaSettings->GetQualityLevel();
+
   if (AppliedLowPostResetQualityLevel == QualityLevel)
   {
     return;
   }
+
+  // enable temporal changes of quality (prevent saving last quality settings to file)
+  Scalability::ToggleTemporaryQualityLevels(true);
 
   switch (QualityLevel)
   {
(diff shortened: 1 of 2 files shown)


jeffreykxiao:
This change seems to work fine for me (Ubuntu and Windows). One thing that I think should be mentioned is that the location of the saved config on Linux was in my user directory under `.config` or something similar to this.

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Settings/CarlaSettingsDelegate.cpp
```diff
@@ -159,13 +165,17 @@ void UCarlaSettingsDelegate::CheckCarlaSettings(UWorld *world)
 
 void UCarlaSettingsDelegate::LaunchLowQualityCommands(UWorld *world) const
 {
+  if (!world)
+  {
+    return;
+  }
+
```
jeffreykxiao: Thanks for adding this! Not sure under what circumstances it would get popped, but I definitely appreciate that the function implementations are now consistent with each other.

bernatx: Yes, I think it is more consistent that way. Thanks.

marcgpuig:
@jeffreykxiao looks like you are blocking the PR, can you accept it?

bernatx:
I think jeffreykxiao needs to resolve a discussion yet. Can you check further? The PR is still blocked.

jeffreykxiao:
> I think jeffreykxiao needs to resolve a discussion yet. Can you check further? The PR is still blocked.

Should be good now - I didn't realize that I needed to review in Reviewable and not github. 🙇

exoticDFT:
> @jeffreykxiao Indeed!
> I had to remove `rm -rf ~/.config/Epic/CarlaUE4/Saved` to make it work again (Ubuntu 18.04).
> I think it will be worth mentioning this somewhere in our docs.
> 
> > Reviewed 2 of 2 files at r1.
> > _[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2352#-:-LyZ9aVG9CxSDH5N4-IT:b-li6b1k)_ status: all files reviewed, 1 unresolved discussion (waiting on @bernatx)

@marcgpuig As a word of caution, this suggestion may be overkill. Specifically for those that stored some previous recordings made in Carla. The only file that needs to be removed in this folder is `~/.config/Epic/CarlaUE4/Saved/Config/LinuxNoEditor/GameUserSettings.ini`. This is the file that stores previously saved configuration for the settings mentioned in these issues.

marcgpuig:
@exoticDFT Indeed, the docs have been correctly updated (not uploaded yet) with this specific information and I've updated my comment to avoid problems to other users. Thanks for pointing it out!

exoticDFT:
@marcgpuig Great to hear it's already in the next documentation update! Glad I could help.

PR #2371, opened 2020-01-20
New weather implementation

Madecu:
#### Description

- Added a new script in PythonAPI/util to be able to change the sun position and the weather parameters individually.
- Activated Mesh Distance Fields.
- Changed `cloudyness` for `cloudiness`.
- Added three new parameters for the weather (Fog density, Fog distance, and Wetness).
- New function "SetWeather" in weather.cpp to be able to update the weather parameters without notifying it to the blueprint's event.

  * **Platform(s):** Windows and Linux
  * **Python version(s):** Python 3.7 

#### Possible Drawbacks

No drawbacks found yet.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2371)


a-jahani:
Mesh Distance Fields was disabled to remove ghosting effect! #1838

Madecu:
> Mesh Distance Fields was disabled to remove ghosting effect! #1838

We decided to enable Mesh Distance Fields again to improve the lighting in our scenes. We still have some ghosting but we managed to minimize it as much as we could. On the other hand, we gain a much better looking with Distance Fields Ambient Occlusion and Distance Fields Shadows combined with Cascade Shadows.
If the ghosting is still an issue, the user can always disable it. We will explain it how in the 0.9.8 release documentation.

PR #2407, opened 2020-01-28
Taking account of the very last waypoint buffer

jahaniam:
This PR is a small hotfix that lets the vehicle go exactly to the desired location instead of stopping before reaching the destination.

Previously the local planner stops sending control commands when __waypoints_queue_ is empty. However, there is also a __waypoint_buffer_ which is not empty. The very last round, when the vehicle reaches the desired destination, not only the __waypoints_queue_  is empty but also the  __waypoint_buffer_  must be empty as well. The way it is now, there is always a bias of _self._buffer_size_ number of waypoints to the desired destination.

  * **Python version(s):** 3.6

#### Possible Drawbacks

I could not find any usage of function 'def done(self):'. This function might be a deprecated function. Nevertheless, I have also updated the function to return the correct value.
Previously the second part was always True: if `len(self._waypoints_queue) == 0` then `all([distance_vehicle(wp, vehicle_transform) < self._min_distance for wp in self._waypoints_queue])` is always True. So there is no point in having the second part.

        return len(self._waypoints_queue) == 0 and all([distance_vehicle(wp, vehicle_transform) < self._min_distance for wp in self._waypoints_queue]

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2407)


germanros1987:
Thanks @a-jahani for your contribution! @pravinblaze this looks like a straightforward PR, could you please take care of the review?

PythonAPI/carla/agents/navigation/local_planner.py
```diff
@@ -199,7 +199,7 @@ def run_step(self, debug=True):
         if not self._global_plan and len(self._waypoints_queue) < int(self._waypoints_queue.maxlen * 0.5):
             self._compute_next_waypoints(k=100)
 
-        if len(self._waypoints_queue) == 0:
+        if len(self._waypoints_queue) == 0 and len(self._waypoint_buffer) == 0:
```
pravinblaze: Agreed :) This is definitely the right way to go. Thanks for pointing this out, I had missed it.

PythonAPI/carla/agents/navigation/local_planner.py
```diff
@@ -243,7 +243,7 @@ def run_step(self, debug=True):
 
     def done(self):
         vehicle_transform = self._vehicle.get_transform()
-        return len(self._waypoints_queue) == 0 and all([distance_vehicle(wp, vehicle_transform) < self._min_distance for wp in self._waypoints_queue])
+        return len(self._waypoints_queue) == 0 and len(self._waypoint_buffer) == 0
```
pravinblaze: Yup, easier this way.

pravinblaze:
The LocalPlanner.done() is used in BasicAgent to check if the agent has reached it's intended destination.

jackbart94:
Tested, working as expected. Looks neat!

jahaniam:
___
*[PythonAPI/carla/agents/navigation/local_planner.py, line 246 at r2](https://reviewable.io/reviews/carla-simulator/carla/2407#-Lze-nTefRNrIT8g-56Z-r2-246:-M1R4OL_7ecty3G1biCI:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/50d061ede2b97a4a511e2f69e3413e2d32d74b5f/PythonAPI/carla/agents/navigation/local_planner.py#L246)):*
<details><summary><i>Previously, pravinblaze (Praveen Kumar) wrote…</i></summary><blockquote>

Yup, easier this way.
</blockquote></details>

Done.

jahaniam:
___
*[PythonAPI/carla/agents/navigation/local_planner.py, line 202 at r2](https://reviewable.io/reviews/carla-simulator/carla/2407#-Lze-nTefRNrIT8g-56Z-r2-202:-M1R42pED1zsHwGHRGxl:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/50d061ede2b97a4a511e2f69e3413e2d32d74b5f/PythonAPI/carla/agents/navigation/local_planner.py#L202)):*
<details><summary><i>Previously, jahaniam (Ali Jahani) wrote…</i></summary><blockquote>

Sounds good. Thank you for your code in the first place. It looks clean and easy to read.
</blockquote></details>

Done.

germanros1987:
@jahaniam if you solve the conflicts @bernatx will do the merge.

jahaniam:
Yeah. I think everything should be good now. Feel free to merge

PR #2414, opened 2020-01-29
Exposed car lights to client

marcgpuig:
#### Description

Now the vehicle lights can be turned on and off from the client-side. `manual_control.py` has been updated to allow this too, so feel free to experiment with it. Open the help panel (`H` key) for more information.
You will need the latest art repo to work with cars that have lights implemented. Right now they are: Tesla **Model3**, **Cybertruck** (yup :wink:) and the Volkswagen **T3**

To control them from the client we exposed the `VehicleLightState` that is an enum that works as a flag container. The values are:
- Position
- LowBeam
- HighBeam
- Brake
- RightBlinker
- LeftBlinker
- Reverse
- Fog
- Interior
- Special1
- Special2

_The special ones are reserved to car specific lights (like ambulance sirens)_

An example opening the position light and low beam light from some vehicle:

```py
# Decide which ones do we want to open, adding these flags
lights_to_open = carla.VehicleLightState(
    carla.VehicleLightState.Position | carla.VehicleLightState.LowBeam)
# Set the state of the car lights to match the flag we created
vehicle.set_light_state(lights_to_open)
```

  * **Python version(s):** 3.6

#### Possible Drawbacks

You can experiment with performance issues when enabling big amount of lights. This is more an Unreal/Art issue but should be taken into consideration while using the API.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2414)


hzyjerry:
The cybertruck is amazing!

PR #2390, opened 2020-01-23
Fix import pipeline crashes and missing semantic tagging

Jazzinghen:
#### Description

 * Changed the import python script to avoid running commands that would crash if no pedestrian navmesh is available.
 * Changed the Move assets commandlet to properly recognize the static meshes and put them in the right folder
 * Changed the Prepare assets for cooking commandlet to not overwrite the correct map with one without the semantic lane data

#### Fixes  

The import pipeline: 

 * Made assumptions on the optional data structures
 * Didn't properly place the static meshes for semantic tagging
 * Generated semantic information, saved the map, deleted the info from the map and then saved the map again

Fixes issues https://github.com/carla-simulator/carla/issues/2332 and https://github.com/carla-simulator/carla/issues/2340

#### Where has this been tested?

  * **Platform(s):** Windows 10 64 bits, Ubuntu Linux 18.04 64 bits
  * **Python version(s):** Python 3.7
  * **Unreal Engine version(s):** 4.22

#### Possible Drawbacks

 * A potential side effect is that the import process now doesn't treat pedestrian nav meshes as necessary, meaning that if someone is expecting it to crash when they are not available the script won't stop.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2390)

PR #2442, opened 2020-02-05
Demonstration of left-handed hack for LibCarla

jeffreykxiao:
#### Description

Hacked in some left-handed traffic assumptions into LibCarla. This changes the assumptions from right-handed traffic (default) to left-handed. As this is not a good change, it is in a draft PR for reference purposes and should never be merged. 

This doesn't resolve any issues but users who want to ONLY work on left-handed traffic can use this fix. 

#### Procedure

First, use RoadRunner and make a map with left-handed traffic (left-hand lanes should have "forward" travel direction, "backward" for right handed lanes), export with the left-hand setting.

1. Get this change
2. `make clean`
3. `make LibCarla` (builds this change. copying into CarlaUE4 and PythonAPI dependencies folder)
4. `make pythonAPI` (builds PythonAPI with this change)
5. `make launch` (starts CarlaUE4 with this change)
6. Import the xodr/fbx you generated (Don't use `make import` because it's broken on master, I rely on [this change that fixes that](https://github.com/carla-simulator/carla/pull/2390))
7. run example

#### Proposal

I created this to show that it's fairly simple to handle left-handed traffic, it's just blocked by some assumptions carla makes. I have some ideas for real support:
- Determine first what kind of left-handed support we want to give. I think it's a requirement to support RHT and LHT in the same build, but is it necessary to support RHT and LHT in the same MAP?
- Ask RoadRunner to update their OpenDrive spec. They are using Opendrive 1.4, but if RoadRunner updated to OpenDrive 1.5 spec from 1.4 and use the updated Road Header Record (http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.5M.pdf, section 5.3.1, rule), it could get rid of the current `<UserData><VectorRoad travelDir=forward/backward>` setup. Then Carla would also be able to support the OpenDrive 1.5 format (instead of needing to specifically support the custom tag in the OpenDrive file produced by RoadRunner). 
- Carla needs to parse either the RoadRunner tag or the OpenDrive 1.5 RHR to determine the handedness of roads. If we don't expect to find LHT and RHT together in a map, as soon as we see a lane of positive lane ID + backward/forward, we can assume the traffic direction for the whole file. If not, then it gets a bit more complicated but still doable.
- When parsing the lanes, since we already check the contents of <left> <center> and <right> headers in OpenDrive, why don't we give Carla Roads more descriptive arrays of left/center/right lanes? 
- If we don't want to do that, can we at least make a helper function for checking the ID? I didn't do it this time, but it is weird to keep doing comparisons against the lane ID. Would be much better to be on a higher level of abstraction and have a clear function that does it instead of a lane ID comparison and a comment every time saying what it's all about.

  * **Platform(s):** ... Windows 10 x64, Ubuntu 18.04
  * **Python version(s):** ... 3.6.9
  * **Unreal Engine version(s):** ... 4.22.3

#### Possible Drawbacks

It literally breaks all right-handed traffic maps on import and driving on left-handed maps, so it has a huge drawback.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2442)

### DIFF
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -43,7 +43,7 @@ namespace road {
   }
 
   static double GetDistanceAtStartOfLane(const Lane &lane) {
-    if (lane.GetId() <= 0) {
+    if (lane.GetId() >= 0) {
       return lane.GetDistance() + 10.0 * EPSILON;
     } else {
       return lane.GetDistance() + lane.GetLength() - 10.0 * EPSILON;
@@ -51,7 +51,7 @@ namespace road {
   }
 
   static double GetDistanceAtEndOfLane(const Lane &lane) {
-    if (lane.GetId() > 0) {
+    if (lane.GetId() < 0) {
       return lane.GetDistance() + 10.0 * EPSILON;
     } else {
       return lane.GetDistance() + lane.GetLength() - 10.0 * EPSILON;
@@ -339,7 +339,8 @@ namespace road {
 
     dp.ApplyLateralOffset(lane_width);
 
-    if (waypoint.lane_id > 0) {
+    // Backwards-facing waypoints (right lanes) need to be turned around
+    if (waypoint.lane_id < 0) {
       rot.yaw += 180.0f + geom::Math::ToDegrees(lane_tangent);
       rot.pitch = 360.0f - rot.pitch;
     } else {
@@ -383,7 +384,7 @@ namespace road {
 
   std::pair<const RoadInfoMarkRecord *, const RoadInfoMarkRecord *>
   Map::GetMarkRecord(const Waypoint waypoint) const {
-    // if lane Id is 0, just return a pair of nulls
+    // if lane Id is 0, just return a pair of nulls (Lane ID 0 = center lane marker)
     if (waypoint.lane_id == 0)
       return std::make_pair(nullptr, nullptr);
 
@@ -511,7 +512,10 @@ namespace road {
       const double distance) const {
     RELEASE_ASSERT(distance > 0.0);
     const auto &lane = GetLane(waypoint);
-    const bool forward = (waypoint.lane_id <= 0);
+
+    // Assume left-handed traffic (left and center lane travel forward)
+    const bool forward = (waypoint.lane_id >= 0);
+
     const double signed_distance = forward ? distance : -distance;
     const double relative_s = waypoint.s - lane.GetDistance() + EPSILON;
     const double remaining_lane_length = forward ? lane.GetLength() - relative_s : relative_s;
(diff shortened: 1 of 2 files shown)


gjossep:
Not sure If im doing something wrong, but after getting the files. Doing a make clean, and then make LibCarla I get the following error:
`LibCarla/source/carla/road/Map.cpp:582:30: error: out-of-line definition of 'GenerateWaypointsOnRoadEntries' does not match any declaration in 'carla::road::Map'
  std::vector<Waypoint> Map::GenerateWaypointsOnRoadEntries() const {
`

jeffreykxiao:
@gjossep Hopefully I'll have some time next week to check this out. Maybe Map.cpp changed on master or something since I am seeing that I can't update this branch from github due to a merge conflict.

I cloned this branch directly and it works fine for me. Maybe you can do that, build libcarla and pythonapi, and copy the files into your own carla. If you're cherry-picking my change onto master I'm sure there could be some issues, but the merge conflict shouldn't be too difficult to solve either.

germanros1987:
@jeffreykxiao kudos to you for this hack. Would you be interested in coordinating with the team to propose a proper solution that can stand the pass of time?

joealan11:
Hello, I am very interested in making this work. I have been trying to adapt this to the new version of CARLA 0.9.9 but it seems that they made a lot of changes in the way they generate the waypoints. Have you had a look at the new version? any advice? 

My other option is to make the route planner work again in the new version, but I do not know if this is viable.

kino3:
Hi, I am also very interested in this work. Is there any update? @jeffreykxiao @gjossep @germanros1987 @joealan11 @marcgpuig 
I searched the issues, but cannot found any feature requests of left-hand driving. [Many countries (like UK, Japan)](https://en.wikipedia.org/wiki/Left-_and_right-hand_traffic) adopt left-hand traffic, so I hope Carla will officially support left-hand traffic.

jxzzhang:
> #### Description
> Hacked in some left-handed traffic assumptions into LibCarla. This changes the assumptions from right-handed traffic (default) to left-handed. As this is not a good change, it is in a draft PR for reference purposes and should never be merged.
> 
> This doesn't resolve any issues but users who want to ONLY work on left-handed traffic can use this fix.
> 
> #### Procedure
> First, use RoadRunner and make a map with left-handed traffic (left-hand lanes should have "forward" travel direction, "backward" for right handed lanes), export with the left-hand setting.
> 
> 1. Get this change
> 2. `make clean`
> 3. `make LibCarla` (builds this change. copying into CarlaUE4 and PythonAPI dependencies folder)
> 4. `make pythonAPI` (builds PythonAPI with this change)
> 5. `make launch` (starts CarlaUE4 with this change)
> 6. Import the xodr/fbx you generated (Don't use `make import` because it's broken on master, I rely on [this change that fixes that](https://github.com/carla-simulator/carla/pull/2390))
> 7. run example
> 
> #### Proposal
> I created this to show that it's fairly simple to handle left-handed traffic, it's just blocked by some assumptions carla makes. I have some ideas for real support:
> 
> * Determine first what kind of left-handed support we want to give. I think it's a requirement to support RHT and LHT in the same build, but is it necessary to support RHT and LHT in the same MAP?
> * Ask RoadRunner to update their OpenDrive spec. They are using Opendrive 1.4, but if RoadRunner updated to OpenDrive 1.5 spec from 1.4 and use the updated Road Header Record (http://opendrive.org/docs/OpenDRIVEFormatSpecRev1.5M.pdf, section 5.3.1, rule), it could get rid of the current `<UserData><VectorRoad travelDir=forward/backward>` setup. Then Carla would also be able to support the OpenDrive 1.5 format (instead of needing to specifically support the custom tag in the OpenDrive file produced by RoadRunner).
> * Carla needs to parse either the RoadRunner tag or the OpenDrive 1.5 RHR to determine the handedness of roads. If we don't expect to find LHT and RHT together in a map, as soon as we see a lane of positive lane ID + backward/forward, we can assume the traffic direction for the whole file. If not, then it gets a bit more complicated but still doable.
> * When parsing the lanes, since we already check the contents of   and  headers in OpenDrive, why don't we give Carla Roads more descriptive arrays of left/center/right lanes?
> * If we don't want to do that, can we at least make a helper function for checking the ID? I didn't do it this time, but it is weird to keep doing comparisons against the lane ID. Would be much better to be on a higher level of abstraction and have a clear function that does it instead of a lane ID comparison and a comment every time saying what it's all about.
> 
> #### Where has this been tested?
> * **Platform(s):** ... Windows 10 x64, Ubuntu 18.04
> * **Python version(s):** ... 3.6.9
> * **Unreal Engine version(s):** ... 4.22.3
> 
> #### Possible Drawbacks
> It literally breaks all right-handed traffic maps on import and driving on left-handed maps, so it has a huge drawback.
> 
> This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2442)

Hi I am also looking into a left handed modification. have you be able to make it work on later version of Carla? thanks

PR #2468, opened 2020-02-12
Doterop/traffic manager

doterop:
#### Description

Implementation of the multiclient architecture for Traffic Manager, including the possibility of specifying on which port to create the TM, having multiple TM inside the same simulation and allowing clients to connect to a specific TM.

Also, it includes commits for fixing a map changing error with TM, some behavior modifications regarding unsignalized junctions, a benchmark for TM performance and small improvements by reducing the number of GetLocation() calls.

  

  * **Python version(s):** 3.6 and 2.7

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2468)


LibCarla/source/carla/trafficmanager/CollisionStage.cpp
```diff
@@ -72,51 +76,45 @@ namespace CollisionStageConstants {
...
+                  NegotiateCollision(ego_actor, other_actor, ego_location, other_location, closest_point, junction_look_ahead)) {
 
-                if (safe_point_junction != nullptr){
-                  if(parameters.GetCollisionDetection(ego_actor, actor) &&
-                    !IsLocationAfterJunctionSafe(ego_actor, actor, safe_point_junction)){
+                  if ((other_actor_type[0] == 'v' && parameters.GetPercentageIgnoreVehicles(ego_actor) <= (rand() % 101)) ||
+                      (other_actor_type[0] == 'w' && parameters.GetPercentageIgnoreWalkers(ego_actor) <= (rand() % 101))) {
 
                     collision_hazard = true;
                     break;
```
pravinblaze: This break statement seems unnecessary. The looping condition at line 80 already tracks the state of variable collision_hazard and exits if it's value becomes true.

PR #2478, opened 2020-02-14
Fixed RoadOptions.STRAIGHT at rural areas

glopezdiest:
#### Description

The threshold used to differentiate between RoadOptions.STRAIGHT and RoadOptions.LEFT, or RoadOptions.RIGHT, has been increased:
  1. at local planner, from 1º to 20º
  2. at global route planner, from 5º to 20º

**EDIT**: Increased to 35

This fixes some intersections outside city areas, where the junctions aren't exactly perpendicular and where some lanes where detected as LEFT / RIGHT. 

  * **Platform(s):** Ubuntu 4.22
  * **Python version(s):** 3.6

#### Possible Drawbacks

None?

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2478)


pravinblaze:
___
*[PythonAPI/carla/agents/navigation/global_route_planner.py, line 264 at r2](https://reviewable.io/reviews/carla-simulator/carla/2478#-M04GE5WBkdGj5yCYekb:-M0MCCz67_rmWgs0TuQn:b-j7f4m9) ([raw file](https://github.com/carla-simulator/carla/blob/def9d3d00296a31ab47817ebfdcc5da8c26a00b8/PythonAPI/carla/agents/navigation/global_route_planner.py#L264)):*
<details><summary><i>Previously, germanros1987 wrote…</i></summary><blockquote>

@pravinblaze do you anticipate that this change could break something?
</blockquote></details>

The default value of threshold is used for determining what angular deviation is acceptable for an intersection exit to be considered straight. Changing it from 5 to 20 degrees won't cause any serious malfunction, but there could be places where a slight turn could be interpreted as a straight exit if the angular deviation is below the threshold.

glopezdiest:
I actually changed it to avoid the opposite problem, as, at rural areas, it was detecting as left turns roads that were going straight. 20 is the minimum number that changed that behavior.

glopezdiest:
Increased the threshold to 33 to include a turn in Town08

PR #2550, opened 2020-03-03
Make PID controller insensitive to time-discretization

### LINKED ISSUE #2544: _pid_control accounts for time discretization twice (not once)
It appears `_pid_control` incorrectly accounts for `self._dt` *twice* (not once), for both the "D" and "I" terms, and is therefore sensitive to the value of the time discretization (bad!). For example, currently the "D" term is:
`self._K_D * _de / self._dt == `
`self._K_D * (self._e_buffer[-1] - self._e_buffer[-2]) / self._dt / self._dt`

But it should be:
`self._K_D * (self._e_buffer[-1] - self._e_buffer[-2]) / self._dt`
by either removing `self._dt` from lines 117-118, or line 123.

https://github.com/carla-simulator/carla/blob/master/PythonAPI/carla/agents/navigation/controller.py#L117-L123

rowanmcallister:
#### Description

Updated PID controller, to make it insensitive to chosen value of time discretization.

Fixes #2544

  * **Python version(s):** 3.6.10

#### Possible Drawbacks

If people have already tuned PID controllers on non-default time discretizations (non 0.05 values), they will need to re-tune their PID `K_I` and `K_D` terms by multiplying and diving by a factor of `self._dt / 0.05` respectively.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2550)


germanros1987:
@pravinblaze could you check that this works as expected? Please, test it for different fixed steps (0.2 s, 0.1 s, etc.).

PR #2623, opened 2020-03-19
Migration ue422 to ue424

Axel1092:
#### Description

Fixes code to upgrade Unreal Engine to 4.24.
Solves many crashes related to Vulkan.

  * **Platform(s):** Ubuntu 18.04, Windows
  * **Python version(s):** 2, 3
  * **Unreal Engine version(s):** UE4.24

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2623)


nsubiron:
UE4.24 upgrades the toolchain to clang-8.0.1 ([release notes](https://docs.unrealengine.com/en-US/Support/Builds/ReleaseNotes/4_24/index.html)), the compiler and libc++/libabi++ should also be upgraded to avoid possible ABI issues.

nsubiron:
*[a discussion](https://reviewable.io/reviews/carla-simulator/carla/2623#-M2rJTVa3KDZI4HmuU3F:-M2s2pdp1wHs2-p5gzZX:b65fuas) (no related file):*

Nestor, currently it compiles and link fine without errors. Do you think we can proceed with the upgrade and later on make the upgrade to Clang 8.0.1?
</blockquote></details>

No, I think you should definitively use same compiler and runtime, that's precisely the problem with ABI issues, it compiles and links without errors but it may crash at runtime at unrelated places. Plus you're compiling with one version of the standard library but then in UE4 you'll link against a different one, there are no guarantees that something may have changed.

---

PR #2647, opened 2020-03-24
Feature/integrate rss 3.0.0

berndgassmann:
#### Description

Move carla::client::Timestamp ostream operator to Timestamp.h to enable
reuse in log messages

Use of generated python interfaces of ad::physics, ad::map and ad::rss
types

RssSensor (and therefore also the RssRestrictor) now work globally on
the whole road network.

RssCheck: object conversion done in parallel using tbb

Fixes #  

  * **Platform(s):** Ubuntu 16.04 / 18.04
  * **Python version(s):** 2.7, 3.5
  * **Unreal Engine version(s):** 4.22, 4.24

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2647)


bernatx:
Hi Bernd,
I'm trying to build with RSS.
I have upgraded to compile version 8.0, so I have changed the 'boost-1.72.0-**c7**-install' to 'boost-1.72.0-**c8**-install'.
Also, there is an error here:
```
/home/jenkins/workspace/carla_bernat_rss/Util/BuildTools/Ad-rss.sh: line 26: colcon: command not found
Util/BuildTools/Linux.mk:129: recipe for target 'ad-rss' failed
```
What is **colcon**?

bernatx:
Thanks Bernd,
Another thing is, we don't use submodules in the repository, we instead clone and build from any repository we need from the setup.sh script. Do you think we can do the same with the RSS submodule. Just to follow the same philosophy. I think something like using **make setup.rss** and then it will download the RSS repository.
What do you think?

marcgpuig:
@berndgassmann Thanks for the effort and the changes, this is really great :)

PR #2648, opened 2020-03-24
DVS camera sensor implementation

jhidalgocarrio:
#### Description

This PR adds a Dynamic Vision Sensor (DVS) camera together with the new sensor data of events and the serialization and deserialization mechanism.

This is how a DVS sensor works: https://youtu.be/LauQ6LWTkxM

  * **Platform(s)**: Ubuntu 18.04 64bits
  * **Python version(s)**: Python 2.7 and Python3.6
  * **Unreal Engine version(s)**: 4.22

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2648)


jhidalgocarrio:
image
 
Standard RGB camera (left) and Events visualization on an "event frame" (right) blue is a positive event (brightness increase) red is a negative event (brightness decrease)

shiba24:
Thank you for this feature. I love it.
Should we support `save_to_disk` function?

shiba24:
^^ I just created a PR to support `save_to_disk`. Hope @jhidalgocarrio and @marcgpuig will take a look sometime. Open to any suggestion and/or discussion. Thank you!

marcgpuig:
Hi @jhidalgocarrio,

Sorry for taking so long to correctly verify your code. First of all, congratulations for this PR, it has been really well integrated with our code architecture. I've made a few changes, most of them are just coding style. That being said, I have found some things that should be addressed.
In order to keep the PythonAPI as clean as possible, I think we should get rid of `ColorVector`, `IntVector`, `ShortVector`, and `IntMatrix`. I've already done it for the `ColorVector` as an example.
We also think that the DVS logic in `SensorData.cpp` could be moved to `DVSEventArray.h`. I've also made an example, moving `ToImage()` there. It would be nice if you could do the same moving the other `to_array_*` functions there. On the other hand, I don't see the purpose of having `to_array` since the DVSEventArray is an array itself, is there something I'm missing?
I've also added an example in `manual_control.py` showing how to convert the whole `raw_data` into a NumPy array in order to generate the image there (as we do with the Lidar sensor):
```py
# Raw data from a carla.DVSEventArray to NumPy array
dvs_event_array = np.frombuffer(dvs_sensor_data.raw_data, dtype=np.dtype([
     ('x', np.uint16), ('y', np.uint16), ('t', np.int64), ('pol', np.bool)]))

# Then the individual lists can be used like this
dvs_event_array[:]['y']
```
This can be added later on in the documentation as well.

Thanks for your time.

sergi-e:
@jhidalgocarrio It would be great to merge this and have some documentation about it. Mind if I ask you some questions about what do users need to know? :)

jhidalgocarrio:
Thanks for your comments. I am currently adapting the code.
@sergi-e where should I include the basic user documentation?

sergi-e:
@jhidalgocarrio 
Basically, sensors appear in three pages in the docs: 
__1. Sensor reference (ref_sensors.md):__ Add a description of how the sensor works, a table with the output attributes of the SensorData and another one with the attributes of the blueprint that the user can customize. If needed, create other text or tables (for example, the radar has two output tables, and the IMU none). 
__2. PythonAPI.__ Sensors appear listed in carla.Sensor and carla.SensorData. Besides that, add whatever information accessible from the API (normally, a child of carla.SensorData). 
__3. First steps, 4th- Sensors and data.__ There is a brief mention of the different sensors available.
Besides that, try to avoid referring to the user directly ("you can..."), and in case you have any doubts on style, you can use the Microsoft style guide as reference: https://docs.microsoft.com/en-us/style-guide/welcome/

And anything you need, here I am. In case you prefer it, I can take care of the documentation if you provide me with the necessary information!

marcgpuig:
@shiba24 I couldn't find your PR regarding the `save_to_disk` function, are u still interested on adding it?

shiba24:
@marcgpuig thanks, the reason why I wanted to implement `save_to_disk` is, when I used the python API `raw_data()` and/or `to_array()` to save the DVS data in client side, I have faced with an issue that sync between client and server was broken. I was not able to find the root cause then but I guess that is because event data's potentially big size on memory.

So if it does not happen with the latest master anymore, we may not need that functionality?
But yes, Im interested in it and happy to commit, if it is still necessary - do you think you still need that function (or at least nice to have)?

marcgpuig:
@shiba24 I see.
Well, I think the main reason for a function like `save_to_disk` is to generate datasets, so I think that if you have already done it, It will be nice to add it in master :)

shiba24:
OK cool, let me try to create a PR for that with the latest master rebased then. Hopefully I'll open it this weekend.
Lmk if you have some schedule (timeline) in mind that you want to merge save_to_disk into master.

marcgpuig:
@shiba24 please, take your time. It will be good to merge it before 0.9.10 (maybe 2-3 weeks from now?) so the DVS will be on that release fully featured.
Thanks for your time.

Cheers

PR #2662, opened 2020-03-27
Fix build on old compilers

nsubiron:
Fixes some minor issues I've found compiling the client side with an older version of clang.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2662)


bernatx:
Thanks for the fixes Nestor !

PR #2690, opened 2020-04-03
sergi-e/asset-import

sergi-e:
#### Description

- New iteration on map creation and importing assets. These are now completely separated. 
- Some modifications to Standalone packages to include only instructions related to this subject. 

  * **Platform(s):** Ubuntu 18.04.4
  * **Python version(s):** 3.6.9 and 2.7.17

#### Possible Drawbacks

- Add vehicles manually is yet to be edited.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2690)


bernatx:
I think we can wait to add the new feature that creates a .json file if no other .json file is found in the folder to import (carla_root/import/).
I'm testing the script right now. If it works, we can add some explanation about not to define any .json file to import only maps. 

- The script automatically search for all .fbx and .xodr files. If it finds a pair of .fbx and .xodr in the same folder with the same name, then it will add an entry in the .json file with that map. 

- The .json file created is called 'automatic.json' and placed in the /import/ folder of Carla.

PR #2722, opened 2020-04-12
Fix variable definition

shiba24:
Before this commit, `synchronous_master` is defined inside
try-block so if any error happens before the definition and args.sync is true,
"ReferenceError: `synchronous_master` is not defined" is raised
in the finally-block.

You can reproduce this error when you run `python spawn_npc.py --sync` without good connection to the server (like `RuntimeError: rpc::timeout: Timeout of 10000ms while connecting to 127.0.0.1:2000`)

#### Description

Described above (and in the commit message too)

  * **Platform(s):** ...Ubuntu 16
  * **Python version(s):** ... python 2.7.13
  * **Unreal Engine version(s):** ... 4.24

I am pretty sure this is OS independent.

#### Possible Drawbacks

None.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2722)

### DIFF
--- a/PythonAPI/examples/spawn_npc.py
+++ b/PythonAPI/examples/spawn_npc.py
@@ -91,17 +91,15 @@ def main():
     all_id = []
     client = carla.Client(args.host, args.port)
     client.set_timeout(10.0)
+    synchronous_master = False
 
     try:
-
         world = client.get_world()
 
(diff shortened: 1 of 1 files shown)


bernatx:
Thanks for the fix shiba24, we are going to merge this in master.
Kind regards

PR #2787, opened 2020-04-26
Building Carla on Ubuntu 20.04 LTS

pravinblaze:
#### Description

Most of the instructions for building Carla on Ubuntu 18.04 works for 20.04 as well except for dropping Python 2.7 related dependencies.
Like so:
```
sudo apt-get install build-essential clang-8 lld-8 g++-7 cmake ninja-build libvulkan1 python3 python3-pip python3-dev libpng-dev libtiff5-dev libjpeg-dev tzdata sed curl unzip autoconf libtool rsync libxml2-dev &&
pip3 install --user setuptools
```

Build scripts are also modified to remove Python 2.7 targets.
One could manually install Python 2.7 and retain support for it but it's probably better to just let go of it.

P.S. No need to add Xenial or Bionic deb repository, which is the first command in the build instructions in the [docs](https://carla.readthedocs.io/en/latest/build_linux/), i.e,
```
sudo apt-get update &&
sudo apt-get install wget software-properties-common &&
sudo add-apt-repository ppa:ubuntu-toolchain-r/test &&
wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add - &&
sudo apt-add-repository "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" &&
sudo apt-get update
```
I was able to build Carla on Ubuntu 20.04 with only Focal Fossa deb repository.

#### Where has this been tested?

  * **Platform(s):** Ubuntu 20.04
  * **Python version(s):** 3.8
  * **Unreal Engine version(s):** 4.24

#### Possible Drawbacks

* To run scripts that use pygame, need to use pygame developent version 2.0.0.dev6 as 1.9.6 can't be installed from pip on Python 3.8.

``` pip3 install pygame==2.0.0.dev6  ```

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2787)

Issue #2752, opened 2020-04-16
docstrings for carla


HarryWindsor:
Is is possible to include the docs from https://carla.readthedocs.io/en/latest/python_api/#carla.Actor in the python library? I would prefer using the docs in my IDE rather than jumping into the browser when I have to look something up.

Thank you.

pperle:
You could try using [Pycharms external documentation](https://www.jetbrains.com/help/pycharm/settings-tools-python-external-documentation.html#) feature. The URL should look something like that `https://carla.readthedocs.io/en/latest/python_api/#carla.{class.name}.{function.name}`.

sergi-e:
Hi, @HarryWindsor !

Did the solution provided by @pperle served you well?

Alby407:
This does not work. PyCharm won't give any information when for example hovering over `carla.World`. If I manually go to e.g. `https://carla.readthedocs.io/en/latest/python_api/#carla.World`, it works.

pperle:
@Alby407 you are right, the URL I posted does not work for classes, only for their functions. For classes, there is one dot to many `https://carla.readthedocs.io/en/latest/python_api/#carla.World.`.
I think the right parameter to use would be `{element.qname} - full-qualified name of element under caret`. The problem is that the fully-qualified name is not `carla.World` but `carla.libcarla.World` which does not match the doc URL (and even this would not fix the problem, looks like PyCharm has problems parsing the docs).

The right solution would be to include the docs.

pperle:
Inspired by [this post](https://stackoverflow.com/a/34898716/9309705) I tried to add the docstring from `carla/PythonAPI/docs/*.yml` to the python stub. You can find the code here: https://github.com/pperle/carla/tree/docstrings_py . It is hacky, I know but it works well enough for me at the moment.

If anyone is interested here is the compiled Linux version [carla-0.9.9-py3.5-linux-x86_64.egg.zip](https://github.com/carla-simulator/carla/files/4565282/carla-0.9.9-py3.5-linux-x86_64.egg.zip). The only thing that does not work are the type hints.

Example preview in PyCharm:
![image](https://user-images.githubusercontent.com/17356275/80822338-ca37b000-8bda-11ea-8994-26777a0a7b3f.png)

germanros1987:
@pperle you have ran 90% of this marathon...would you like to finish it by making a PR to CARLA?

BruceXSK:
[...]

PR #2804, opened 2020-05-01
documentation (docstrings) in python

pperle:
#### Description

This pull request adds the docstring from https://carla.readthedocs.io/en/latest/python_api/ into the python source code, ad discussed in #2752. This is still in an early stage, not all references are resolvable, and there is a problem with overloaded functions for which it is not clear which doctrine to choose. Nevertheless, I publish the pull request in order to draw more attention to this topic.

I hope this improves your experience with Carla, you can find a precompiled egg file for Linux (for Carla 0.9.9) [here](https://github.com/carla-simulator/carla/files/4565282/carla-0.9.9-py3.5-linux-x86_64.egg.zip).

  * **Platform(s):** Linux Ubuntu 18.04.4 LTS
  * **Python version(s):** 3.5.6

#### Possible Drawbacks

This is still very hacky, not all references are resolvable (sometimes this is because of the naming schema and sometimes because there is no documentation). I have only updated the python3 part of `BuildPythonAPI.sh` because I only have a Linux machine to run Carla and I do not see a reason to support python2 (won't the support be dropped anyway with #2787?).

Please have a look where the documentation doesn't look correct, I guess there are some edge cases I did not cover.


marcgpuig:
Hi @pperle, that sounds really great!
Which is the advantage of this instead of using Python stubs? These can be also generated from our `.yaml` docs, I've made some experiments in the past and looks like it can be done, but I've never found the time to properly work on it.

pperle:
It has no advantage in using this instead of Python stubs. This is only the first solution I came up with.

marcgpuig:
Maybe @bernatx and @nsubiron could join the discussion
My concern is that using stubs should be better since it doesn't require modifying all the Python bindings at all. Your solution could be temporal until we have time to rewrite this system, what do you think?

PR #2827, opened 2020-05-07
Expose the curvature of the road

daedric:
Checklist:

  - [x] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [x] Code compiles correctly
  - [x] All tests passing with `make check`

#### Description

This PR adds functions to get the curvature of a lane at a given waypoints as well as some utility functions.

  * Ubuntu 18.04 - 19.10
  * Python 3.7
  * UE 4.22 + UE 4.24 (Since Carla 0.9.6)

#### Possible Drawbacks

None as it only adds functions and nothing existing has been changed

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2827)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
   * Added parameter to enable/disable pedestrian navigation in standalone mode
   * Improved mesh split in standalone mode
   * Fixed large RAM usage when loading polinomial geometry from OpenDRIVE
-
+  * Added API functions to get the lane curvature
 
 ## CARLA 0.9.9
 
--- a/LibCarla/source/carla/client/Geometry.cpp
+++ b/LibCarla/source/carla/client/Geometry.cpp
@@ -32,5 +32,17 @@ namespace client {
     return _geometry.GetCurvature(s);
   }
 
+  double Geometry::GetLength() const {
+    return _geometry.GetLength();
+  }
+
+  double Geometry::GetStartOffset() const {
+    return _geometry.GetStartOffset();
+  }
+
+  double Geometry::GetHeading() const {
+    return _geometry.GetHeading();
+  }
+
 } // namespace client
 } // namespace carla
--- a/LibCarla/source/carla/client/Geometry.cpp
+++ b/LibCarla/source/carla/client/Geometry.cpp
@@ -1,0 +1,28 @@
+// Copyright (c) 2020 Computer Vision Center (CVC) at the Universitat Autonoma
+// de Barcelona (UAB).
+//
+// This work is licensed under the terms of the MIT license.
+// For a copy, see <https://opensource.org/licenses/MIT>.
+
+#include "carla/client/Geometry.h"
+
+#include "carla/road/element/Waypoint.h"
+#include "carla/road/element/Geometry.h"
+#include "carla/road/element/RoadInfoGeometry.h"
+#include "carla/client/Map.h"
+
+namespace carla {
+namespace client {
+
+  static const road::element::Geometry& GeometryAt(const Map& map, const road::element::Waypoint& waypoint) {
+    const auto &lane = map.GetMap().GetLane(waypoint);
+    const auto &road = lane.GetRoad();
+    const auto& geom = road->GetInfo<road::element::RoadInfoGeometry>(waypoint.s);
+    return geom->GetGeometry();
+  }
+
+  Geometry::Geometry(SharedPtr<const Map> parent, const road::element::Waypoint& waypoint)
+    : _parent(std::move(parent)), _geometry(GeometryAt(*_parent, waypoint)) {}
+
+} // namespace client
+} // namespace carla
--- a/LibCarla/source/carla/client/Geometry.cpp
+++ b/LibCarla/source/carla/client/Geometry.cpp
@@ -24,5 +24,13 @@ namespace client {
   Geometry::Geometry(SharedPtr<const Map> parent, const road::element::Waypoint& waypoint)
     : _parent(std::move(parent)), _geometry(GeometryAt(*_parent, waypoint)) {}
 
+  road::element::GeometryType Geometry::GetType() const {
+    return _geometry.GetType();
+  }
+
+  double Geometry::GetCurvature(double s) const {
+    return _geometry.GetCurvature(s);
+  }
+
 } // namespace client
 } // namespace carla
--- a/LibCarla/source/carla/client/Geometry.cpp
+++ b/LibCarla/source/carla/client/Geometry.cpp
@@ -7,60 +7,86 @@
 #include "carla/client/Geometry.h"
 #include "carla/client/Waypoint.h"
 
-#include "carla/road/element/Waypoint.h"
+#include "carla/client/Map.h"
 #include "carla/road/element/Geometry.h"
 #include "carla/road/element/RoadInfoGeometry.h"
-#include "carla/client/Map.h"
+#include "carla/road/element/Waypoint.h"
 
 namespace carla {
 namespace client {
 
-  static const road::element::Geometry& GeometryAt(const Map& map, const road::element::Waypoint& waypoint) {
-    const auto &lane = map.GetMap().GetLane(waypoint);
-    const auto &road = lane.GetRoad();
-    const auto& geom = road->GetInfo<road::element::RoadInfoGeometry>(waypoint.s);
-    return geom->GetGeometry();
-  }
-
-  Geometry::Geometry(SharedPtr<const Map> parent, const road::element::Waypoint& waypoint)
-    : _parent(std::move(parent)), _geometry(GeometryAt(*_parent, waypoint)) {}
+  Geometry::Geometry(SharedPtr<const Map> parent, const road::Road& road, const road::element::Geometry& geo)
+    : _parent(std::move(parent)), _road(road), _geometry(geo) {}
 
   road::element::GeometryType Geometry::GetType() const {
     return _geometry.GetType();
   }
-
   double Geometry::GetCurvatureFromWaypoint(const Waypoint& w) const {
-    auto& other_geometry = GeometryAt(*_parent, {
-                                                    w.GetRoadId(),
-                                                    w.GetSectionId(),
-                                                    w.GetLaneId(),
-                                                    w.GetDistance(),
-                                                });
-    if (std::addressof(other_geometry) != std::addressof(_geometry)) {
-      throw std::runtime_error("Waypoint is on another geometry");
+    if (_road.GetId() != w.GetRoadId()) {
+      throw std::runtime_error("waypoint is not on the road");
     }
-
-    auto s = w.GetDistance() - _geometry.GetStartOffset();
-    return GetCurvature(s);
+    return GetCurvature(w.GetDistance());
   }
-
   double Geometry::GetCurvature(double dist) const {
+    dist = dist - _geometry.GetStartOffset();
     if (dist < 0 || dist > _geometry.GetLength()) {
-      throw std::invalid_argument("dist must be within the geometry length");
+      throw std::out_of_range("distance is out of range for the current geometry");
     }
-    return _geometry.GetCurvature(dist);
+    return _geometry.GetCurvature(dist - _geometry.GetStartOffset());
   }
-
   double Geometry::GetLength() const {
     return _geometry.GetLength();
   }
-
   double Geometry::GetStartOffset() const {
     return _geometry.GetStartOffset();
   }
-
   double Geometry::GetHeading() const {
     return _geometry.GetHeading();
+  }
+
+  RoadGeometries::RoadGeometries(SharedPtr<const Map> parent, const road::element::Waypoint& waypoint)
+    : _parent(std::move(parent)), _road(*_parent->GetMap().GetLane(waypoint).GetRoad()) {}
+
+  road::RoadId RoadGeometries::GetRoadId() const {
+    return _road.GetId();
+  }
+
+  SharedPtr<Geometry> RoadGeometries::GetGeometryFromWaypoint(const Waypoint& w) const {
+    auto& geo = GeometryAt(w);
+    return SharedPtr<Geometry>(new Geometry(_parent, _road, geo));
+  }
+
+  SharedPtr<Geometry> RoadGeometries::GetGeometry(double dist) const {
+    auto& geo = GeometryAt(dist);
+    return SharedPtr<Geometry>(new Geometry(_parent, _road, geo));
+  }
+
+  double RoadGeometries::GetCurvatureFromWaypoint(const Waypoint& w) const {
+    auto& geom = GeometryAt(w);
+    auto s = w.GetDistance() - geom.GetStartOffset();
+    return GetCurvature(s);
+  }
+
+  const road::element::Geometry& RoadGeometries::GeometryAt(const Waypoint& waypoint) const {
+    if (_road.GetId() != waypoint.GetRoadId()) {
+      throw std::runtime_error("waypoint is not on the road");
+    }
+    auto geom = _road.GetInfo<road::element::RoadInfoGeometry>(waypoint.GetDistance());
+    return geom->GetGeometry();
+  }
+
+  const road::element::Geometry& RoadGeometries::GeometryAt(double distance) const {
+    auto geom = _road.GetInfo<road::element::RoadInfoGeometry>(distance);
+    if (!geom)
+      throw std::out_of_range("distance is out of range for the current geometry");
+    return geom->GetGeometry();
+  }
+
+  double RoadGeometries::GetCurvature(double dist) const {
+    const auto geom = _road.GetInfo<road::element::RoadInfoGeometry>(dist);
+    if (!geom) throw std::out_of_range("distance is out of range for the current geometry");
+    const auto& g = geom->GetGeometry();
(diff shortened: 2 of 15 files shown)


Axel1092:
Hi @daedric,
Thank you for this contribution. Retrieving the curvature of the road at any point is very useful in many applications and we are happy to include this to Carla. That said, I found some issues that should be addressed.
- First, I have observed that you have added a new class `CurvatureAtDistance` in `Geometry.h` to encapsulate the curvature of the road at a certain point and it is even exposed to the PythonAPI. I think this class should be removed both in c++ and Python code as it is redundant with getting the curvature from a waypoint `double Map::GetCurvature(const Waypoint waypoint)`.
- Second, I don't understand the purpose of the function ´Map::GetClosestWaypointsOnRoad´, do you want closest waypoint for different lanes/roads? What do you intend to do with this information? 
Similarly, I do not understand the need of `Map::GetCurvatureList` when you can retrieve the curvature from waypoints distributed through the road as you need, I don't think this function is necessary. The same goes for related functions `Map::GetWaypointList` and `Map::GetCurvatureList`.
- Finally, retrieving the curvature should be implemented for all geometries, I can do this for `GeometryPoly3` and `GeometryParamPoly3` which are a bit special, once the other issues are solved.

Thank you for your time.

daedric:
> Hi @daedric,
> Thank you for this contribution. Retrieving the curvature of the road at any point is very useful in many applications and we are happy to include this to Carla. That said, I found some issues that should be addressed.
> 
>     * First, I have observed that you have added a new class `CurvatureAtDistance` in `Geometry.h` to encapsulate the curvature of the road at a certain point and it is even exposed to the PythonAPI. I think this class should be removed both in c++ and Python code as it is redundant with getting the curvature from a waypoint `double Map::GetCurvature(const Waypoint waypoint)`.
> 

This class is actually used by a new C++ / Python API called GetCurvatureList (ref. carla/client/Map.h) which is meant to provide the geometry at each "inflection point" of the road.
This list is used to adapt the NV speed in function of the worst-case curvature ahead. In short, we need a way to obtain from the hdmap parser, the list of all geometry changes on the current road, not just at a given waypoint.

>     * Second, I don't understand the purpose of the function ´Map::GetClosestWaypointsOnRoad´, do you want closest waypoint for different lanes/roads? What do you intend to do with this information?

This function is used instead of the legacy GetClosestWaypointOnRoad. Indeed, within junctions, GetClosestWaypointOnRoad prove to be too limited as when entering a junction, we need to obtain all the connection roads and not just the "closest".

>       Similarly, I do not understand the need of `Map::GetCurvatureList` when you can retrieve the curvature from waypoints distributed through the road as you need, I don't think this function is necessary. The same goes for related functions `Map::GetWaypointList` and `Map::GetCurvatureList`.

I hope the answer to previous question provides the necessary context. If not, let's talk further!

> 
>     * Finally, retrieving the curvature should be implemented for all geometries, I can do this for `GeometryPoly3` and `GeometryParamPoly3` which are a bit special, once the other issues are solved.
> 

That's great, thanks !
> 
> Thank you for your time.

Axel1092:
Hi,
Thank you for your answer I think I understand your goals now. I will comment on every point individually.

> This function is used instead of the legacy GetClosestWaypointOnRoad. Indeed, within junctions, GetClosestWaypointOnRoad prove to be too limited as when entering a junction, we need to obtain all the connection roads and not just the "closest".

This can also be solved using the existing class `carla.Junction` in the PythonAPI, from which you can obtain waypoints for every road within the junction, making this function unnecessary.

> This class is actually used by a new C++ / Python API called GetCurvatureList (ref. carla/client/Map.h) which is meant to provide the geometry at each "inflection point" of the road.
This list is used to adapt the NV speed in function of the worst-case curvature ahead. In short, we need a way to obtain from the hdmap parser, the list of all geometry changes on the current road, not just at a given waypoint.

I understand that you want the curvature for every change in the geometry of the road. This is interesting, however we would like to expose this information information in a different way. We prefer exposing to the Python API the c++ classes `Geometry` and their children. The idea would be, from a waypoint to add a function get geometries that returns a list of these every geometry in the road. This would let you get the curvature of the road at the ends of the geometry. We are still thinking what would be the best way to expose this information while taking into account the lane shifts. Any ideas about this are always welcome. In addition, I think your approach may encounter problems when dealing with Cubic curves supported by OpenDRIVE as the inflection point may be in the middle of the road. To solve this, maybe the functions `next_until_lane_end` and `previous_until_lane_start` from waypoint may be helpful for you to sample the curvature at different points in the road.

Therefore, for this pull request, we prefer keeping only the functions in the c++ class `Geometry` family and querying it from waypoint. We do not want to introduce functions and classes that may be subject to further change as we want to keep the API as final as possible.

daedric:
> [...]
> This can also be solved using the existing class `carla.Junction` in the PythonAPI, from which you can obtain waypoints for every road within the junction, making this function unnecessary.

Indeed thanks to the addition of the Junction class, it is now possible to implement this function in python. so we are good indeed. Thanks!

> [...]
> I understand that you want the curvature for every change in the geometry of the road. This is interesting, however, we would like to expose this information in a different way. We prefer exposing to the Python API the c++ classes `Geometry` and its children. The idea would be, from a waypoint to add a function `get_geometries` that returns a list of these every geometry in the road. This would let you get the curvature of the road at the ends of the geometry. We are still thinking about what would be the best way to expose this information while taking into account the lane shifts. Any ideas about this are always welcome. In addition, I think your approach may encounter problems when dealing with Cubic curves supported by OpenDRIVE as the inflection point may be in the middle of the road. To solve this, maybe the functions `next_until_lane_end` and `previous_until_lane_start` from waypoint may be helpful for you to sample the curvature at different points in the road.
> 
> Therefore, for this pull request, we prefer keeping only the functions in the c++ class `Geometry` family and querying it from the waypoint. We do not want to introduce functions and classes that may be subject to further change as we want to keep the API as final as possible.

 Thanks for the suggestions, looks good overall.

Let us summarize what we understand would need to be added:

* Waypoint class:
   * add a geometry property returning a Geometry instance.
   * add a `next_geometries(self, distance)` returning a list of every geometry ends in the road up. the distance parameter would allow to bound the list up to a distance from the waypoint. And then indeed for parametric cubic curves, the client would have to use `next_until_lane_end` to get an estimate of the worst-case curvature.
* Geometry class: would expose same API as the C++ Geometry class (including GetCurvature) 


Is this proposal aligned with what you suggested?

Would you want us to update the curvature branch with above changes?

Axel1092:
Yes, it would be really awesome if you could do that! This would be an excellent addition to the API.
As promised, here are the functions of the curvature for the spline geometries:
```
  double GeometryPoly3::GetCurvature(double dist) const {
    auto result = _rtree.GetNearestNeighbours(
        Rtree::BPoint(static_cast<float>(dist))).front();

    auto &val1 = result.second.first;
    auto &val2 = result.second.second;

    double rate = (val2.s - dist) / (val2.s - val1.s);
    double u = rate * val1.u + (1.0 - rate) * val2.u;
    double v = _poly.Evaluate(u);
    double t_v = _poly.Tangent(u);
    double sqr_uv = u*u + v*v;
    double sqrt_uv = sqrt(sqr_uv);
    double dT_x = 1.0/sqrt_uv - u*(u + v *t_v)/sqrt(sqr_uv*sqr_uv*sqr_uv);
    double dT_y = t_v/sqrt_uv - v*(u + v *t_v)/sqrt(sqr_uv*sqr_uv*sqr_uv);

    return sqrt(dT_x*dT_x + dT_y*dT_y) / sqrt(1.0 + t_v*t_v);
  }

  double GeometryParamPoly3::GetCurvature(double dist) const {
    auto result = _rtree.GetNearestNeighbours(
        Rtree::BPoint(static_cast<float>(dist))).front();

    auto &val1 = result.second.first;
    auto &val2 = result.second.second;

    double rate = (val2.s - dist) / (val2.s - val1.s);
    double u = rate * val1.u + (1.0 - rate) * val2.u;
    double v = rate * val1.v + (1.0 - rate) * val2.v;
    double t_u = (rate * val1.t_u + (1.0 - rate) * val2.t_u);
    double t_v = (rate * val1.t_v + (1.0 - rate) * val2.t_v);
    double sqr_uv = u*u + v*v;
    double sqrt_uv = sqrt(sqr_uv);
    double dT_x = t_u/sqrt_uv - u*(u*t_u + v *t_v)/sqrt(sqr_uv*sqr_uv*sqr_uv);
    double dT_y = t_v/sqrt_uv - v*(u*t_u + v *t_v)/sqrt(sqr_uv*sqr_uv*sqr_uv);

    double sqr_tuv = t_u*t_u + t_v*t_v;
    if(sqr_tuv < std::numeric_limits<double>::epsilon()) {
      return 0.0;
    } else {
      return sqrt(dT_x*dT_x + dT_y*dT_y) / sqrt(t_u*t_u + t_v*t_v);
    }
  }
```
It would be great if you can include these in your PR.

daedric:
Thanks for the clarification :) I'll get on it this week hopefully !

Axel1092:
By the way, let me know when you are done with your changes so that I can start to review them.

daedric:
Will do, I've been busy on other issues unfortunately. I hope to go back to this pull request next week.

I've a design issue I need to write about so we can discuss it. I'll try to do the write up tomorrow :)

Cheers,

daedric:
Hello,
As previously said I am uncomfortable with the approach we're currently taking.
While it does make sense to have a starting point via the Waypoint to get the
geometry I do wonder if this is actually a good idea.


### The problem

First let's layout the facts:

* One road have potentially several geometries;
* A waypoint represent a specific location in a (road, lane, s);
* For a Waypoint, `s` is an offset from the beginning of the road;
* A geometry has among others a start offset from the road beginning and a length;


The current proposition is to get the road Geometry via a Waypoint instance.
To further get a curvature at a given location, there are two idiomatic ways
given the Carla API: either you get a distance or a waypoint.

If you get a Waypoint, there is no guarantee that the given waypoint actually
points onto the same geometry. As a result, the API has to check whether this is
the case and raise an error if need be. With the distance, it is quite natural
to write:

    wp = ......
    geom = wp.get_geometry()
    curvature_at_wp = geo.get_curvature(wp.s)

But this is wrong, the right code would be:

    curvature_at_wp = geo.get_curvature(wp.s - geo.get_start_offset())

Currently, I've implemented a sub efficient (IMO) way of getting the curvature
from a geometry:

    waypoint = # ....
    geo = waypoint.get_geometry()
    geo.get_curvature_from_waypoint(waypoint)

And a get_curvature taking a distance from the beginning of the geometry.

The usage of the API also feels like there is a circular dependency.

### Alternative

What I'd suggest is that since the Geometry are more or less PODs to expose all
fields of all the Geometry class hierarchy.

Algorithms such as the GetCurvature does not necessarily need to be accessible
from the Geometry instance itself.  I would rather have the curvature directly
accessible within the waypoint along a reference on the Geometry for people
needing the actual geometry.

It would work quite well as someone could sample the curvature by iterating
over the waypoints and be error-proof as the underlying geometry while exposed
does not require one to know too much about how it actually works to get
something as simple as the curvature.

Not forgetting that the curvature _is_ a property of the location pointed by the
waypoint so we could see the curvature as actually a missing data.

wdyt ?

Axel1092:
Hi,
I see your point, this requires some extra thought. I'll share some ideas to solve these issues and let me know what you think.
- Regardless of what we do with the `Geometry` class I would keep the `get_curvature` function from the waypoint.
- About the geometry, an alternative to try to solve these issues would be to change the `Geometry` class, so that it represents the collection of geometries of the entire road in a single object. 
The idea would be to have code like `pos = geometry.get_location(waypoint.s)` to work as we expect and let the c++ side will figure out to what geometry type the 's' corresponds and retrieve the correct point. Under the hood I think this can be easily implemented by just keeping a reference to the road object and then use `road->GetInfo<road::element::RoadInfoGeometry>(s)` to do the queries. 
So, instead of `Geometry` representing a line, spiral, etc, it would represent the entire road.
Then we could have additional functions to detect the geometry type details such as `type = geometry.get_type_at(s)` and `list = geometry.get_type_changes()` (a different name would be better) to detect at what s-position the geometry type changes.
I think this would make an easier interface and better code in the python side for people wanting additional details of the geometry of the road and easily enable functionality that you needed in the first version, the inflection points.

Let me know what you think about these ideas if it is useful or any other thoughts you may want to share (alternatives, something unclear, difficulties, etc).

daedric:
>     * Regardless of what we do with the `Geometry` class I would keep the `get_curvature` function from the waypoint.
> 

:+1: 

>     * About the geometry, an alternative to try to solve these issues would be to change the `Geometry` class, so that it represents the collection of geometries of the entire road in a single object.
>       The idea would be to have code like `pos = geometry.get_location(waypoint.s)` to work as we expect and let the c++ side will figure out to what geometry type the 's' corresponds and retrieve the correct point. Under the hood I think this can be easily implemented by just keeping a reference to the road object and then use `road->GetInfo<road::element::RoadInfoGeometry>(s)` to do the queries.
>       So, instead of `Geometry` representing a line, spiral, etc, it would represent the entire road.
>       Then we could have additional functions to detect the geometry type details such as `type = geometry.get_type_at(s)` and `list = geometry.get_type_changes()` (a different name would be better) to detect at what s-position the geometry type changes.
>       I think this would make an easier interface and better code in the python side for people wanting additional details of the geometry of the road and easily enable functionality that you needed in the first version such as the inflection points.

Fair enough, it is indeed better but I am not too convinced yet on the advantages over having the curvature within the Waypoint itself :)

I'll implement the suggested change though as a new base for discussion :)

> Let me know what you think about these ideas if it is useful or any other thoughts you may want to share (alternatives, something unclear, difficulties, etc).

Axel1092:
> I am not too convinced yet on the advantages over having the curvature within the Waypoint itself :)

You mean leaving that function only within the `Geometry` class alone? Yeah, maybe it is better to have it in a single place.

```
geometry = waypoint.get_road_geometry()
curvature1 = geometry.get_curvature(waypoint.s)
curvature2 = geometry.get_curvature(waypoint.s + delta)
```
An advantage to query curvature in the geometry object is that we are not generating a waypoint which has some additional cost which may be an issue if we are sampling a lot of points.

Thank you for your efforts, as we discuss this I think we will manage to get a cool feature here.

daedric:
> > I am not too convinced yet on the advantages over having the curvature within the Waypoint itself :)
> 
> You mean leaving that function only within the `Geometry` class alone? Yeah, maybe it is better to have it in a single place.
> 
> ```
> geometry = waypoint.get_road_geometry()
> curvature1 = geometry.get_curvature(waypoint.s)
> curvature2 = geometry.get_curvature(waypoint.s + delta)
> ```

This is what I currently have now (not yet pushed).
While it does improve the usability, you are able to request a curvature outside of the range (I do raise an error in this case) and you need a bit of work when your next waypoint is outside of the road to get the new "current road" geometry.
 
> An advantage to query curvature in the geometry object is that we are not generating a waypoint which has some additional cost which may be an issue if we are sampling a lot of points.

I understand that's your concern and this is also a concern of mine. I am just not sure how to weight a safe API versus an efficient one in this precise case :)

> 
> Thank you for your efforts, as we discuss this I think we will manage to get a cool feature here.

That's no problem at all :)

Axel1092:
> While it does improve the usability, you are able to request a curvature outside of the range (I do raise an error in this case) and you need a bit of work when your next waypoint is outside of the road to get the new "current road" geometry.

Maybe the `Geometry` can keep a `road_id` field so that the user can check it with the waypoint's `road_id` attribute.

Let me know of you have a specific API usage as you work on implementing this :)

daedric:
> Maybe the `Geometry` can keep a `road_id` field so that the user can check it with the waypoint's `road_id` attribute.
This is what I do :)
> 
> Let me know of you have a specific API usage as you work on implementing this :)
Will do !

daedric:
I've pushed the change we talked about. We will work on our code to migrate onto this API so that we see whether it actually works :)

Cheers,

33v33:
Hello @Axel1092 
We encountered another issue that we would love to get your feedback on.
The issue is that when an actor is in a junction, we would like to get the list of the closest waypoints to the actor's location, not just the closest waypoint as returned by `carla.Map`'s `get_waypoint(self, location, project_to_road=True, lane_type=carla.LaneType.Driving)`.
Our usecase for such a feature is the following, suppose we come into the following junction:
junction
in the indicated area, as the vehicle oscillates a bit from the center of the lane, we can get either a closest waypoint on either lane.

Axel1092:
Hi @33v33,
Returning a number of closest waypoints is not properly defined continuously along the road. However, you need a list of waypoints for every road in a junction. You can obtain this information through the junction object: https://carla.readthedocs.io/en/latest/python_api/#carla.Junction
```
waypoint = carla_map.get_waypoint(location)
junction = waypoint.get_junction()
waypoint_list = junction.get_waypoints()
```
In this case `waypoint_list` will contain a list of pairs of waypoints (`waypoint_list = [(w_start,w_end),(w_start,w_end),(w_start,w_end),...]`) where every pair is the start and end of a lane in the junction and the whole list contains all the lanes in the junction. You can then sort these waypoints by distance and obtain what you need.

Axel1092:
Hi @daedric,
How is it matching the new API? I left a comment with an improvement for the geometry class that should provide the information we both needed. Let me know your opinion.

PR #2885, opened 2020-05-25
add support for additional TraCI clients

patmalcolm91:
#### Description

Adds support for the Sumo parameter "num-clients" via a command-line parameter "--additional-traci-clients". This allows additional TraCI scripts to be run in parallel with the co-simulation, a common use-case in the field of transportation research.

  * **Platform(s):** Linux
  * **Python version(s):** 3.7.7

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2885)


joel-mb:
Hi @patmalcolm91 
Thank you very much for this contribution!

I think this is a great addition to the integration. However, I think that it would be necessary to add a few things more to completety be able to use additional multiple SUMO clients.

As far I've seen it is needed to explicitly specify the execution order of the clients using the `SetOrder` command when using multiple traci clients. Therefore, I think it would be needed to add a `--client-order` parameter to the run synchronization script to let the user specify it.

I think it would be great if you can add these modifications to the PR!

patmalcolm91:
> Hi @patmalcolm91
> Thank you very much for this contribution!
> 
> I think this is a great addition to the integration. However, I think that it would be necessary to add a few things more to completety be able to use additional multiple SUMO clients.
> 
> As far I've seen it is needed to explicitly specify the execution order of the clients using the `SetOrder` command when using multiple traci clients. Therefore, I think it would be needed to add a `--client-order` parameter to the run synchronization script to let the user specify it.
> 
> I think it would be great if you can add these modifications to the PR!

Hi @joel-mb, thanks for the comment. While this ```traci.setOrder()``` call is not strictly necessary at least in some cases (I tested the code successfully without it), I agree that this is a useful parameter to include. I've made the corresponding changes.

joel-mb:
> > Thanks for the modification! I've tested it and works perfectly.
> > Just one last thing, could you add the parameter `--client-order` to the run_syncrhonization script as well? Just adding the parameter and modifying the call to SumoSimulation (line `233`) to add the new argument should be enough. I tried to do the modifications myself but I had a weird error when pushing the commit to the PR.
> > Remember to update the CHANGELOG as well.
> > Thanks!!!
> 
> Ah, yes, I forgot about the run_synchronization script. I've added the `--client-order` parameter as requested. As I'm sure you're aware, the `--additional_clients` parameter doesn't make sense in this script, as the number of total TraCI clients can be specified directly in the sumocfg file using the `<num-clients value="x"/>` tag. Changelog has also been updated.

Exactly! Everything seems ready to be merged. We will do so during today.
Thank you very much for your contribution!

PR #2907, opened 2020-06-02
Fix inconsistency in the doc with the actual API.

squizz617:
#### Description
`spawn_actor()` raises an exception, and `try_spawn_actor()` returns
`None` on failure. However, the documentation swapped the descriptions
of those two. This commit simply fixes the documentation.

  * **Platform(s):**  Ubuntu 18.04.4, kernel 5.3.0-46-generic
  * **Python version(s):** Python 3.6.9

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2907)

### DIFF
--- a/Docs/core_actors.md
+++ b/Docs/core_actors.md
@@ -70,8 +70,8 @@ __The world object is responsible of spawning actors and keeping track of these.
 
 The world has two different methods to spawn actors.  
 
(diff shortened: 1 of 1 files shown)


bernatx:
Thank you for the fix squizz617, we appreciate.

PR #2929, opened 2020-06-08
Fix for slow TrafficManager setup on large XODR only maps.

pravinblaze:
#### Description

Reverting back to using string keys to identify spatial grids as we found it empirically faster
and solves the problem of TrafficManager setup taking too long in case of large .xodr only maps.

  * **Python version(s):** 3.8

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2929)

### DIFF
--- a/LibCarla/source/carla/trafficmanager/InMemoryMap.cpp
+++ b/LibCarla/source/carla/trafficmanager/InMemoryMap.cpp
@@ -10,7 +10,6 @@
 namespace carla {
 namespace traffic_manager {
 
-
   namespace cg = carla::geom;
   using namespace constants::Map;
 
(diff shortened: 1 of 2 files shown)


doterop:
Is almost impossible that string keys are faster than int keys.

marcgpuig:
Instead of this, we could try to use boost spatial trees, since they are working really well in the new WaypointAPI.

pravinblaze:
> @pravinblaze Is almost impossible that string keys are faster than int keys.

@doterop, I know it is strange but we did measure, averaging time over a large number of samples. It seems, two bitwise ORs and 32 left shifts are more expensive (1.5us) than constructing a string of 3-7 characters with two concentrations (0.4us). Anyways, we have been wanting to use spatial index trees instead of hash buckets for a long time, we will replace this.


> Instead of this, we could try to use boost spatial trees, since they are working really well in the new WaypointAPI.

Hey @marcgpuig, absolutely. Spatial index trees are much better for what we are trying to do and boost R-trees are super easy to use.

Closing this PR as we have raised PR#2933 with boost r-tree for spatially indexing and querying waypoints.

PR #2931, opened 2020-06-08
added cpp example for spawning + bug fix in Simulator

brezhnyev:
#### Description

Bug in the Simulator.cpp when loading new scene leads to the situation where the walkers were not "seen" by the cars causing too many hits of cars with walkers. The line
_episode.reset();
is needed to update the Episode id
otherwise the IsAlive() for walkers returned false in LocalizationStage::Action() (~line 289)
The DEBUG_ASSERT(_crowd != nullptr);
should not be present since _crowd is initialized right after it and this causes unjustified termination in the debug mode.
Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2931)

### DIFF
--- a/Examples/CppClientsCmake/example_01/CMakeLists.txt
+++ b/Examples/CppClientsCmake/example_01/CMakeLists.txt
@@ -1,0 +1,25 @@
+# standard version  3.5.2 for Ubuntu 16
+# statndard version 3.10.2 for Ubuntu 18
+cmake_minimum_required (VERSION 3.5)
+
+project (carla_client_sync)
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../../../LibCarla/source)
+
+link_directories(${CMAKE_CURRENT_SOURCE_DIR}/../../../Build/LibCarla/cmake/client)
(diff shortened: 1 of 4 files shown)


bernatx:
Hi, thanks for the fixes. The assert about walkers was fixed time ago, and now your PR conflicts with that. Can you update the PR so we can merge it?
Thanks in advance for all the effort.

brezhnyev:
Hi, bernat,
I fixed the conflict. However the CI python test fails for my PR. I have no
idea why, since I actually do not do anything with Python.
I suggest you add the missing line yourself

_episode.reset();
in the
LibCarla/source/carla/client/detail/Simulator.cpp
<https://github.com/carla-simulator/carla/pull/2931/commits/5a9fe47d312ea1a14b60780f5e1420fc4e9caec2#diff-102788e7098791711f7271f62b376fd0>
As I already mentioned in the commit description without this line, any NEW
scene loaded in the cpp client will cause frequent accidents of cars and
walkers, since the walkers are just not visible for cars.
The rest (like cpp client example) is not so very important, so you can
just remove my PR, or i can do it.

Regards
KB.

PR #2933, opened 2020-06-09
Boost R-tree for waypoint indexing and querying in TrafficManager

pravinblaze:
#### Description

This PR replaces the earlier hash buckets used for waypoint indexing and querying with Boost's R-tree.

This also solves the problem of slow TrafficManager setup on large .xodr only maps.

  * **Platform(s):** Ubunru 20.04
  * **Python version(s):** 3.8

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2933)


marcgpuig:
Yay!

PR #2965, opened 2020-06-18
Improved stream performance

Axel1092:
#### Description

This small change disables delays in the tcp communication from server to client, dramatically improving performance in synchronous mode in linux systems.

  * **Platform(s):** Ubuntu18.04, Windows
  * **Python version(s):** 2, 3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2965)


thytless:
Hi @Axel1092 , sorry to ask here (failed to find your email) but, would you please give some more explanation? For example how could I reproduce this performance improvement?

marcgpuig:
@thytless this only affects the synchronous mode, wherein Ubuntu it was limited to 20 fps, now is up to 60 (if everything else runs at 60).

PR #2998, opened 2020-06-26
Fixed gyro to report angular velocity in local sensor frame

### LINKED ISSUE #2939: IMUMeasurement gyroscope values in world-frame instead of body-frame
Hello,

First of all, thank you very much for this great simulator, I've been using it extensively for the last few days and so far it's been a great experience.
I'm currently focused on retrieving IMU data and feeding it into a GNSS/INS fusion algorithm, so the first thing is to ensure that I'm passing valid IMU data to the INS.
I'm using pre-built binaries, of version 0.9.9.1, on Ubuntu 18.04.04 LTS.

I'm encountering an issue related to the gyroscopes measurements. They seem to be output in the world frame, instead of the vehicle body frame. 
The accelerometers look ok. 
This issue can be easily reproduced by driving a vehicle in a straight line (I'm using VehiclePIDController to drive the vehicle for a few tens of meters). 
During the drive, the vehicle pitch fluctuates. This should be reflected in a corresponding Y axis rotation-rate, in the vehicle-frame. The vehicle experiences acceleration, which should be reflected mostly in X-axis acceleration, in the vehicle-frame. 
We can now drive the vehicle, in a straight line, in two different orientations:
1. Vehicle is east-bound (UE world yaw = 0[deg]). IMU-frame should be aligned with UE-frame.
2. Vehicle is south-bound (UE world yaw = 90[deg]). IMU-frame X should be UE-frame Y, IMU-frame Y is UE-frame (-X). 

In both scenarios, we would expect similar vehicle pitch profile and body-frame IMU measurements.
The actual results show a similar vehicle pitch profile and accurately reported body-frame IMU accelerometers values.
However it seems as though the gyroscopes report world-frame rotation-rates... 
We can see the issue in the following figures:
East-bound scenario has similar world/imu accelerations/rotation-rates, as expected.
South-bound scenario has different world/imu accelerations, as expected BUT similar rotation-rates, which shouldn't be the case, because the vehicle is rotated w.r.t the world.

I've retrieved the world-frame accelerations and rotation-rates using Actor's get_acceleration and get_angular_velocity() methods.

p.s I'm sorry if this is too much information for a very simple issue to describe...

Thank you very much!
Gabi Shahmayster

Vehicle heading East acceleration
Vehicle heading East gyro
Vehicle heading East pitch
Vehicle heading South acceleration
Vehicle heading South gyro
Vehicle heading South pitch

ItsTimmy:
#### Description

The gyroscope in the IMU Sensor reported angular velocity in world frame. Real IMUs would report it in the local sensor frame. This PR fixes that.

Fixes #2939. 

  * **Python version(s):** 2.7 and 3.6

#### Possible Drawbacks

None known

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/2998)


ItsTimmy:
This is marked as a draft because I still need to do some testing. That is in progress right now and I should be able to report back in a few hours.

ItsTimmy:
I did some testing, and it seems to work. But I don't have a lot of experience with Unreal, so I think it would be best for someone else to double-check.

PR #3001, opened 2020-06-27
Added configurable noise to Lidar sensor

ItsTimmy:
#### Description

Other sensors, such as the IMU and GPS, have configurable noise. This PR adds configurable noise to the Lidar sensor.

The noise is simulated by randomly perturbing each point along the vector of the laser ray. In effect, this simulates a Lidar with perfect angular positioning of each ray, but noisy distance measurements.

The scale of this noise is controlled by the new `noise_stddev` attribute. It defaults to 0, so the default behavior is identical to the old behavior.

  * **Python version(s):** 2.7, 3.6

#### Possible Drawbacks

This noise model might not be completely accurate to real life Lidar sensors, which presumably also have some noise in the angular position of each raycast. However, it is probably a good first step.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3001)


marcgpuig:
Hi @ItsTimmy 
Really well integrated, thanks for that great work!
We were discussing with @DSantosO about the performance impact of too many randoms since we're just added some more rand operations to discard points, did you experienced some frame drop?
I'll test your code soon but so far looks good to me!

ItsTimmy:
For my needs, I always run Carla in synchronous, fixed time-step mode (0.005s time-step), and because my laptop is fairly low-powered, it runs very slow (~1/15 real time). So I don't get dropped frames, and I haven't measured the actual performance impact of this. Sorry that I can't be more helpful here!

marcgpuig:
Great! Thanks for the update!
Can you rebase master that will fix an issue related to Travis?

ItsTimmy:
I rebased, but Travis is still failing. Is there something I can do to fix this, or is this an unrelated issue?

marcgpuig:
@ItsTimmy
Don't worry, Travis is failing everywhere because some dependency we have to fix in Python3.
I'm currently testing your code to merge it.

PR #3006, opened 2020-06-29
Added timeout function to spawn_npc.py script. Resolves #2999.

### LINKED ISSUE #2999: Timeout parameter for spawn_npc.py
I run in synchronous mode and I develop carla scripts and such using a lower power computer, review the results and then run on a more powerful machine when it becomes available. I have been git stash / applying a change to allow a timeout parameter to be passed for the wait_for_tick so it is above the default of 10.0s. 

I've put this into a branch and can push it. It would be good to have it as part of the repo. 
I did not see anything related to the documentation that requires modification for this update but I may have missed it.

faultfactory:
Resolves Issue #2999 to add timeout parameter to spawn_npc.py.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3006)


faultfactory:
Will resubmit shortly. Need to clear out author info.

PR #3007, opened 2020-06-29
Allow timeout value to be passed to spawn_npc.py from command line, Resolves #2999

### LINKED ISSUE #2999: Timeout parameter for spawn_npc.py

faultfactory:
#### Description

Fixes #2999 
Timeout is now a command line parameter for the script. 

Tested on Ubuntu 18.04
Python 2.7
Using the Carla runtime. Should not affect builds. 

#### Possible Drawbacks

This is added functionality, should not create issues elsewhere

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3007)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   * Fixed delay in the tcp communication from server to client, improving performance in synchronous mode in linux systems
   * Fixed large RAM usage when loading polinomial geometry from OpenDRIVE
   * Fixed collision issues when debug draw(debug.draw_line) is called 
+  * Added timeout parameter to spawn_npc.py script. 
 
 ## CARLA 0.9.9
 
--- a/PythonAPI/examples/spawn_npc.py
+++ b/PythonAPI/examples/spawn_npc.py
@@ -91,10 +91,10 @@ def main():
         help='Enanble car lights')
     argparser.add_argument(
         '--timeout',
-	metavar='Tout',
-	default=10.0,
-	type=float,
-	help='Override timeout settings (default: 10.0)')
+        metavar='Tout',
+        default=10.0,
+        type=float,
+        help='Override timeout settings (default: 10.0)')
     args = argparser.parse_args()
 
     logging.basicConfig(format='%(levelname)s: %(message)s', level=logging.INFO)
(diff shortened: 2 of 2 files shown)


jackbart94:
Hi,

Thanks for opening this PR! Just a few things:

-  The client timeout is different than the wait_for_tick timeout. They are meant for two different things, so they should be separated.

- I don't think that having a timeout in wait_for_tick is as useful as one might think. Since it's the server who sends the tick, why should we mess with that unless the server goes down? This is also the reason why we have such an high threshold for it.

- But, if we really want to have this for wait_for_tick, let's make sure it's in all of them for this particular script, not only in the final one, and why not also put it in the tick since that would make a little more sense in my opinion.

faultfactory:
@jackbart94  Thanks for the review. I left some comments to get clarification before I push a change so I get it right 

> * The client timeout is different than the wait_for_tick timeout. They are meant for two different things, so they should be separated.
Understood, I will add a separate parameter for that. 

> * I don't think that having a timeout in wait_for_tick is as useful as one might think. Since it's the server who sends the tick, why should we mess with that unless the server goes down? This is also the reason why we have such an high threshold for it.
The line I added this timeout argument on was the one that I had issues with timing out. The 10s is high but my code was running but exceeding the 10s threshold and the error message would point to that line. Adding the timeout kept everything running and let me debug things further. Perhaps there were hangs/bugs/lags that have been since fixed making the 10 second timeout appropriate.  When i'm using this timeout parameter  I am running the system with 5 radar, 1 lidar and 6 cameras on a Geforce 1030 so each loop takes a long time. 


> * But, if we really want to have this for wait_for_tick, let's make sure it's in all of them for this particular script, not only in the final one, and why not also put it in the tick since that would make a little more sense in my opinion.
I can do that. I was unsure if it had an effect since this was not in synchronous mode.

PR #3013, opened 2020-07-01
New lidar intensity

DSantosO:
#### Description

This PR improves the LiDAR raycast sensor with two major new features:
- A value associated to each point cloud with an approximation of the intensity, right now only based on distance
- The addition of drop off lidar points to increase the realism. The percentage of points dropped can be modeleed
  in two different ways: 
      - A general value method: this way also improves performance because the points are dropped before even been raycasted.
      - A intensity-based method: with two new parameters we can linearly drop points based in the intensity in the proportion that             we are interested.

  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** 4:24

#### Possible Drawbacks

With this change, the output of the sensor changes to incorporate the intensity to each cloud point. 
Any script that uses this sensor needs to change to accommodate to this. You can check manual_control.py or
raycast_sensor_testing.py examples to see an example of this small but needed change.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3013)


marcgpuig:
The new LiDAR looks awesome:

image
image

zeyu-hello:
This function seems not been called by any other function?

PR #3061, opened 2020-07-15
Fixed import pipeline

marcgpuig:
#### Description
Fixed the import pipeline due to changes in the Unreal Engine 4.24 code.

  * **Python version(s):** 3.7

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3061)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   * Added vehicle light and street light data to recorder
   * Added API function `add_angular_impulse()` to add angular impulse to any actor
   * Fixed rain drop spawn issues when spawning camera sensors
+  * Fixed assets import pipeline
   * Fixed Update.sh from failing when the root folder contains a space on it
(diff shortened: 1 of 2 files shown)


papabst:
hey, does this fix the issue where the import command won't look for CARLA assets in all folders? I noticed that I have to manually change this in the Import Options dialog in the UE editor because otherwise _make import_ will create new materials

marcgpuig:
@papabst what do you mean by:
> CARLA assets in all folders

Can you elaborate a bit on your problem?

papabst:
When using _make import_ with a vanilla UE editor it will create new blank materials with names of existing ones because in the material options of the UE FBX import commandlet (https://docs.unrealengine.com/en-US/Engine/Content/Importing/FBX/ImportOptions/index.html#materialoptions) the default Search Location is set to Local. This must be set to Under Root or All Assets to look in the Content directory of the CARLA project. 
I do not know where this option is actually stored but I was able to test and prove that changes persist when restarting UE editor

PR #3169, opened 2020-08-06
Remove unnecessary resource locking by reducing the level of parallel…

elvircrn:
…ism in the inner loop.

Checklist:

  - [x] Your branch is up-to-date with the `master` branch and tested with latest changes
  - [x] Code compiles correctly
  - [ ] All tests passing with `make check`:

```
[  FAILED  ] 1 test, listed below:
[  FAILED  ] benchmark_streaming.image_1920x1080_mt
```
(not sure if this is caused by this change, as it also fails on master branch on my machine)

#### Description

This PR makes the parallelization more coarse (switching from task-per-point to task-per-channel) allowing me to remove the mutex for the critical section that deals with emplacement of elements, as each task now emplaces the elements into it's own vector separate vector.

Result of running `python raycast_sensor_testing.py -ln 1 -ln 0 --profiling`
Before the change: https://pastebin.com/RSxhKx90
After the change: https://pastebin.com/Y5R8v6iy

Result of running `python raycast_sensor_testing.py -sln 1 -ln 0 --profiling`
Before the change: https://pastebin.com/Hk7av3F1
After the change: https://pastebin.com/dziSeVd4

We can observe a noticeable difference in performance for the non-semantic lidar, and on-par or better performance on the semantic lidar, likely due to the fact that the hit registration is a more expensive computation for the semantic lidar (as per discussion with Daniel Santos), therefore making it so that the threads spend less time contending for the mutex since they have more work to the before getting to it.

  * **Unreal Engine version(s):** Unreal 4.24

#### Possible Drawbacks

Not only does this improve the performance, but it also simplifies the source code. Therefore, I do not see any obvious drawbacks.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3169)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastSemanticLidar.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastSemanticLidar.cpp
@@ -100,20 +100,16 @@ void ARayCastSemanticLidar::SimulateLidar(const float DeltaTime)
 
   GetWorld()->GetPhysicsScene()->GetPxScene()->lockRead();
   ParallelFor(ChannelCount, [&](int32 idxChannel) {
-
-    FCriticalSection Mutex;
-    ParallelFor(PointsToScanWithOneLaser, [&](int32 idxPtsOneLaser) {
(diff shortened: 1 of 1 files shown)


DSantosO:
Everything is good, ready to merge!
Thanks!

PR #3262, opened 2020-09-03
Fixes add entry issue for applying parameters more than once

### LINKED ISSUE #2966: Traffic Manager Problems
Hello Carla-Community,

I want to use the TrafficManager on my own map in the OpenDRIVE standalone mode to simulate some traffic. During some first tests with a single vehicle some problems occurred:

1. The vehicle did not change the lane in a curve with 2 existing lanes, both in the driving direction of the vehicle. The lanes are existing and the Client can get access to both of them. I checked that:
```
loc = self.vehicle.get_location()
waypoint = self.map.get_waypoint(loc, True)
left_lane = waypoint.get_left_lane()
right_lane = waypoint.get_right_lane()
```
Either right_lane or left_lane contains a waypoint, depending on the current lane of the vehicle. 

2. I want to change the velocity of the vehicle at specific positions. These informations are not provided in the .xodr-file, where the speed limit is defined (if I understand it right), therefore I want to change the percentage speed difference of the vehicle with the command:
`self.traffic_manager.vehicle_percentage_speed_difference(self.vehicle, self.speed_percentage)`
But this did not affect anything. Is there another way to perfom such a behaviour?

### LINKED ISSUE #3224: Traffic Manager doesn't react on multiple speed changes
Carla: 0.9.9.4
OS: Windows10
Python: 3.7.9

Hello, currently I am working with multiple (around 200) vehicles that are subscribed to one traffic manager with the autopilot on. In my project I need to change multiple times the speed of individual vehicles, but I noticed that sometimes the traffic manager doesn't react to speed changes. For example, in the code below the simulation reacts to the first speed change, but after the timeout the same vehicles won't speed up.

````
tm = client.get_trafficmanager(8000)
for npc in npc_list:
    tm.vehicle_percentage_speed_difference(npc, 90.0)

time.sleep(10)

for npc in npc_list:
    tm.vehicle_percentage_speed_difference(npc, 30.0)
````

Do I need to create batches and apply them with the "apply_batch_sync()" method? If so, how would that work? Or is the issue somewhere else?

jackbart94:
#### Description

Changes in Traffic Manager to fix an issue related to applying specific parameters more than once in the same execution.

Fixes #3224
Fixes #2966

  * **Platform(s):** Ubuntu 18
  * **Python version(s):** 3.6

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3262)


pravinblaze (approved):
Calling std::unordered_map::insert to replace the value for an existing key is erroneous.
Best to check if the key exists and replace the value only for that key.
Thanks for the PR:)

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3262#-:-MGNlsQICPobjWPJGI6E:b-ejla7r)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3264, opened 2020-09-04
Fix a local planner bug when changing its route

glopezdiest:
### Description

The local planner route is divided in two parts. The first part is the *waypoint_buffer*, which englobes the very first waypoints. The rest are part of the *waypoint_queue*.

**Previously**, when resetting the route, the *waypoint_queue* was reset, but not the buffer, causing the new route to be composed of the previous buffer + the new route. **Now**, the buffer is also reset.

As a note, when testing, I **haven't detected any issues with the current agents**, as the buffer is very small, and the planner is smart enough to remove the waypoint that are behind the actor, but this was definitely a problem if someone used the planner with large buffers.

Also, I have additionally changed the local planner of the **Behavior Agent** to account for this error too. However, this agent can add waypoints to the route, so the buffer doesn't always have to be reset, only when the previous route is cleaned.

  * **Python version(s):** 3.6

### Possible Drawbacks

None that I can think of

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3264)


pravinblaze (approved):
Not clearing the buffer was definitely an issue, although it wasn't resulting in problems with a small buffer.
As you mentioned, it would be a problem with larger buffers. Thanks for the fix :)

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3264#-:-MGNg2wJ8TvnUwgaUahM:b-vn65z)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jackbart94)

jackbart94 (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3264)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #3288, opened 2020-09-14
Fixes a map change error when TM is in synchronous mode

### LINKED ISSUE #2789: Traffic Manager hangs after reload
Hello,

In the lastest version (0.9.9), `client.get_trafficmanager()` will hang if the world is reloaded and a traffic manager has been created with sync mode before. Am I missing something? A minimal example to reproduce this is 

```python
import carla

client = carla.Client('127.0.0.1', 2000)
world = client.load_world('Town01')
print ("Loaded world")

tm = client.get_trafficmanager()
tm.set_synchronous_mode(True)
print ("TM set to sync")

world = client.load_world('Town01')
print ("Loaded world")

tm = client.get_trafficmanager() # This will hang
print ("Got TM")
```

I am using Python 3.7 and Ubuntu 18.04.

### LINKED ISSUE #3220: traffic manager synchronous mode prevents apply_settings call after loading a new world
Hi,

I am running code in the form of
```
      client = carla.Client("localhost", port)  # pylint: disable=no-member
      client.set_timeout(20)
      client.load_world(map_name="Town01")
      world = client.get_world()
      world.set_weather(carla.WeatherParameters.ClearNoon)  # pylint: disable=no-member
      world.wait_for_tick()
      frame = world.apply_settings(
          carla.WorldSettings(  # pylint: disable=no-member
              no_rendering_mode=False,
              synchronous_mode=True,
              fixed_delta_seconds=1.0 / fps,
          ))
```

Where after some time I want to load a new world (using the same code).
The problem is that if in between I spawn vehicles, then when running this for the second time, the `apply_settings` call never returns. This only happens when spawning vehicles, not when using an empty map or spawning only pedestrians.

```
     client.load_world(map_name="Town01")
     world = client.get_world()
      frame = world.apply_settings(
          carla.WorldSettings(  # pylint: disable=no-member
              no_rendering_mode=False,
              synchronous_mode=True,
              fixed_delta_seconds=1.0 / 20,
          ))
```
Note that I can successfully retrieve the new map after loading it, as well as do things like applying a new wheather condition etc.. It is only (re-)applying the setttings that fails.

I've tried to first ensure that all actors are destroyed (using the code below), but that didn't make any difference. I've also tried to first do `world.wait_for_tick()` before applying the settings. Again to no effect.
```
self._client.apply_batch([carla.command.DestroyActor(x) for x in [self._hero] + self._vehicles + self._pedestrians])
```

Am I missing a required step to reset the world? (The documentation seems to state that loading a new world automatically takes care of everything).
```
Client API version     = 0.9.9.4-255-gc56a7738
Simulator API version  = ba812e75  (latest docker image)
```

jackbart94:
#### Description

An erroneous update of the TM while in synchronous mode was causing the client to hang on map changes.

Fixes #3220
Fixes #2789

  * **Platform(s):** Ubuntu 18.04, Windows

#### Possible Drawbacks

Seems to not be working for Windows, further testing is required.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3288)


bernatx (approved):
> Reviewed 1 of 1 files at r3, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3288)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @doterop and @pravinblaze)

PR #3297, opened 2020-09-15
Fixed ugly z-fighting issue with OSM maps

Axel1092:
#### Description

This PR fixes issues of z-fighting in OSM maps where short roads with very high curvature can be generated, producing visible issues in standalone mode. The normal vector of the mesh vertices is checked to be pointing upwards and inverted if this condition is not met as road geometry should always point towards the sky.

This also adds a new small actor `AProceduralMeshActor` which allows for better inspection in the editor.

  * **Python version(s):** 3

#### Possible Drawbacks

The geometry of the walls is also checked but as their normals are perpendicular to the vertical direction they shouldn't be affected.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3297)


marcgpuig (approved):
> Reviewed 2 of 3 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3297)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sergi-e)

PR #3365, opened 2020-09-24
Hotfix to cope with boost-python issues on RSS integration

berndgassmann:
boost 1.72.0 doesn't provide correct output on
find_package(Boost COMPONENTS python3x)
in case before another call to a non supported e.g. python27 version was
queried.

Therefore this patch prevents the ad-rss-lib components to query the not
suported python version.

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3365)


berndgassmann:
Since CARLA 0.9.10 is out I will provide a better solution based on the new python interface (see also: https://github.com/carla-simulator/map/pull/23) that get's rid of the RSS import hack... in the next few days.

PR #3394, opened 2020-10-05
Fix dead-lock in synchronous mode at high frame rate

bernatx:
#### Description

Fixed a dead lock in synchronous mode happening randomly at high frame rate. 

It was caused by a race condition between messages to be sent through the network (the message causing the dead-lock was the one who has the episode data). The messages are added in a queue and processed in order. While one is being sending a flag is marked, and then the flag is unmarked once the message has been sent. But the unmarking of the flag was coded in a handler that was added to the queue when the message has been sent, instead of processing it at the instant. That was producing the race condition when another message was added to the queue before this handler, because if the flag is not unmarked the next message will be discarded. 

In the case of a message with the episode data, if it is skipped in synchronous mode then the client will never receive that episode, and will wait forever, while the server will never receive another tick, so both will be in dead-lock.

The fixes just do two things:

* communicate to the stream server in which mode the server is running (synchronous or asynchronous)
* process the handler of a message sent at the instant, instead of adding to the queue. 
* in synchronous mode the server will wait until he can send a message (wait for the flag unmarked)
* in asynchronous mode we will skip sending a message while flag is marked (for slow networks)

Fixes # 2809

  * **Platform(s):** Windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3394)

### DIFF
--- a/LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
+++ b/LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
@@ -79,9 +79,8 @@ namespace tcp {
       if (!_socket.is_open()) {
         return;
       }
-      if (_is_writing) {
-        log_debug("session", _session_id, ": connection too slow: message discarded");
-        return;
+      while (_is_writing) {
+        std::this_thread::yield();
       }
       _is_writing = true;
 
@@ -102,7 +101,7 @@ namespace tcp {
       boost::asio::async_write(
           _socket,
           message->GetBufferSequence(),
-          boost::asio::bind_executor(_strand, handle_sent));
+          handle_sent);
     });
   }
 
@@ -126,7 +125,6 @@ namespace tcp {
   }
 
   void ServerSession::CloseNow() {
-    DEBUG_ASSERT(_strand.running_in_this_thread());
     _deadline.cancel();
     if (_socket.is_open()) {
       _socket.close();
--- a/LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
+++ b/LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
@@ -5,6 +5,7 @@
 // For a copy, see <https://opensource.org/licenses/MIT>.
 
 #include "carla/streaming/detail/tcp/ServerSession.h"
+#include "carla/streaming/detail/tcp/Server.h"
 
 #include "carla/Debug.h"
 #include "carla/Logging.h"
@@ -15,6 +16,7 @@
 #include <boost/asio/post.hpp>
 
 #include <atomic>
+#include <thread>
 
 namespace carla {
 namespace streaming {
@@ -25,9 +27,11 @@ namespace tcp {
 
   ServerSession::ServerSession(
       boost::asio::io_context &io_context,
-      const time_duration timeout)
+      const time_duration timeout,
+      Server &server)
     : LIBCARLA_INITIALIZE_LIFETIME_PROFILER(
           std::string("tcp server session ") + std::to_string(SESSION_COUNTER)),
+      _server(server),
       _session_id(SESSION_COUNTER++),
       _socket(io_context),
       _timeout(timeout),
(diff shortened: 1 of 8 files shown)


doterop (changes requested):
> Reviewed 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3394)* status: all files reviewed, 1 unresolved discussion (waiting on @bernatx and @marcgpuig)

___
*[LibCarla/source/carla/streaming/detail/tcp/Server.h, line 61 at r2](https://reviewable.io/reviews/carla-simulator/carla/3394#-MIyKRkP3Af2OQV82M7K:-MIyKRkQEvw3j75hcDIe:b-resfae) ([raw file](https://github.com/carla-simulator/carla/blob/7e3e40d922110d1b236ae303797dc9f2152f6219/LibCarla/source/carla/streaming/detail/tcp/Server.h#L61)):*
> ```
> void
> ```
Remove this void and add const to the function: `bool IsSynchronousMode(void) const`

bernatx (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3394)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @doterop and @marcgpuig)

___
*[LibCarla/source/carla/streaming/detail/tcp/Server.h, line 61 at r2](https://reviewable.io/reviews/carla-simulator/carla/3394#-MIyKRkP3Af2OQV82M7K:-MIyNxDv5CNX23EszBfT:b-896fix) ([raw file](https://github.com/carla-simulator/carla/blob/7e3e40d922110d1b236ae303797dc9f2152f6219/LibCarla/source/carla/streaming/detail/tcp/Server.h#L61)):*
<details><summary><i>Previously, doterop wrote…</i></summary><blockquote>

> ```
> void
> ```
Remove this void and add const to the function: `bool IsSynchronousMode(void) const`
</blockquote></details>

Done.

doterop (approved):
:lgtm:

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3394#-:-MJ-_yNS6yDJO6m4Vk8V:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marcgpuig)

LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
```diff
@@ -126,7 +138,6 @@ namespace tcp {
   }
 
   void ServerSession::CloseNow() {
-    DEBUG_ASSERT(_strand.running_in_this_thread());
     _deadline.cancel();
     if (_socket.is_open()) {
       _socket.close();
```
nsubiron, line 143: The assert here shouldn't be removed, it is not safe to close the socket concurrently, [needs to be wrapped in the strand](https://stackoverflow.com/a/35411254).

LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
```diff
@@ -102,7 +114,7 @@ namespace tcp {
       boost::asio::async_write(
           _socket,
           message->GetBufferSequence(),
-          boost::asio::bind_executor(_strand, handle_sent));
+          handle_sent);
```
nsubiron, line 117: This is problematic, the `handle_sent` method cannot run concurrently as it is now, you would need to change it, call `Close()` instead of `CloseNow()` and make `_is_writing` atomic.

LibCarla/source/carla/streaming/detail/tcp/ServerSession.cpp
```diff
@@ -80,8 +84,16 @@ namespace tcp {
         return;
       }
       if (_is_writing) {
-        log_debug("session", _session_id, ": connection too slow: message discarded");
-        return;
+        if (_server.IsSynchronousMode()) {
+          // wait until previous message has been sent
+          while (_is_writing) {
+            std::this_thread::yield();
+          }
```
nsubiron, line 91: This could cause a dead-lock if you have as many or more sensors than threads. If you're very unlucky every session may end up spinning on this loop and there won't be any threads available to execute the callback.

nsubiron (commented):
Sorry for commenting quite late, but you introduced a potential crash here.

PR #3477, opened 2020-10-19
Edited Windows.mk for directories with spaces

Computer-CGuy:
Recently added BuildOSM2ODR.bat wasn't quoted, meaning it may not work properly with directories having spaces. Eg.  Space between "My" and "PC" in C:/Users/My PC. This will lead to ignoring the BuildOSM2ODR.bat and ultimately inaccurate build. The issue is fixed in this commit.

#### Description

Fixes #  

  * **Platform(s):** ... Windows
  * **Python version(s):** ... 3.7.6-3.8.0
  * **Unreal Engine version(s):** ... 4.24.3

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3477)


bernatx:
Thank you for the fix

PR #3548, opened 2020-11-05
Update CarlaRecorder.cpp

yankagan:
Fixed bug in Recorder: vehicle bounding-boxes and trigger-volumes of traffic signs were not saved to recorder-file.

  - [V] up-to-date with the `dev` branch 
  - [X] tested with latest changes
  - [V] Code compiles correctly
  - [X] All tests passing with `make check` (only Linux)

Description:
------------
Bounding-boxes and trigger-volumes were not added to the boxes arrays in the Ticking loop of the recorder but only once on start. Due to this, no bounding/trigger boxes were not actually saved to recorder-file. 

Tested on:
   Platform(s): Windows...
   Python version(s): 3.6
   Unreal Engine version(s): 4.24.3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3548)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp
@@ -101,6 +101,7 @@ void ACarlaRecorder::Ticking(float DeltaSeconds)
           if (bAdditionalData)
           {
             AddActorKinematics(View);
+			AddActorBoundingBox(View);
           }
           break;
 
@@ -117,6 +118,8 @@ void ACarlaRecorder::Ticking(float DeltaSeconds)
         // save the state of each traffic light
         case FActorView::ActorType::TrafficLight:
           AddTrafficLightState(View);
+		  ATrafficSignBase* TrafficSign = Cast<ATrafficSignBase>(View.GetActor());	
+		  AddTriggerVolume(*TrafficSign);
           break;
       }
     }


Axel1092:
Hi @yankagan, thank you for contributing to the project. I have a question with this PR, the problem is that trigger boxes and bounding boxes of traffic lights and signs are only recorded once at the beginning of the recording right? This is done by design as traffic lights and signs are static elements in the simulation, therefore, there is no need to repeat information every frame. Is there a special situation where recording this information every frame is necessary?

yankagan:
> Hi @yankagan, thank you for contributing to the project. I have a question with this PR, the problem is that trigger boxes and bounding boxes of traffic lights and signs are only recorded once at the beginning of the recording right? This is done by design as traffic lights and signs are static elements in the simulation, therefore, there is no need to repeat information every frame. Is there a special situation where recording this information every frame is necessary?

Hi Axel,

The problem was that  NO VEHICLE bounding boxes were saved except for the first frame which i think were all zeros - but let me check again. I expected to see BB for every frame.  I missed the part about the traffic signs - you are right of course. Also later today i realized that bounding boxes for vehicles are in **local** coordinates so they too do not change (i was expecting them in *global* coordinates - perhaps it is worth emphasising this point in documentation?).
I  will check the first frame to see if bounding boxes are saved correctly. Sorry for the confusion. 

Yan

yankagan:
OK after checking, I realized the Bounding boxes are written only in the first frame the vehicles are created. In my case this was frame no.10 so I missed that. Closing the PR. Sorry.

PR #3566, opened 2020-11-10
Fixed: update_information method of BehaviorAgent does not work with an instance of carla.libcarla.World class.

fnozarian:
#### Description

The `update_information` method of the `BehaviorAgent` currently takes as input an instance of the `World` class defined in `automatic_control.py`. Therefore, the user can't use this agent independent of the `World` class in the provided example. According to the documentation of the method, the `update_information` requires an instance of `carla.world`, however, `carla.libcarla.World` does not have a player attribute. To fix this issue and use `BehaviorAgent` outside of the Carla `automatic_control.py` example, I simply removed the world parameter from the method signature and instead used the `self.vehicle` attribute. I also changed the `automatic_control.py` accordingly to reflect the new signature of the `update_information` method.

Fixes #  

  * **Platform(s): Ubuntu 20.04
  * **Python version(s): 3.8 

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3566)


jackbart94 (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3566)* status: :shipit: complete! all files reviewed, all discussions resolved

jackbart94:
Thanks for opening the PR! This change totally makes sense, looks good to me.

PR #3662, opened 2020-12-02
Toggle ragdoll physics + Upgrade PhysXVehicle for sweep wheel collisions

Axel1092:
#### Description

This PR combines 2 features:
1)
Added `WorldSettings.deterministic_ragdolls` to enable or disable ragdoll physics and death animations.
For now this only has an effect with content branch `axel/ragdoll_animations`.

2)
We have upgrade the PhysXVehicle plugin of Unreal to allow for sweep collision for wheels.
This results in much more realistic wheel-road interaction, this is important especially in curbs, but have a performance hit. 
This is disabled by default but it can be enable at any moment throught the physics control of the vehicle:

```
physics_control = vehicle.get_physics_control()
physics_control.use_sweep_wheel_collision = True
vehicle.apply_physics_control(physics_control)
```

Fixes #  

  * **Python version(s):** 3

#### Possible Drawbacks

If enabled, it is expected to have a performance hit. If it not, everyhing should works as previously.
This requires the Unreal Engine to be patch (in the PhysXVehicle plugin) so it will require to build the Engine from source in Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3662)


doterop (approved):
> Reviewed 8 of 10 files at r1, 3 of 5 files at r2, 1 of 1 files at r3, 3 of 3 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3662)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #3675, opened 2020-12-05
Update build_windows.md

BAROBA-3d:
This worked for me, I was getting errors with (C)make and redirecting it to Cmake and Nmake fixed it.

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3675)

### DIFF
--- a/Docs/build_windows.md
+++ b/Docs/build_windows.md
@@ -4,7 +4,7 @@
 *   [__Requirements__](#requirements)  
 	*   [System specifics](#system-specifics)  
 *   [__Necessary software__](#necessary-software)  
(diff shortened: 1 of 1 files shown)


BAROBA-3d:
After a succesful build things have changed..
You still need make to be able to "make build" and "make launch".
So make is required.
I think it was a PATH issue, with a wrong configured path to cmake and nmake. You have to make sure that the right paths are choosen to the VS-programs. ( can be done with Where cmake.exe and where nmake.exe) Also make sure where vsvars64.bat is located, also an important one.

PR #3833, opened 2021-01-27
Fixed bug related to RoadOptions

### LINKED ISSUE #3462: Enum comparison in behavior_agent fails even if True
This is regarding the enum comparisons in the behavior_agent here, https://github.com/carla-simulator/carla/blob/31fff28bc6d24ddae0b3977f8ac343ed16019202/PythonAPI/carla/agents/navigation/behavior_agent.py#L271

I am on Python 3.7 and Carla 0.9.9.4, 

The enum comparison doesnt work directly,

for example `self.direction == RoadOption.LANEFOLLOW` on line 286 is always `False`, even if the direction is indeed `LANEFOLLOW`

I have to use the `.value` attribute of the enum to ensure it works correctly,

Here's a screenshot from my debugging console,

image

glopezdiest:
### Description

Fixed a bug at the BehaviourAgent causing the RoadOptions to not work as intended.

This was caused by having two instances of the RoadOption enumerator, one at the *local_planner.py* and another one at the *local_planner_behavior.py*. Despite the fact that the latter file has one, this same file also uses the *global_route_planner.py* to calculate the route of the agent, which in turn uses the RoadOptions at the *local_planner.py*. This causes some unintended equality checks between these two different RoadOption enums, which will always be False.

This has been fixed by removing the RoadOptions enum at the *local_planner_behavior.py*, and always using the one at the *local_planner.py*

Fixes  #3462

  * **Python version(s):** 3.6

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3833)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Latest
+
+  * Fixed bug causing the RoadOptions at the BehaviorAgent to not work as intended
+
+
 ## CARLA 0.9.11
 
   * Improved the documentation for use with pandoc tool by converting html tags to their markdown equivalent
--- a/PythonAPI/carla/agents/navigation/behavior_agent.py
+++ b/PythonAPI/carla/agents/navigation/behavior_agent.py
@@ -12,7 +12,8 @@ import random
 import numpy as np
 import carla
 from agents.navigation.agent import Agent
-from agents.navigation.local_planner_behavior import LocalPlanner, RoadOption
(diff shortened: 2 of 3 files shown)


jackbart94 (approved):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3833)* status: :shipit: complete! all files reviewed, all discussions resolved

Issue #3758, opened 2020-12-28
rsync: change_dir "/home/ue4/carla//./Plugins" failed: No such file or directory (2)


will-sloan:
I'm trying to build the docker image, and I am at the last step. 
`docker build -t carla -f Carla.Dockerfile .`
But it gives the error below. 
I am on Ubuntu 18.04, and the carla version is whichever is used by the dockerfile. 

I included some lines above for context, not sure if that helps. 
```
Sumo/util/create_sumo_vtypes.py
Sumo/util/netconvert_carla.py
Sumo/util/sequential_types.py
Sumo/util/data/
Sumo/util/data/opendrive_netconvert.typ.xml
rsync: change_dir "/home/ue4/carla//./Plugins" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1196) [sender=3.1.2]
make: *** [package] Error 23
Util/BuildTools/Linux.mk:16: recipe for target 'package' failed
The command '/bin/sh -c cd /home/ue4 &&   if [ -z ${GIT_BRANCH+x} ]; then git clone --depth 1 https://github.com/carla-
simulator/carla.git;   else git clone --depth 1 --branch $GIT_BRANCH https://github.com/carla-simulator/carla.git; fi &&   
cd /home/ue4/carla &&   ./Update.sh &&   make CarlaUE4Editor &&   make PythonAPI &&   make build.utils &&   
make package &&   rm -r /home/ue4/carla/Dist' returned a non-zero code: 2
```

will-sloan:
The referenced issue fixed it for me.

thillRobot:
I am having the same issue. @will-sloan, Can you elaborate on the fix?

will-sloan:
@thillRobot My fix was to change line 166 in `Package.sh` in the Docker container to:
`copy_if_changed "./Unreal/CarlaUE4/Plugins/" "${DESTINATION}/Plugins/"`
Here are the steps I followed: 
1 - In `Util/Docker`, create a file called `MyPackage.sh` 
2 - `cd ..`
3 - run `echo BuildTools/Package.sh > Docker/MyPackage.sh`
4 - Change line 166 in `MyPackage.sh` to what I put above.
5 - `cd Docker`
6 - change `Carla.Dockerfile` to: 
```
FROM carla-prerequisites:latest

ARG GIT_BRANCH

USER ue4

RUN cd /home/ue4 && \
[...]

PR #3894, opened 2021-02-17
Fix adding plugins to the package

ARSekkat:
This produces the error in #3758
Several persons confirmed that it worked https://github.com/carla-simulator/carla/commit/7374804542ce1796ddfc022ef91603f526d2cf68#r45498679

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3894)

### DIFF
--- a/Util/BuildTools/Package.sh
+++ b/Util/BuildTools/Package.sh
@@ -163,7 +163,7 @@ if ${DO_CARLA_RELEASE} ; then
 
   copy_if_changed "./Co-Simulation/" "${DESTINATION}/Co-Simulation/"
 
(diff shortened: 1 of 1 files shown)


bernatx:
Thanks, but the problem is already fixed in current **dev** branch.

Your solution is not correct, as the plugins to copy to the package are not related to Unreal Engine but to CARLA simulator. The folder path is **carla/plugins** and not **Unreal/CarlaUE4/Plugins/**. They are not the same.

btw: an easy fix is just to create the plugins folder inside carla. The current solution is to copy that folder only if exists.

PR #3896, opened 2021-02-17
Carsim integration, stop at collisions and Lincoln 2020 dimensions

bernatx:
#### Description

Now the CarSim simulation stops when the vehicle collides with anything. That is the way we found to be realistic until CarSim can handle collisions with external objects.

Also, we added the Lincoln 2020 vehicle dimensions as a .PAR file (for the sprung mass), so the user can take that file and setup his car in Carsim. The rest of the properties about the engine and other parts of the vehicle needs to be setup by the user.

  * **Platform(s):** Windows, Ubuntu
  * **Python version(s):** 3.7

#### Possible Drawbacks

to avoid unreal simulation of physics, we stop Carsim simulation after the first collision of the vehicle with the world. That could change in the future when Carsim handles collisions.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3896)


Axel1092 (approved):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3896)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #3990, opened 2021-03-16
Improved visualization of the OpenDRIVE Standalone mode

fitnessflo:
#### Description

The code changes improve the visualization of the road mesh for a given .xodr file. For one a texture gets applied to the generated road mesh and additionally a separated mesh for the lane markings gets generated. This is necessary since only one texture can be applied to one mesh. Currently only the marking types _"solid"_ and _"broken"_ are supported and for anything else will default to _"solid"_.

The code for the mark mesh follows the same scheme as the generation of the road mesh, i.e. a **ChunkedGeneration** iterating over every Road -> every Section -> every Lane. The differences follow from the fact that _"broken"_ markings are more complex to handle, since one needs to remember if currently a _"line"_ or a _"gap"_ has to be generated and the remaining length of said line/gap as well. A HashMap is used to track this information for a pair of **<RoadId, LaneId>**, I will refer to this as the _"mark_info"_ in the following. This mark_info is particular important if one road ends and another starts. Thus the successor(s) of a road are needed. If the successor is a junction the information about the _actual_ RoadIds cannot be accessed in **MeshFactory.cpp**. Hence an additional HashMap gets generated in **Map.cpp** that maps a RoadId to a vector of RoadIds i.e. the successors of the road. Thus the order in which the roads are handled is important, currently this is done in ascending order of their RoadId. The reason for this decision follows from the fact that in general the road network can be viewed as a _cyclic_ graph and I am not aware of an algorithm that can order such a graph. Hence the assumption "for every road _x_ the Id(s) of the successor road(s) are bigger than x.Id" was made. The code still works in case this assumption is violated, but might generate lines/gaps of varying length around transitions of roads. The overall length of the line/gap of a _broken_ mark are defined in **road_params**.

When **GenerateMark** is called on a lane the **RoadInfoMarkRecord** gets looked up first, to acquire the information about the type of marking (solid, broken) and the width. Then a key with the current RoadId and LaneId is used to check if there is already a mark_info in the HashMap and if not then a new entry is generated. The 3d points of the marking gets calculated similarly to the points of a road. In a loop the corner points of the lane at the current position are calculated and depending on the LaneId one or the other is used as the center for the marking. Using this point and the width of the marking the actual two points are calculated and recorded. Additionally if the remaining length of a line/gap is less then the **road_param.resolution** then the end points of the line/gap are calculated and the mark_info gets updated, i.e. switch line to gap and vice versa and set the new remaining length. Finally depending if it is the end of a section or the end of the road the mark_info gets recorded for the current key or for the successors.

Fixes #  

  * **Python version(s):** 3.8.5

#### Possible Drawbacks

None that I could think of, as only visuals are added.

#### Screenshots

Before:
OpenDriveBefore
After:
OpenDrive-final

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3990)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## CARLA 0.9.11
 
   * Improved the documentation for use with pandoc tool by converting html tags to their markdown equivalent
+  * Improved visuals of OpenDRIVE Standalone mode generated maps
   * Refactored FAQ section of docs to use minimal html and fix broken layout
   * Extended the local planner with a lateral `offset`
   * Upgraded to DirectX 12 on Windows
@@ -17,6 +18,13 @@
     - Added `enable_environment_objects`call to enable/disable objects of the level
     - Added `horizontal_fov` parameter to lidar sensor to allow for restriction of its field of view
     - Added `WorldSettings.deterministic_ragdolls` to enable deterministic or physically based ragdolls
+    - Added `GenerateMarkMesh()` function that generates a mesh for lane markings to the OpenDRIVE map generation
+    - Added textures for the road/marking meshes to the OpenDRIVE map generation
+    - Added new method to `Map`: `GenerateMarkMesh(Parameters)`
+    - Added new methods to `MeshFactory`: `GenerateMarkWithMaxLen(LaneSection)`, `GenerateMark(LaneSection)`, `GenerateMark(Lane)` and `GenerateMark(Lane, s_start, s_end)`
+    - Added new paramters to `road_params`: `mark_height`, `broken_mark_len` and `broken_gap_len`
+    - Added `unordered_map connections` to track the successor RoadIds of every road (to circumvent JunctionIds) to `MeshFactory`
+    - Added `unordered_map mark_lane_memory` to track information about markings of type _broken_ to `MeshFactory`
   * Fixed RSSSensor python3 build and import of open drive maps by updating to ad-rss v4.2.0 and ad-map-access v2.3.0. Python import of dependent 'ad' python modules reflects now the namespaces of the C++ interface and follow doxygen documentation
   * Fixed sensor transformations and sensor data transformations mismatch in IMU and camera-based sensors
   * Fixed random dead-lock on synchronous mode at high frame rate
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -1100,7 +1100,7 @@ namespace road {
       std::vector<road::RoadId> succs;
       if (_data.ContainsRoad(road.GetSuccessor())) {
         succs.push_back(road.GetSuccessor());
-      } else if (road.GetSuccessor() >= 0) { 
+      } else { 
         for (auto tmp : road.GetNexts()) {
           if (tmp->GetJunctionId() != -1 && tmp->GetPredecessor() == road.GetId()) {
             succs.push_back(tmp->GetId());
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -8,6 +8,7 @@
 #include "carla/Exception.h"
 #include "carla/geom/Math.h"
 #include "carla/road/MeshFactory.h"
+#include "carla/road/RoadTypes.h"
 #include "carla/road/element/LaneCrossingCalculator.h"
 #include "carla/road/element/RoadInfoCrosswalk.h"
 #include "carla/road/element/RoadInfoElevation.h"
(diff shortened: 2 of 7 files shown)


germanros1987:
@Axel1092 as main feature reviewer.

Axel1092:
Hi @fitnessflo, sorry for the late response. 
Thank you for this pull request, this looks a very nice addition to the standalone mode. We will start revising and revising it and give you feedback soon.

Axel1092 (changes requested):
> Reviewed 3 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/3990)* status: 3 of 7 files reviewed, 9 unresolved discussions (waiting on @fitnessflo)

___
*[LibCarla/source/carla/road/Map.cpp, line 1097 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYG6q762dwDAh0UZXpt:-MYG6q762dwDAh0UZXpu:b-rztukd) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/Map.cpp#L1097)):*
> ```cpp
>     keys.reserve(_data.GetRoadCount());
> 
>     for (auto &&pair : _data.GetRoads()) {
> ```

Why is this loop necessary? You don't need any information about successors to put lane marking across the road.
___
*[LibCarla/source/carla/road/Map.cpp, line 1116 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYG7jl47Bc8jlNYpbIa:-MYG7jl47Bc8jlNYpbIb:b-9sqg9g) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/Map.cpp#L1116)):*
> ```cpp
>     std::sort(keys.begin(), keys.end());
> 
>     for (auto i : keys) {
> ```

You should just loop  `_data.GetRoads()` here.
___
*[LibCarla/source/carla/road/MeshFactory.h, line 126 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYG7uoy7U1rYKcOZTOC:-MYG7uoz60qAxhf1Xx5d:b-tlua84) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.h#L126)):*
> ```objc
>     RoadParameters road_param;
> 
>     std::unordered_map<road::RoadId, std::vector<road::RoadId>> connections;
> ```

Why is it necessary to store connection information=
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 37 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYG8DUxEUHsQPaOhCk3:-MYG8DUxEUHsQPaOhCk4:bqx2jwk) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L37)):*
> ```cpp
>   // Holds information for a successor/predecessor lane if a mark/gap went past a section/road
>   typedef std::pair<road::RoadId, road::LaneId> pair;
>   static std::unordered_map<pair, std::pair<bool,double>, boost::hash<pair>> mark_lane_memory;
> ```

Why do you need to store this information what does std::pair<bool,double> represent? Also, this should be a class member, not a static variable if it is really needed. Please, explain what is the role of this variable, why do we need to store information of the lanes that we have already generated markings.
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 38 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYG9NQi6pXRAnwpfQ0X:-MYG9NQi6pXRAnwpfQ0Y:b5imnm5) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L38)):*
> ```cpp
>   typedef std::pair<road::RoadId, road::LaneId> pair;
>   static std::unordered_map<pair, std::pair<bool,double>, boost::hash<pair>> mark_lane_memory;
>   static std::unordered_map<road::RoadId, std::vector<road::RoadId>> connections;
> ```

Static variable. Hides member in the header file.
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 259 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYGAJ0B1vHVxBq77_3Y:-MYGAJ0B1vHVxBq77_3Z:b-6ts464) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L259)):*
> ```cpp
>     // The lane with lane_id 0 have no physical representation in OpenDRIVE
>     Mesh out_mesh;
>     if (lane.GetId() == 0) {
> ```

Lane marking records are present in lane 0 so it should not be skipped.
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 265 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYGD5ob4NZAYK3h0IjX:-MYGD5ob4NZAYK3h0IjY:bp1xchp) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L265)):*
> ```cpp
> 
>     // get the marking information, type, height and lateral offset
>     const auto mark_info = lane.GetInfo<road::element::RoadInfoMarkRecord>(s_current);
> ```

There can be a lot of different mark info sections within the lane. Use `lane.GetInfos<road::element::RoadInfoMarkRecord>()` to obtain a list. You would need to iterate over that list to insert different marking types on different s values
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 293 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYGBJUHDYlrUHq8n4hY:-MYGBJUHDYlrUHq8n4hZ:b-v8q9wr) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L293)):*
> ```cpp
>     }
> 
>     auto mark_memory = mark_lane_memory.at(key);
> ```

Here I think you could just do:  `mark_memory = std::make_pair(true, road_param.broken_mark_len)` or `std::make_pair(true, s_end - s_start)` as `mark_lane_memory`
___
*[LibCarla/source/carla/road/MeshFactory.cpp, line 367 at r2](https://reviewable.io/reviews/carla-simulator/carla/3990#-MYGHEyo9FmSNCkkErwj:-MYGHEyo9FmSNCkkErwk:b8p8v2c) ([raw file](https://github.com/carla-simulator/carla/blob/06cbe29f2e439a810ec509933720b2f352d1ada3/LibCarla/source/carla/road/MeshFactory.cpp#L367)):*
> ```cpp
>         auto nxt_key = std::make_pair(nxt_road_id, nxt_lane_id);
>         mark_lane_memory.erase(nxt_key);
>         mark_lane_memory.insert(std::make_pair(nxt_key, mark_memory));
> ```

Why store the mark of the current road in the successor road?

Axel1092:
Hi @fitnessflo, I left a few comments on specific lines but overall I don't really understand why you keep track of successors and the `mark_lane_memory` variable. It seems to me that for a lane part of s1 to s2 you only need the current information to paint a solid line or a broken one. 
In addition, the algorithm seems to be a bit broken:
image
In this example I used Town10HD (CarlaUE4/Content/Carla/Maps/OpenDrive/Town10HD.xodr) and only a few marking are present while most are missing.

CaffreyXu:
Hi @Axel1092 , I also want to draw lane mark in this mode. Do you have any other advises? I have already tried to modify the source code refering this website, but still struggling with some bug in Make launch. Thank you for your time!

CaffreyXu:
Hello @fitnessflo. I wish to draw lane mark in this mode. Could you give me some advice? I study your code for some time, which inspired me. But I am still having some trouble on building it. Did you have other update recently? Thank you!

CaffreyXu:
@fitnessflo I am sorry  for the late reply, and i sincerely appreciate your reply! Thank you!  Now, I can rebuild your work, some changes may take into consideration, since the version of Carla is 0.9.13 now. If you want to know the change, i can update it for you. Here is my email: 332603717@qq.com. I am a undergraduate who will go to Singapore for EE master from China.

PR #4041, opened 2021-03-30
Enable the no_delay option on sockets

bernatx:
#### Description

The **no_delay** socket option was only enabled on stream sockets on the server side, but not on the client side (client send the stream id at the beginning through the steam) and also not in any of the RPC sockets, causing a discontinuous delay on the tick() packet.

We have changed the RPCLib fork to enable this option on both client and server RPC. 
Now we are using the RPCLib tag v2.2.1_c5

  * **Platform(s):** Ubuntu
  * **Python version(s):** 3.7
  * **Unreal Engine version(s):** UE 4.26

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4041)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
   * Fixed bug causing the RoadOptions at the BehaviorAgent to not work as intended
   * Upgrading to Unreal Engine 4.26
   * Added Lincoln 2020 vehicle dimensions for CarSim integration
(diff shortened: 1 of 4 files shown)


Axel1092 (approved):
> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4041)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #4121, opened 2021-04-21
Improved Traffic light API

Axel1092:
#### Description
This PR adds new functions to improve gathering information about the traffic lights.

New `TrafficLight` actor functions:
- `get_affected_lane_waypoints()` returns a list of waypoints at the start of the lanes affected by the traffic light (inside the junction).
- `get_stop_waypoints()` returns a list of waypoints where the vehicles should stop (computed using the traffic light trigger volume).
- `get_light_boxes()` returns a list of oriented bounding boxes corresponding to the light boxes of the traffic light.
- `get_opendrive_id()` returns the OpenDRIVE id of the traffic light

New `World` functions:
- `get_traffic_lights_from_waypoint(waypoint, distance)` returns a list of traffic lights affecting the path between the `waypoint` and `distance` ahead following the road.
- `get_traffic_lights_in_junction(junction_id)` returns a list of traffic lights affecting a specific `junction_id`.

  * **Python version(s):** 3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4121)


glopezdiest:
Hey, a "couple" comments about the functions themselves, not their implementation.
- `get_traffic_lights_from_waypoint(waypoint, distance)`. This function will be useful as a means to do traffic light search on agent runtime. Maybe it would be a good idea to add an additionally parameter to make it optional to stop at the next junction, as that will be the next challenge of the agent, pretty much ignoring whatever comes next (in terms of traffic lights).

- `get_effect_waypoints()`: this one is completely different than what I thought it did. From the PR comment, it *returns a list of waypoints where the traffic light is having an effect (where the vehicles should stop)* and it seems to be getting the first points inside the junction, one for each direction inside it. If that's the intended behavior, that feels contradictory to me, as the bounding boxes are already the "effect zones" of the traffic light (check the image below, where *cyan* is the boudning box waypoint, *pink* is the lane's last waypoint; and *black* the ones gotten using this function)

image

- Regarding `get_traffic_lights_in_junction(junction_id)`, why is this part of the `carla.World`? Why not make it part of the `carla.Junction` class? Additionally, this function is quite similar to `carla.TrafficLight.get_group_traffic_lights()` because aren't all traffic lights in the same junction part of the same group? Regardless of its place, this one is definitely a useful function, and, as a matter of fact, we are already manually calculating at ScenarioRunner, so having them as built-in functions will indeed prove its value.

- Following the previous point, the reverse action could also be useful, i.e. getting the junction from a traffic light (something like `get_junction(traffic_light)`, which can return None is it outside a junction.

- Lastly, as we already have `carla.World.reset_all_traffic_lights()`, can the same be done for only one junction?

bernatx (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4121)* status: 0 of 19 files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #4207, opened 2021-05-14
Large Maps

Axel1092:
#### Description
This PR adds the first iteration of the large map feature. It goes with a related content branch named `doterop/large_map` as it requires the LargeMap umap template to work.

This feature enables loading maps of virtually infinite size. The large map assumes a map divided in small portions (tiles) of a maximum side size of 2km. The tiles need to be square. 

The LargeMap streams the tiles in and out of the server, limiting the elements that participate in the simulation at any given time. Only the regions around the hero/ego vehicle is loaded at any given time including static geometry and dynamic actors. 

This introduces the concept of `dormant actors`, which are actors that are not simulating physics or being rendered. When the hero/ego vehicle moves tiles and actors that become too far away are converted to dormant and woken up once the distance is reduced.

Dormant actors are present as normal actors in the client API and receive commands but are processed once they wake up. The `actor.is_dormant` flag has been added to the python API to allow the client know when an actor is dormant or not.

In the Unreal side, dormant AActor are destroyed and recreated on demand while the registry of actors keeps track of them.

In the LargeMap there is an actor called LargeMapManager that handles the tile information:
image

It is necessary to edit the following fields:
`Large Map Tile Path`: the path to the folder with the tiles.
`Large Map Name`: the name of the map that needs to match the OpenDRIVE.
`Tile 0Offset`: The position of the (0,0) tile.
`Tile Side`: the dimensions of the tile in cm.

  * **Python version(s):** 3

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4207)


bernatx (approved):
> Reviewed 15 of 34 files at r1, 22 of 31 files at r2, 1 of 3 files at r3, 1 of 1 files at r4, 2 of 4 files at r5, 10 of 10 files at r6.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4207)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Axel1092)

PR #4309, opened 2021-06-14
Lane markings color fix

algababr:
#### Description

Fixed bug where color markings would not match with other parts of the road

  * **Platform(s):** Windows
  * **Python version(s):** 3.7.9

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4309)


bernatx (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4309)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @algababr)

___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp, line 219 at r1](https://reviewable.io/reviews/carla-simulator/carla/4309#-Mc8YZIlD6NABhtalScJ:-Mc8YZIlD6NABhtalScK:bgz2sn) ([raw file](https://github.com/carla-simulator/carla/blob/7b790793ea4c563899d24855b9e8e6217e92cbdd/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp#L219)):*
> ```
> GetNumMaterials() == 1
> ```
I think we don not to separate cases by number of materials. We need to replace all materials we find in the mesh, in function if we find the **yellow** keyword or not.
I think your version can fail for assets with two materials that are saved in different order.

algababr:
___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp, line 219 at r1](https://reviewable.io/reviews/carla-simulator/carla/4309#-Mc8YZIlD6NABhtalScJ:-Mc8bgG01vZBvpzR_bXO:bagjzwn) ([raw file](https://github.com/carla-simulator/carla/blob/7b790793ea4c563899d24855b9e8e6217e92cbdd/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp#L219)):*

> ```
> GetNumMaterials() == 1
> ```
I think we don not to separate cases by number of materials. We need to replace all materials we find in the mesh, in function if we find the **yellow** keyword or not.
I think your version can fail for assets with two materials that are saved in different order.
</blockquote></details>

I've changed it, but I wanted to make sure that if it only has 1 material, look for the yellow keyword just in case the mesh has multiple materials, but only 1 is used. But I suppose when importing from roadrunner, only imports the material that is going to be used.

bernatx (dismissed):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4309)* status: all files reviewed, 1 unresolved discussion (waiting on @algababr)

___
*[Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp, line 219 at r1](https://reviewable.io/reviews/carla-simulator/carla/4309#-Mc8YZIlD6NABhtalScJ:-Mc8eERF-lpymv8QK5PX:bah10y) ([raw file](https://github.com/carla-simulator/carla/blob/7b790793ea4c563899d24855b9e8e6217e92cbdd/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Commandlet/PrepareAssetsForCookingCommandlet.cpp#L219)):*
<details><summary><i>Previously, algababr (Roel Algaba Brizuela) wrote…</i></summary><blockquote>

I've changed it, but I wanted to make sure that if it only has 1 material, look for the yellow keyword just in case the mesh has multiple materials, but only 1 is used. But I suppose when importing from roadrunner, only imports the material that is going to be used.
</blockquote></details>

The point is that you made the check for the keyword only if it has only one material, and if it has more then you assign the materials in a hardcoded way (same order always), what seems not be the case wirh RoadRunner (it uses random order).

bernatx (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4309)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @algababr and @bernatx)

bernatx (approved):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4309)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @algababr)

PR #4356, opened 2021-07-01
Required files from server to client cache

AlejandroFraga:
#### Description

Required map files (OpenDRIVE, Traffic Manager...) are now downloaded from the server to the client local cache.

This local cache is stored by default in a folder called carlaCache under the User (Windows) or Home (Linux) folder.
It can be changed with the `set_files_base_folder` command.

To get the names of the required files for the current map use the `get_required_files` command.
Optionally, you can specify a folder under the current loaded ".umap" location to look for the files (e.g. "OpenDrive" to just get the files inside that folder), and you can either store the files or just get the names.

To download a exact file for the current map use the `request_file` command, passing its path and name starting from the current loaded ".umap" location (e.g. "OpenDrive/Map_1.xodr").

  * **Platform(s):** Ubuntu 18.04.5
  * **Python version(s):** 3.6.9
  * **Unreal Engine version(s):** 4.26.2

#### Possible Drawbacks

Possible automatic data transfer from server to client.

In this version, it's checked if the file is already present on the client cache folder to avoid downloading it again, the content is not checked. If a file is updated, the system won't detect that. The file will have to be removed and downloaded again.

If a map is loaded in the server by Unreal's content explorer, the `get_required_files` command will not work correctly. The proper way to change the map is with the `load_world` command.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4356)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   * Added 'check_lidar_bb' util script
   * Added optional flag to `client.replay_file()` `replay_sensors` to enable or disable the replaying the sensors
   * Improved manual_control: now cameras are set in relation with car size
+  * Client required files are sent from the server to a local cache (OpenDRIVE, Traffic Manager...)
   * Added CHRONO library for vehicle dynamics simulation (https://projectchrono.org/)
     - Supported JSON vehicle definition
     - Unsupported collision dynamics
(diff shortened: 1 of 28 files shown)


bernatx (approved):
> Reviewed 14 of 23 files at r1, 1 of 11 files at r3, 1 of 14 files at r4, 2 of 3 files at r5, 9 of 11 files at r6, 2 of 2 files at r7, 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4356)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @AlejandroFraga)

PR #4549, opened 2021-08-17
Fixes nullptr bug in CarlaRecorder::AddCollision 

amparore:
#### Description

Bugfix for a segmentation fault that happens in the server when a vehicle collides with an environment object when the recording is in progress.

How to reproduce the bug:
 1. start the manual_control
 2. start recording (ctrl+R)
 3. make a collision with an environment object (a flower, a fountain, ...). This should cause the server to trigger a segmentation fault in ACarlaRecorder::AddCollision.

The segmentation fault is triggered in ACarlaRecorder::AddCollision since one of the colliding objects is not a registered Carla actor, but no check is done on the pointer returned by the Episode->GetActorRegistry().FindCarlaActor() method.

The submitted patch does these things:
1. In ACarlaRecorder::AddCollision an additional check is performed to ensure that each colliding actor is a registered Carla actor, otherwise a non-valid id (-1) is set in the Collision.DatabaseId* field written in the recording file.
2. when the CarlaRecorderQuery::QueryCollisions reads back the recording file, the code checks if the Collision.DatabaseId* are valid Carla actor ids, otherwise sets the actor type to 'o' ('other' actor).

Checklist:

  - [x] Your branch is up-to-date with the `dev` branch and tested with latest changes
  - [x] Code compiles correctly
  - [ ] All tests passing with `make check` (only Linux) -> All 48 tests passes.
  Currently, the dev branch does not pass all the tests with make change on my system, but this is not related to this pull request. I have tested the submitted code with my scripts, to ensure that the bug is fixed, and the client.show_recorder_collisions() works correctly.

  * **Platform(s):** Linux Ubuntu 20.04
  * **Python version(s):** 3.7

#### Possible Drawbacks

The code is a bugfix, preventing the server to crash. 
However a different solution could be considered: instead of writing the DatabaseId* in the .rec file (which assumes that only registered Carla actors may collide, which is not true), the server could write the blueprint id, and search if the id correspond to a carla actor in the CarlaRecorderQuery::QueryCollisions method. This however would break the backward compatibility with the .rec file format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Fixed bug causing the server to sigsegv when a vehicle collides an environment object in recording mode.
   * Added API functions to 3D vectors: `squared_length`, `length`, `make_unit_vector`, `dot`, `dot_2d`, `distance`, `distance_2d`, `distance_squared`, `distance_squared_2d`, `get_vector_angle`
   * Added API functions to 2D vectors: `squared_length`, `length`, `make_unit_vector`
 
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Recorder/CarlaRecorder.cpp
@@ -469,22 +469,34 @@ void ACarlaRecorder::AddCollision(AActor *Actor1, AActor *Actor2)
     Collision.IsActor2Hero = false;
 
     // check actor 1
-    if (Episode->GetActorRegistry().FindCarlaActor(Actor1)->GetActorInfo() != nullptr)
-    {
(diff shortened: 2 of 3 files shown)


Axel1092:
Hi @amparore, thank you for your contribution and for detecting this issue and sorry for the late reply to this pull request. I doing the review, our only complain is just the naming convention (variables should start with capital letters according to Unreal convention). Also, the branch has some conflicts with dev, if these two things are fixed we would merge this PR asap.

amparore:
I have rebased the patch, it should now merge with the most recent dev branch in the upstream. Variable names should be compliant. It compiles and runs on my pc.

Axel1092:
Hi @amparore, wait be careful something went wrong when you did the rebase. You have introduced a number of changes in the PR from previous commits which will pollute the repo (even more) and probably cause errors. We should see only your changes and not older commits here. Could you please fix this issue? If you have trouble fixing this, I can help you but I would need you to revert the state of the branch to before the rebase.

PR #4676, opened 2021-09-24
Fix global route planner python agent

berndgassmann:
Ensure waypoint.right_lane_marking/left_lane_marking exist before
accessing

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4676)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
   * Added API functions to 3D vectors: `squared_length`, `length`, `make_unit_vector`, `dot`, `dot_2d`, `distance`, `distance_2d`, `distance_squared`, `distance_squared_2d`, `get_vector_angle`
   * Added API functions to 2D vectors: `squared_length`, `length`, `make_unit_vector`
   * Added missing dependency `libomp5` to Release.Dockerfile
+  * Fixed global route planner crash when being used at maps without lane markings
 
 ## CARLA 0.9.12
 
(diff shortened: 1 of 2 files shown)


glopezdiest:
Hey @berndgassmann, as far as I know all waypoints have the lane marking, and while its type or lane change might be None, this shouldn't result in any errors. Could you provide a use-case where your changes are indeed needed?

berndgassmann:
We are currently using an external xodr file of a public research project. It seems that for that one, the auto-generated waypoints are missing those. Therefore, we were not able to run a scenario on this map without these changes. Maybe the root cause lies at a different point where the waypoints are generated from the opendrive or the lane markings in the open drive are not there. But most probably other xodr files might have the same issue. Therefore, fixing this here in the python code should be done in each case. I am also happy to support in further analysis if you like to find out more on why the lane markings are not there in our case.

glopezdiest:
Okay. In that case, I'll dig a bit about how lane marking are created, maybe its an xodr parsing problem. In any case, I don't see any issues with the PR, and it can be merged, but if I prefer to wait in case we find something elsewhere. I'll report back

glopezdiest (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4676)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @joel-mb)

glopezdiest:
PR accepted (and I added the CHANGELOG). Apparently, the lane markings work the same way as getting a right / left lane, where it can return None. I actually didn't know that and it is indeed an error. Thanks for the fix :slightly_smiling_face:

PR #4737, opened 2021-10-11
New python examples for spawn point extraction and OSM2XODR conversion

hitabm:
Created two new and simple examples for getting Spawn Points of map and converting OSM files into XODR files.

  * **Platform(s):** Windows 10
  * **CARLA version:** 0.9.12
  * **Python version(s):** 3.7.9

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4737)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   * Added API functions to 2D vectors: `squared_length`, `length`, `make_unit_vector`
   * Added missing dependency `libomp5` to Release.Dockerfile
   * Fixed global route planner crash when being used at maps without lane markings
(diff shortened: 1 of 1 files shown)


glopezdiest:
Great, everything looks good now! Thank you very much for all the effort, we'll merge it whenever we can :slightly_smiling_face:

RehabHsn:
how to import the opendrive file generated to CARLA? 
I build a map with this opendrive but when i try to start the generated map , it crashed

barasm-hita:
> how to import the opendrive file generated to CARLA? I build a map with this opendrive but when i try to start the generated map , it crashed

First make sure that your OSM file has a road or highway. Then after converting it to XODR using example you can load it into your instance of carla server using script named **_config.py_**. Even you can load OSM file directly without any conversion. Take a look at [here ](https://carla.readthedocs.io/en/0.9.13/tuto_G_openstreetmap/#ingest-into-carla)to see how you can load your map into carla simulation server.

PR #4786, opened 2021-10-25
New Traffic Manager features for 0.9.13

jackbart94:
#### Description

New Traffic Manager features for 0.9.13. This PR includes:

1. **The option for users to use ``traffic_manager.set_path(actor, path)`` to pass a list of carla.Location elements to a vehicle controlled by the Traffic Manager, which will follow the given path.** This has some limitations, given by the amount and the location of points in the path. Users should be aware that, in order for it to work, the locations should not be inside intersections (because routes sometimes overlap) and should give at least one location between two different intersections.

2. **Introduction of a RoadOption element in each SimpleWaypoint, which can be: Void, Left, Right, Straight, LaneFollow, RoadEnd.** Currently developing a way to have more options, so that if an intersection has more than one exits that go left, they can be identified separately.

3. **Users can now get the next action that a vehicle will perform (``traffic_manager.get_next_action(actor)``), or the whole action buffer that the vehicle has already planned. (``traffic_manager.get_all_actions(actor)``).** The first one will return a list with two elements, the first one being the next RoadOption, the second being the waypoint where it's going to be performed. If there's no particular action other than following the lane, the action returned will be LaneFollow and the waypoint will be the furthest in the buffer. The second one will return a list of lists with two elements, which will be the same as before, with each RoadOption that will be performed and the waypoint where said RoadOption is going to begin. If it's only lane following, the same principle as get_next_action is applied.

3. **The option for users to use ``traffic_manager.set_route(actor, route)`` to pass a list of RoadOptions (in the form of strings) to a vehicle controlled by the Traffic Manager, which will follow the given high-level route.** This has currently some limitations, given by the fact that there might be more than one exit with the same RoadOption name, and also doesn't work with lane changes.

4. **Fixes ``traffic_manager.keep_right_rule_percentage(actor, percentage) parameter`` and traffic_manager.set_distance_leading_vehicle(actor, percentage)**

5. **Introduces ``traffic_manager.random_left_lanechange_percentage(actor, percentage)`` and ``traffic_manager.random_right_lanechange_percentage(actor, percentage)`` , which allow the user to set a random chance for a vehicle to perform a lane change at any given frame.** Vehicles already have a policy for lane changes, this parameter will just try to trigger a forced lane changes, and only while the vehicle is moving with some velocity.

6. **Improved braking of vehicles, according to new physics.**

7. **Changes in new VehicleLightStage.**

  * **Platform(s):** Ubuntu 18

#### Possible Drawbacks

Requires quite some testing to avoid possible edge cases.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4786)


joel-mb (dismissed):
> Reviewed 1 of 19 files at r1, 1 of 17 files at r3, 2 of 13 files at r6, 1 of 7 files at r8, 7 of 13 files at r9, 1 of 5 files at r10, 1 of 4 files at r11, 4 of 5 files at r12.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4786)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jackbart94)

joel-mb (approved):
> Reviewed 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4786)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jackbart94)

PR #4817, opened 2021-10-29
Fixed enum to wrong string cause by misuse GetNameStringByValue and GetNameStringByIndex

julee:
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4817)


julee:
> Hi Julee, Thanks for the revision, but can you explain what you are solving and what was the problem? I think the way you are modifying this index by value is not right. Can you explain more? The enums are assigned a value from 0 to N, and that is the index they represent in the enum. Why using an arbitrary value? Also, in that case, we need to cast the value to 64 bits, because they are not 32bits value like the index is.
> 
> Regards

No, you can't assume the enum index is the enum value, because enum value can assign to any value!

Check the QualityLevelUE.h: 

```cpp

UENUM(BlueprintType)
enum class EQualityLevel : uint8
{
  Null = 0, // Workarround for UE4.24 issue with enums
  Low    = CARLA_ENUM_FROM_RPC(Low)    UMETA(DisplayName = "Low"),
  // Medium = CARLA_ENUM_FROM_RPC(Medium) UMETA(DisplayName = "Medium"),
  // High   = CARLA_ENUM_FROM_RPC(High)   UMETA(DisplayName = "High"),
  Epic   = CARLA_ENUM_FROM_RPC(Epic)   UMETA(DisplayName = "Epic"),

  SIZE      UMETA(Hidden),
  INVALID   UMETA(Hidden)
};
```

We can see enum index is not the with enum value ( EQualityLevel::Low index is 1, but the enum value is 0).
And we can't assume assigning enum value is a special case, it's very common.

Unreal Engine documentation also [pointed out the issue](https://docs.unrealengine.com/4.27/en-US/API/Runtime/CoreUObject/UObject/UEnum/GetIndexByValue/):

> Index is the internal index into the Enum array, and is not useful outside of the Enum system Value is the value set in the Enum Class in C++ or Blueprint Enums can be sparse, which means that not every valid Index is a proper Value, and they are not necessarily equal It is not safe to cast an Index to a Enum Class, always do that with a Value instead Gets the internal index for an enum value.


You say:
> Also, in that case, we need to cast the value to 64 bits

This is just according whether the enum is defined to 64 bits int or not, if not, cast to int32_t doesn't matter. Unreal give the parameter with int64_t is because the enum can be defined to 64 bits int, it must support all enum definitions.

julee:
This bug should be fixed

PR #4857, opened 2021-11-09
Bernat/cache version

bernatx:
#### Description

Add an extra folder with the CARLA version (like: 0.9.12 or 0.9.11-654-g62630cef0-dirty). This is needed to avoid problems with different binary formats that could arise with different versions of CARLA.

carlaCache\\**0.9.11-654-g62630cef0-dirty**\Carla\Maps\OpenDrive\

  * **Platform(s):** Windows 10
  * **Python version(s):** 3.7

#### Possible Drawbacks

Currently the cache never is cleaned. That means that with each different version of CARLA, the cache will grow, but currently the files are so small that is not a concern. Probably in the future we add more features to the cache, like an auto-cleaning function.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4857)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
   * Added a RoadOption element in each SimpleWaypoint to specify which action will the vehicle perform if it follows that route.
   * Added the option for users to set a route using RoadOption elements to a vehicle controlled by the Traffic Manager.
   * Fixed keep_right_rule parameter.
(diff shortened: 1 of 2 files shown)


Axel1092 (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4857)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #4858, opened 2021-11-09
Fixed an error caused during map name being called

Anshu-man567:
#### Description
It fixes the errors arising due to `map_name` being called which is not an attribute of `world`

Fixes #  

  * **Python version(s):** Python 3.6.9

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4858)


bernatx:
Thanks for the fix.
We use this line instead, to avoid long names with folders:
`'Map:     % 20s' % world.map.name.split('/')[-1],`
Do you mean to use this?

Anshu-man567:
Actually yeah!

bernatx (approved):

bernatx:
thanks

PR #4989, opened 2021-12-16
TM determinism fixes

glopezdiest:
### Description

This PR fixes some non-deterministic behaviors that were introduced after 0.9.11. These includes:
- Sorting the vehicles at the *ApplyBatchSync* so that the TM always registers them in the same order, avoiding the need to sort them 
- Changed one variable from *unordered map* to *map*, avoiding having to sort it when looping through its keys.
- TM now uses one global seed for all vehicles, instead of one per vehicle.
- Changing a vehicle's physics is now a synchronous call. Due to its high interaction with other commands, as well as its expensive computations, the async nature of the call was causing other commands called after this one, to sometimes finish earlier, breaking determinism and creating unexpected behaviors 
- Updated the determinism smoke test to also use the hybrid mode

  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** CARLA fork

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4989)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Latest
+
+  * Fixed bug causing the Traffic Manager to not be deterministic when using hybrid mode
+
 ## CARLA 0.9.13
 
   * Added new **instance aware semantic segmentation** sensor `sensor.camera.instance_segmentation`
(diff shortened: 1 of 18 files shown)


jackbart94 (approved):

jackbart94 (approved):
> Reviewed 10 of 13 files at r1, 8 of 8 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/4989)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

Issue #3765, opened 2020-12-30
Access to wheels rotation angle in PythonAPI


unumkam:
This feature might be helpful for precise position estimation. Coupling position sensors with wheel odometry seems to be more intuitive way to build feedback-based control of car. I was wondered Carla not provides this data.
Locally I have done it through ACarlaWheeledVehicle, then filled ActorDynamicState from WorldObserver and get it accessible from Python Vehicle type

corkyw10:
@DSantosO could you follow up on this please?

DSantosO:
Hello @unumkam,

This is currently being implemented in PR #3868 and it will be soon in dev branch so I close the issue. Feel free to reopen if necessary.

NishanthARao:
@DSantosO Hi.

From the PRs, I see that two new functions are added: SetWheelSteerAngle and GetWheelSteerAngle.

However, the question asks the rotation angle of the wheel and not the steer angle of the wheel right? Could we use the GetRotationAngle() that UVehcileWheel provides?

PR #5295, opened 2022-03-28
Set wheel orientation

anrp-tri:
Allow the wheels of a vehicle to be positioned fully instead of just the steering angle.

Also properly resolves https://github.com/carla-simulator/carla/issues/3765

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5295)


AlexanderRex:
Hello, thank you for your work this is very important module for self-driving car development! But I have some troubles with merge your branch

AlexanderRex:
after merging and running manual_control.py i have this problem:
Fatal Python error: (pygame parachute) Segmentation Fault
Python runtime state: initialized

Thread 0x00007fe9305a6740 (most recent call first):
  File "manual_control.py", line 1244 in game_loop
  File "manual_control.py", line 1370 in main
  File "manual_control.py", line 1378 in <module>
Aborted (core dumped)

anrp-tri:
FYI I tried to build this locally with on tip editor and carla/dev but the editor exited after clicking play, then I tried applying the vulkan patch in s3 but it does not cleanly apply so I got stuck.

PR #5378, opened 2022-04-21
Fix topology with ending lanes

glopezdiest:
### Description

When using the `Map.get_topology`, there is an error that causes lanes with no successors to not be added. This resulted in something like the image below, where the red lines are the lanes part of the topology. As it can be seen, the topology is missing the lane on the left.

image

This is caused because the topology pairs the start of one lane with the start of the next ones. As such, if no successors are found, the lane's starting point isn't linked with anything. This has been changed and now, if no successors are found, the start of the lane is paired with the its end.

  * **Python version(s):** 3.6

#### Possible Drawbacks

This changes the topology a bit, so there might be some cases where some logic is affected. I don't think this happens anywhere in CARLA though.

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5378)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Fix a bug at `Map.get_topology()`, causing lanes with no successors to not be part of it.
   * Added new ConstantVelocityAgent
   * Added new parameter to the TrafficManager, `set_desired_speed`, to set a vehicle's speed.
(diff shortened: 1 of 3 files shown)


Axel1092 (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5378)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #5402, opened 2022-05-02
Extended Python agents initializations

glopezdiest:
### Description

The Python API agents now have two additional optional inputs. *map_inst* is meant to be a **carla.Map** instance while *grp_inst* is an instance of the **GlobalRoutePlanner** class. 

Both of these classes are, by default, computed by the BasicAgent on initialization, but they are a **very expensive call for large maps** (up to several seconds). As these classes only depend on the static topology and, as such, won't be changed during the simulation, these new inputs can speed up the initialization of the agents

  * **Python version(s):** 3.6

### Possible Drawbacks

None? Nothing is removed and by default it still works the same as before

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5402)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Python agents now accept a carla.Map and GlobalRoutePlanner instances as inputs, avoiding the need to recompute them.
   * Fix a bug at `Map.get_topology()`, causing lanes with no successors to not be part of it.
   * Added new ConstantVelocityAgent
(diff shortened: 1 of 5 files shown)


joel-mb (approved):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5402)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #5407, opened 2022-05-04
Fix TM blocking the simulation

glopezdiest:
### Description

Fixed a bug that caused the TM to sometimes calculate a vehicle's speed as a ridiculously high number.

The TM calculates the speed of physicsless vehicles as the position difference between ticks. However, this fails if another client teleports the vehicle, creating an insanely high spee , and generally causing the LocalizationStage to try to create a larger waypoint buffer than the map, blocking the simulation.

As such, a new `hybrid_end_location` has been added, which is the location that the TM calculates for that vehicle at the end of its execution, and it's that position, and not the one gottetn by the server, the one that's used to calcualte the speed. For all other purposes, the "normal" position is used.  

  * **Python version(s):** 3.6

### Possible Drawbacks

Some corner cases might have been changed, such as when the TM doesn't have the actor's information saved

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5407)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Fixed bug causing the TM to block the simulation when another client teleported a vehicle with no physics.
   * Python agents now accept a carla.Map and GlobalRoutePlanner instances as inputs, avoiding the need to recompute them.
   * Fix a bug at `Map.get_topology()`, causing lanes with no successors to not be part of it.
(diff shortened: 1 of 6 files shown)


joel-mb (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5407)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #5443, opened 2022-05-16
Improved TM's non signalized junctions

glopezdiest:
### Description

Improved how the TM vehicles handle the non signalized intersections, resulting in an overall more fluid traffic.

Previously, a series of tickets where issued according to their time of arrival, which dictated the order of the vehicles. However, this approach had some minor bugs that caused vehicles to wait, even though the junction was empty, and overall, their behavior wasn't that intuitive.

The new approach creates a list of all entering vehicles according to their order of arrival. From there, lane priority is given to the first vehicle that arrived, and once its maneuver ends, the next one starts entering the junction. To have a smoother traffic, the maneuver ends when the vehicle has left the center of the intersection behind, and not once it has exited the junction. Note that while this might not work for 100% of junction configurations, this priorty is "just" for the TrafficLightStage, and it is later followed by the CollisionStage, which will handle possible collisions between the entering and exiting vehicles. Additionally, all vehicles will break when entering the junction and wait at the stop sign for a minimum amount of time, even if no vehicles are inside it

Additionally, a TM bug has been fixed that was causing the TM to block the simulation when trying to calculate the waypoint buffer of a vehicle that was traversing a very short road that looped onto itself.

  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** CARLA's 4.26

### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5443)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Latest
 
   * Fixed bug causing the TM to block the simulation when another client teleported a vehicle with no physics.
+  * Fixed bug causing the TM to block the simulation when travelling through a short roads that looped on themselves.
+  * Improved the TM's handling of non signalized junctions, resulting in a more fluid overall behavior.
   * Added check to avoid adding procedural trigger boxes inside intersections.
   * Python agents now accept a carla.Map and GlobalRoutePlanner instances as inputs, avoiding the need to recompute them.
   * Python agents now have a function to lane change.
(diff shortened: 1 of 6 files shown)


joel-mb (approved):
> Reviewed 4 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5443)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

joel-mb (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5443)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #5446, opened 2022-05-16
Updated the information that traffic lights sent to vehicles

glopezdiest:
### Description

This PR fixes several weird behaviors regarding the information that the traffic lights were sending to the vehicles affected by it. In other words, this changes the output of the `Vehicle.get_traffic_light_state()`, `Vehicle.is_at_traffic_light()` and `Vehicle.get_traffic_light()` functions.

While I consider these changes as bug fixes, this behavior could have made sense for other features such as the (now very old) autopilot, so it can also be seen as an update to an old behavior, which currently is more harmful than helpful.

This PR focuses in the automatically generated traffic light (used by Large maps and OpenDrive maps), as it is the only one that is code based. All the other traffic lights have had the same changes applied but as they are implemented as Blueprints in Unreal, that will be a content PR.

**Currently**, vehicles that enter the bounding box of a traffic light get added to the affected vehicles by the TL, and information is sent them. This only happenes, with either yellow or red traffic lights, being green traffic light non-existent for vehicles. Once the traffic light turns green, the vehicle was removed from the TL queue, no longer received its information, and having the default _Green_ state instead, which allowed them to keep circulating.

From that behavior, three bugs can be seen:
1) As said, green traffic lights are non-existent, so the vehicle can't access their information. `Vehicle.get_traffic_light_state()` returns *false* and `Vehicle.get_traffic_light()` returns *None*. `Vehicle.get_traffic_light_state()` returns *green*, but that's just because it is the default value (which it isn't that helpful to be honest).
2) The green-yellow and yellow-red state transitions weren't notified to the vehicle.
3) It was never checked whether or not the vehicle left the TL's bounding box, so the vehicle kept the TL's information until it turned green. This resulted in some TM vehicle getting stopped in the middle of junction, blocking the traffic.

For **the new behavior**, the vehicle will receive the TL's information as long as it is inside the bounding box, losing it once it leaves the bounding box. While inside it, and regardless of the state, `Vehicle.is_at_traffic_light()` will return *True*, `Vehicle.get_traffic_light()` will return the TL instance, and `Vehicle.get_traffic_light_state()` will be the TL state at that exact frame.

  * **Python version(s):** 3.6
  * **Unreal Engine version(s):** CARLA's 4.26

### Possible Drawbacks

This severly changes the information received by the vehicle so some users might have to adapt. From CARLA's side, the TM has been tested and it works without any issues

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5446)


Axel1092 (approved):
> Reviewed 3 of 4 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5446)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #5451, opened 2022-05-18
Fixed indentation of the assignment statement to control variable

Jaivra:
#### Description

Fixed indentation of the assignment statement to control variable, without it the previous assignment statements to this variable are useless. 

  * **Platform(s):** Ubuntu Server 20.04
  * **Python version(s):** Python 3.8.10
  * **Unreal Engine version(s):** UE4+Release-4.26

#### Possible Drawbacks

Now the agent considers the car in front of him when it's not close

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5451)


Jaivra:
I agree with you for the same assignments to *control* variable, but what about the assignment at line 290?
```python 
control = self.car_following_manager(vehicle, distance)
```
Without the correct indentation it's useless

PR #5588, opened 2022-07-15
Fixed decals when importing a map

bernatx:
#### Description

When a map is imported, some decals are created over the road to add dirty. Those decals are configured through a .json file and also use the .xodr of the map. But in the import process those files were find in all the content of the project and not only on the content of the package created. That means that if other .json file or .xodr was found, that was used instead of the correct one.

  * **Platform(s):** Windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5588)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
   * Some of the warnings shown when building a map will now only be showing when debugging.
   * Fixed bug causing traffic signals at the end points of a road to sometimes create malformed waypoints.
+  * Fixed decals when importing maps. It was using other .json files found in other packages.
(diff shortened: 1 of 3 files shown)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5588)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Axel1092 and @XGodina)

glopezdiest:
Loos good to me. Tests with the new map shows the decals correctly placed at the driving lanes

PR #5633, opened 2022-07-30
Updating docker file and docs for using github PAT when downloading Unreal Engine

DanielAtt2000:
#### Description

GitHub doesn't support passwords anymore and therefore a Personal access token needs to be used. The appropriate changes where made both in the docs and in the dockerfile. 

Fixes #  

  * **Platform(s):*Linux 18.04 LTS* ...
  * **Python version(s):*3.6.9* ...
  * **Unreal Engine version(s):*4.26* ...

#### Possible Drawbacks

None

---
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5633)

### DIFF
--- a/Docs/build_docker_unreal.md
+++ b/Docs/build_docker_unreal.md
@@ -52,10 +52,10 @@ The following steps will each take a long time.
 
 __1. Build the CARLA prerequisites image.__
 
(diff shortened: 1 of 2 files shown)


Bonifatius94:
I wouldn't name it GIT_USER. Instead I would use GITHUB_USER_WITH_UE_ACCESS to make clear that it's about a GitHub account that's allowed to download the Unreal Engine's source code.

PR #5672, opened 2022-08-16
Expose Sun Intensity from Blueprints to PythonAPI.

MarcelPiNacy:
#### Description

Exposes the sun intensity value to the Python API.

  * **Platform(s):** Windows 10
  * **Python version(s):** 3.8
  * **Unreal Engine version(s):** 4.26/CARLA

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5672)

### DIFF
--- a/LibCarla/source/carla/client/Client.h
+++ b/LibCarla/source/carla/client/Client.h
@@ -66,6 +66,10 @@ namespace client {
       _simulator->RequestFile(name);
     }
 
+    float GetSunIntensity() const {
+      return _simulator->GetSunIntensity();
+    }
+
     World ReloadWorld(bool reset_settings = true) const {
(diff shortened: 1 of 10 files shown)


glopezdiest:
Hey @MarcelPiNacy, while the implemention is correct, this was always meant to be used in a specific use-case from my side, but after some testing, the sun intensity doesn't seem to be reliable enough to use it, so I'm closing the PR.

eleboss:
I think the sun intensity in carla is not physically correct where intensity are somehow limit to around 5k lux in the bpsky and cannot be modified. I need to change the sunlight to a physically correct value, either on the UE4 side or the pythonAPI (better).

PR #5673, opened 2022-08-16
Compute displacement direction when moving signals

joel-mb:
#### Description

When moving signals out of roads we now compute the desired displacement direction. Instead of moving signals always to the right we now check the right and left lane types.

The algorithm works in the following way:
1. If the right lane is of type `Driving` we move the signal to the left.
2. If the left lane is of type `Driving` we move the signal to the right.
3. If both lanes (right and left) are of type `Driving` we don't move the signal.

  * **Python version(s):** 3,7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5673)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Improved algorithm to move signals out of the road by computing the desired displacement direction.
   * Added `TrafficManager.vehicle_lane_offset(actor, offset)` and `TrafficManager.global_lane_offset(offset)` methods.
   * Some of the warnings shown when building a map will now only be showing when debugging.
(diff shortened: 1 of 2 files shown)


glopezdiest (approved):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5673)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @joel-mb)

glopezdiest (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5673)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @joel-mb)

PR #5838, opened 2022-10-17
Prevent from segfault OnTriggerEndOverlap

berndgassmann:
In case the OtherActor is not a Vehicle just return

#### Description

Fixes #  

  * **Platform(s):** ... Ubuntu 20.04
  * **Python version(s):** ... 3.7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5838)


bernatx:
Thanks for the fix

PR #5937, opened 2022-11-13
Upgrade boost to 1.80

jerry73204:
#### Description

The major work is to upgrade the boost version to 1.80. The benefit is to allow CARLA to be compiled with Python 3.10 on Ubuntu 22.04. It also fixes compilation errors produced by newer C/C++ compilers.

- Upgrade downloaded boost package to 1.80.
- Add a missing `#include <atomic>` in `test_recurrent_shared_future.cpp`
- Add an impl of `throw_exception(const std::exception &e, boost::source_location const & loc)` that routes exceptions to `carla::throw_exception()`.

This command tested the changes with multiple Python versions. [deadsnakes](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa)'s Python binaries were used.

```bash
make PythonAPI ARGS="--python-version='3.6 3.7 3.8 3.9 3.10'"
```

  * **Platform(s)**: Ubuntu 20.04
  * **Python version(s)**: 3.6, 3.7, 3.8, 3.9, 3.10
  * **Unreal Engine version(s)**: carla's ue4 was used (commit [a6fbc841](https://github.com/CarlaUnreal/UnrealEngine/commit/a6fbc841dff629e96e4d00a167a1c873a8652e4f))

#### Possible Drawbacks

The patch was tested on Ubuntu 20.04. I don't have a Windows PC/laptop so I'm not sure it works on Windows.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5937)


jerry73204:
The PR is ready for review. The boost update enables compatibility with newer Python versions. I successfully build Python client packages with Python 3.9 and 3.10.

jerry73204:
Also, I back-ported the commits to 0.9.13. If this PR is working, I would consider opening a backporting PR.

Axel1092:
Hi @jerry73204, sorry for the late response and thank you for working on this upgrade. We indeed want to move to newer Boost versions to support latest python versions, however we cannot accept this PR as it needs to include the windows upgrade as well. In addition, there are a lot of changes `#!/usr/bin/env bash` which should not be included. We are currently working on doing the upgrade to 1.81 on both versions ourselves, so look forward to it very soon.

jerry73204:
Thanks to @Axel1092. That's good news if the Python issue can be fixed.
Would you accept a standalone PR for `#!/usr/bin/env bash`? That's a sake for portability because some Linux distros have `/bin/bash` instead of `/usr/bin/bash`. You can see this [pose](https://www.baeldung.com/linux/bash-shebang-lines) for detailed comparison.

Axel1092:
In order to accept that PR it should be tested that it does not cause any issues in our main platform targets (Ubuntu desktop). Our continuous integration pipeline would probably pick up problems but additional verification won't hurt.

PR #5973, opened 2022-11-22
Fix pedestrian bones displaced from body

bernatx:
#### Description

This is a fix for three different problems identified related to the skeleton of a pedestrian:

1. The skeleton of a pedestrian at some points was displaced from the body. That was caused by getting the bones of the pedestrian at the wrong moment. 

2. In function of the network congestion, could happen that the server start to build the frame before all the pedestrian position are updated from the client (pedestrian navigation is managed in the client side). That has been fixed using a synchronous call instead of an asynchronous call. So in synchronous mode now we are sure to update all pedestrian positions before do the tick.

3. When a pedestrian was reaching a point of the path for the route he is following, it was stop by 1 frame, and that was causing an interference of the animation because the speed was lower. The pause has been removed and now there are no glitches in the animation.

Fixes #  4980 (partially)

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5973)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   * Added `TrafficManager.vehicle_lane_offset(actor, offset)` and `TrafficManager.global_lane_offset(offset)` methods.
   * Some of the warnings shown when building a map will now only be showing when debugging.
   * Fixed bug causing traffic signals at the end points of a road to sometimes create malformed waypoints.
+  * Fixed pedestrian skeleton frame, where sometimes it was draw displaced from the body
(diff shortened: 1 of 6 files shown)


Axel1092 (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/5973)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #6042, opened 2022-12-13
Some sensors are forced to be always on the primary server in multi-g…

bernatx:
#### Description

In multi-GPU mode some sensors need to be always on the primary server, which is in charge of the physics calculation. So, the collision sensor needs to be created always on the primary server. Also, the same happens with the world observer, that needs to be always in the primary server.

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6042)


Axel1092 (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6042)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #6049, opened 2022-12-16
Fix TM hybrid respawn bug

glopezdiest:
### Description

When the Traffic Manager has the `set_respawn_dormant_vehicles` `parameter activated, it will detected dormant actors and respawn them closer to the ego vehicle.

While this works fine, using this with hybrid mode was causing the simulation to run substantially slower. This is because the teleportation makes the TM bug out and think that the vehicle is moving extremely fast, resulting in a ridiculously long waypoint buffer. 

This PR makes the TM update the saved state of the teleported vehicle,  removing the aforementioned bug.

  * **Python version(s):** 3.8

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6049)


joel-mb (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6049)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #6058, opened 2022-12-22
Multi-GPU synchronization of actors like TrafficLights

bernatx:
#### Description

In multi-GPU, each time a new connection is done (secondary server connecting to the primary server), with the frame data we include all existing actors, so that way all the traffic lights and other actors are sent to all secondary servers (specially the new one).

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

Will happen that we send actors that already exist on some secondary server. So if the actor with the Id is found, the actor is not created but reused.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6058)

### DIFF
--- a/LibCarla/source/carla/multigpu/router.cpp
+++ b/LibCarla/source/carla/multigpu/router.cpp
@@ -70,6 +70,11 @@ void Router::SetCallbacks() {
   log_info("Listening at ", _endpoint);
 }
 
+void Router::SetNewConnectionCallback(std::function<void(void)> func)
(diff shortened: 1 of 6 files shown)


Axel1092 (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6058)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #6115, opened 2023-01-13
Fix extra bytes in DirectX textures (Windows)

bernatx:
#### Description

Some extra data is added by DirectX to align pixel rows at the boundary of 256by (D3D12_TEXTURE_DATA_PITCH_ALIGNMENT).
In those cases where the size of the texture is not aligned to 256, then we need to copy row by row to remove the extra data, instead of doing just 1 copy of all the pixels.

Fixes # 6085

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

Copy an image row by row instead of just a memcpy() is slower.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Latest
+
+  * Fixed the extra data in Directx textures, so we need to copy row by row on Windows to remove extra bytes on images
+
 ## CARLA 0.9.14
 
   * Fixed bug in FrictionTrigger causing sometimes server segfault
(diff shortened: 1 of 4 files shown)


Axel1092 (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6115)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #6127, opened 2023-01-18
Correct incoherent structuring of tutorial for adding sensor to CARLA.

kis-balazs:
Add a couple corrections to the tutorial related to adding a new sensor to CARLA.
1. Change numbering style of sections
2. Adapt links to text
3. Notes are very important, and difficult to follow in the old version. Adapted them to be clear, easy to see and read.

  * **Platform(s):** GitHub Pages
  * **Python version(s):** Not applicable
  * **Unreal Engine version(s):** Not applicable

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6127)


MattRowe18 (changes requested):
Hi, thanks for these changes. One thing is that we would prefer to keep the notes in breakout boxes, if you could change the notes back into breakout boxes, we will merge this.

kis-balazs:
Hi @MattRowe18!

Would you like to keep the highlighting, but not to add the text box? Any other suggestion for highlighting which would be preferred?
Thanks!

kis-balazs:
Hi @MattRowe18!

I would like to shortly follow up on the breakout box idea. The notes, before my modifications, were simply highlighted in text, using "!!!" in front of the text. I added the blockquote solution, and I am not sure what your comment refers to.
Would highly appreciate to follow up on this, to add the changes and let's merge the documentation changes!

Cheers!

MattRowe18:
Hi @kis-balazs 

The "!!! note" is an instruction for the markdown parser (in this case mkdocs) to create a breakout box. When mkdocs builds the documentation (readthedocs.io uses mkdocs, or something similar, to build the docs) and reads an indented block with that qualifier to start, it creates a breakout box like the one in this picture. 

image

It looks odd in raw format or in GitHub preview, but once the documentation is parsed it looks much clearer. If you change the note breakouts back to the original format, we'll be pleased to merge this pull request.

Thanks

Matt

kis-balazs:
Hi @MattRowe18

You are right, I double-checked the compiled version and you are right. Spotted this already, but failed to acknowledge!

Thanks, will change asap, and then we can merge!

MattRowe18:
Great, thanks.

If you ever want to build the docs locally, to see a preview of how they will look on readthedocs.io, you can install mkdocs, then run `mkdocs serve` in the CARLA root folder, you will then find the docs hosted at `localhost:8000` in a browser. But only do it if you need to, for various configurations, things like mkdocs can have some tricky installation hiccups.

kis-balazs:
Indeed, I am running it locally. Simply missed this detail, while focusing on keeping structure :D

PR #6132, opened 2023-01-19
Use Unreal's host SDK on Linux

anrp-tri:
The supported compiler is clang, but that is not the default on Ubuntu. Unreal Editor itself ships clang-10 for use on linux (along with the libc++ that Unreal itself uses) so just use those in CARLA.

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6132)

### DIFF
--- a/PythonAPI/carla/setup.py
+++ b/PythonAPI/carla/setup.py
@@ -61,7 +61,8 @@ def get_libcarla_extensions():
                 '-Wpessimizing-move', '-Wold-style-cast', '-Wnull-dereference',
                 '-Wduplicate-enum', '-Wnon-virtual-dtor', '-Wheader-hygiene',
                 '-Wconversion', '-Wfloat-overflow-conversion',
-                '-DBOOST_ERROR_CODE_HEADER_ONLY', '-DLIBCARLA_WITH_PYTHON_SUPPORT'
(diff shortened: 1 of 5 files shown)


Axel1092:
Hi @anrp-tri, thank you for your contribution on the project. This is an interesting change, however we cannot accept it, for starters jenkins is not able to compile the project so these changes break the build system. Then, I understand the motivation to use the same compiler as Unreal uses however these changes need to be thoroughly tested so that no modules break. We understand the limitations and issues of our build system but for a change like this one we would need to allocate resources on our side to make sure that everything keeps working as it should.

Axel1092:
Hi @anrp-tri, thank you for the changes. We are looking into it to check that no other conflicts occur. If no other errors are detected we will merge this. Thank you for your effort.

Axel1092 (approved):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6132)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @anrp-tri)

PR #6158, opened 2023-01-27
Added ID to get_level_bbs()

cscribano:
#### Description

I extended the function get_level_bbs() to return a list of pairs (id, BoundingBox), where the id is obtained from UE, identically to the ID returned by the instance segmentation sensor. 
Thanks to this change, the returned Bounding Boxes can be matched with the Instance Segmentation in order to effectively filtering the visible vehicles. I believe this feature could be extremely handy to generate object detection labels.

  * **Python version(s):** 3.8

#### Possible Drawbacks

* In order to match the IDs from the InstSeg, only the 16 LSB of the 32-bit AActor.GetuniqueID() are kept, this might potentially lead to colliding IDs.
* This might generate confusion with the Actor.id

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6158)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
   * Fixed the extra data in Directx textures, so we need to copy row by row on Windows to remove extra bytes on images
   * Restored gamma value to 2.2 instead of 2.4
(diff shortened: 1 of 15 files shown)


cscribano:
I think I somehow screwed the commit history, I will open a new PR as soon as I figure this issue.
Sorry for the inconvenience.

PR #6166, opened 2023-01-31
Adding --ignore-spectator to the replayer

bernatx:
#### Description

When using a package to record and replay a simulation, the spectator can be replayed or not with this new API function/flag.
By default the spectator is ignored when replaying.

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6166)


LuisPoveda:
Also tested in Ubuntu 20.04 with Python 3.8

PR #6169, opened 2023-02-01
Fixed Traffic Manager's unstuck logic

### LINKED ISSUE #6110: Traffic Manager blocked vehicle despawn logic chooses wrong time threshold
The traffic manager has a logic to de-spawn vehicles that are standing still for longer than 90 seconds.
This time should get extended to 180 seconds if the vehicle is standing at a red light.
The relevant code logic that, determines whether a vehicle counts as stuck, can be found here:
https://github.com/carla-simulator/carla/blob/be069dda537afa2fff698b70431714489f5c0564/LibCarla/source/carla/trafficmanager/ALSM.cpp#L361

I think the if statement logic is bugged:
```
if ((!tl_state.at_traffic_light && tl_state.tl_state != TLS::Red && delta_idle_time >= BLOCKED_TIME_THRESHOLD)
    || (delta_idle_time >= RED_TL_BLOCKED_TIME_THRESHOLD)) {
      ... despawn vehicle
    }
```
You can go through different cases to see what happens:

- The vehicle is at a traffic light: This will make the entire first big statement false, without checking for the traffic light state and use the extended time for red lights. (This is unintended, as in cases of green lights the normal threshold BLOCKED_TIME_THRESHOLD should be used.)
- The vehicle is not at a traffic light: Now the code will check the traffic light state, which is obviously nonsense. Depending on the state of a traffic light the car is NOT affected by, one of the two thresholds will be chosen.

The time constants are defined here:
https://github.com/carla-simulator/carla/blob/be069dda537afa2fff698b70431714489f5c0564/LibCarla/source/carla/trafficmanager/Constants.h#L28

glopezdiest:
### Description

The Traffic Manager has a logic that checks there are no vehicles stuck, removing them after a certain amount of time has passed without them moving. This logic, however, is broken in two ways, which this PR fixes them.

1) When deciding which vehicle is the "most stuck" (aka the one that has been stopped the most time), the logic considered all hero vehicles, regardless of whether or not they were part of the TM. This breaks the logic if an unregistered hero is the one actually stuck.
2) The time needed to be considered stuck is increased for vehicles are red lights / stops, but the differentiation between this and the "normal" case is not done correctly. 

Fixes #6110

  * **Python version(s):** 3.8

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6169)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Fixed bug causing the TM's unstuck logic to incorrectly remove the vehicles in some situations.
   * Fixed the extra data in Directx textures, so we need to copy row by row on Windows to remove extra bytes on images
   * Restored gamma value to 2.2 instead of 2.4
 
--- a/LibCarla/source/carla/trafficmanager/ALSM.cpp
+++ b/LibCarla/source/carla/trafficmanager/ALSM.cpp
@@ -181,20 +181,19 @@ void ALSM::UpdateRegisteredActorsData(const bool hybrid_physics_mode, ALSM::Idle
     if (is_respawn_vehicles) {
       track_traffic.SetHeroLocation(hero_actor_info.second->GetTransform().location);
     }
-    UpdateData(hybrid_physics_mode, max_idle_time, hero_actor_info.second, hero_actor_present, physics_radius_square);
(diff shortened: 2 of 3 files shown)


joel-mb (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6169)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @glopezdiest)

PR #6246, opened 2023-02-24
Add keyword arguments for carla.TrafficManager Python API functions

xqgex:
#### Description

Add keyword arguments for `carla.TrafficManager` Python API functions.

The kwargs are based on https://carla.readthedocs.io/en/0.9.14/python_api/#carlatrafficmanager

  * **Python version(s):** 3.7


glopezdiest:
Hey, thanks for the PR. First of all, I see no issues with accepting this, as this will only add functionailities, and compatibility shouldn't be an issue here. A very minor concern, any reason why you decided to add default values to some of the functions?

xqgex:
There shouldn't be any different of "default values" from what is written in the API, which function are you referring to?

glopezdiest:
Both *vehicle_percentage_speed_difference* and *global_percentage_speed_difference* have a default value of 30? Why not 0?
Also, some functions have default values set to 0, such as *vehicle_lane_offset*, but why doesn't *random_left_lanechange_percentage* have it too?

xqgex:
It's all according to the API (https://carla.readthedocs.io/en/0.9.14/python_api/)

### [`vehicle_percentage_speed_difference`](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.TrafficManager.vehicle_percentage_speed_difference)

> * `vehicle_percentage_speed_difference(self, actor, percentage)`
>   Sets the difference the vehicle's intended speed and its current speed limit. Speed limits can be exceeded by setting the perc to a negative value. **Default is 30**. Exceeding a speed limit can be done using negative percentages.
>   * Parameters:
>     * `actor` ([carla.Actor](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.Actor)) - Vehicle whose speed behaviour is being changed.
>     * `percentage` (float) - Percentage difference between intended speed and the current limit.

### [`vehicle_lane_offset`](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.TrafficManager.vehicle_lane_offset)

> * `vehicle_lane_offset(self, actor, offset)`
>   Sets a lane offset displacement from the center line. Positive values imply a right offset while negative ones mean a left one. **Default is 0**. Numbers high enough to cause the vehicle to drive through other lanes might break the controller.
>   * Parameters:
>     * `actor` ([carla.Actor](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.Actor)) - Vehicle whose lane offset behaviour is being changed.
>     * `offset` (float) - Lane offset displacement from the center line.

### [`random_left_lanechange_percentage`](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.TrafficManager.random_left_lanechange_percentage)

> * `random_left_lanechange_percentage(self, actor, percentage)`
>   Adjust probability that in each timestep the actor will perform a left lane change, dependent on lane change availability.
>   * Parameters:
>     * `actor` ([carla.Actor](https://carla.readthedocs.io/en/0.9.14/python_api/#carla.Actor)) - The actor that you wish to query.
>     * `percentage` (float) - The probability of lane change in percentage units (between 0 and 100).

glopezdiest:
I see. Well they aren't really default values, but the way the API is written seems to indicate it so.

Can you please remove those values? These functions shouldn't really have a default, as they are meant for users to change the behaviors to specific values. With that, I can resolve the CHANGELOG conflicts and merge this

xqgex:
So if the following call is not a valid API call

```python
tm.vehicle_lane_offset(my_actor)
```

Then the documentation should be updated...

---

I want to summarize this so you can confirm what the intended behavior is from the API calls.

The following should have default values:

1. `set_hybrid_physics_mode(enabled=False)`
1. `set_hybrid_physics_radius(r=50.0)`
1. `set_osm_mode(mode_switch=True)`
1. `set_respawn_dormant_vehicles(mode_switch=False)`
1. `set_synchronous_mode(mode_switch=True)`

The following should NOT have default values:

1. `global_lane_offset(offset)`
1. `global_percentage_speed_difference(percentage)`
1. `update_vehicle_lights(actor, do_update)`
1. `vehicle_lane_offset(actor, offset)`
1. `vehicle_percentage_speed_difference(actor, percentage)`

And what about the following functions:

1. `auto_lane_change(actor, enable=True)`
1. `collision_detection(reference_actor, other_actor, detect_collision=True)`

?

xqgex:
None of the 12 functions should have a default value? if so there must be a complimentary update to the API documentation.

Because for example `set_hybrid_physics_mode(self, enabled=False)` (https://carla.readthedocs.io/en/0.9.14/python_api/#carla.TrafficManager.set_hybrid_physics_mode) the documentation specifically says there is a default value for `enabled`

glopezdiest:
While that is true, the docs are wrong. You actually said it yourself, that you had to add the default values.

Don't worry about the docs though, we'll update them to match these new arguments. We'd normally ask you to do so, but as we are actually revising the PythonAPI docs right now, we can do it ourselves along with the rest of the changes. Tagging @MattRowe18

glopezdiest:
Great, this should be everything, thanks :) We'll merge it whenever we can

PR #6289, opened 2023-03-07
Fixing import script that can find a TilesInfo.txt from other existing map

bernatx:
#### Description

When the import script runs to import a large map, it searches for a TilesInfo.txt file to know where the top,left cell has the center in the world. The problem is that in the destination folder can exist other large maps with that file, and the import script was searching recursively from the base folder (the package folder) and was taking the first TilesInfo.txt found, that could be from another existing map.

That has been fixed, and now the TilesInfo.txt file is searched only in the destination map folder, instead of the destination package folder.

  * **Platform(s):** windows
  * **Python version(s):** 3.7

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6289)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
   * Fixed bug causing the TM's unstuck logic to incorrectly remove the vehicles in some situations.
   * Fixed the extra data in Directx textures, so we need to copy row by row on Windows to remove extra bytes on images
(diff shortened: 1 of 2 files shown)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6289)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @bernatx)

PR #6526, opened 2023-06-02
Anrp/sysroot

anrp-tri:
Force Unreal toolchain to use Unreal sysroot
    
    Not doing this can lead to the Unreal toolchain examining the host system,
    which can cause it to attempt to use glibc features newer than the Unreal
    target of 2.17 provides (specifically, rpclib can accidentaly pick up the
    symbol pthread_cond_clockwait, which is provided by Ubuntu 20.04 glibc 2.31
    but *not* by Unreal glibc 2.17, which then causes a missing symbol linker
    error when trying to build CARLA.)

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6526)


bernatx:
Thank you, we merge this

PR #6628, opened 2023-07-11
[Recorder-Log-fix]: There are format issue with recorder log file created by API 'show_recorder_file_info'

rajesh-gurgoan:
#### Description

I have made changes for the following reasons
1. When I called client API 'show_recorder_file_info', we have text log file. Here I found there are some mismatch in the pattern for 'Positions" attribute
      e.g. 
            Id: 1 Location: (-17219.7, 18385.9, 2763.81) Rotation (0, 0, -45.5921)
      We do not have COLON(:) for Rotation attribute. It should be like
            Id: 1 Location: (-17219.7, 18385.9, 2763.81) Rotation: (0, 0, -45.5921)

2. Similar issue with 'Vehicle animations' attribute  
      e.g. 
            Id: 24 Steering: 0 Throttle: 0.341554 Brake 0 Handbrake: 0 Gear: 2
      We do not have COLON(:) for Brake attribute. It should be like
            Id: 24 Steering: 0 Throttle: 0.341554 Brake: 0 Handbrake: 0 Gear: 2
Fixes #  No issue in the fix

  * **Platform(s):** Windows10
  * **Python version(s):** 3.7.2
  * **Unreal Engine version(s):** 0.9.14

#### Possible Drawbacks

There should not be any drawback. As it is a enhancement

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6628)


bernatx:
Thank you for fixing it, that makes sense. 
But we need to merge to **dev** branch instead of **master**.

bernatx (approved):
Thank you for fixing it, that makes sense. But we need to merge to **dev** branch instead of **master**.

rajesh-gurgoan:
> Thank you for fixing it, that makes sense. But we need to merge to **dev** branch instead of **master**.

+1, yes

PR #6778, opened 2023-09-17
Add VS2022 and Ninja build support for Libcarla and OSM2ODR on Windows

alamo3:
#### Description

Fixes #  

The current Windows build system was hard-coded to only work with Visual Studio 2019. This change opens it up to using other visual studio versions and build systems such as Ninja which is already being used for the Linux build. Every build script has an extra "generator" flag that can override the default setting of Visual Studio 16 2019 which is still the default version used. Documentation has also been updated to reflect this. 

  * **Platform(s):** Windows
  * **Python version(s):** N/A
  * **Unreal Engine version(s):** N/A

#### Possible Drawbacks

I do not forsee any negative impacts since the default build method is still there (MSVC 2019), this change just makes it easier for others building libcarla to integrate it into their own cmake builds.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6778)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Latest
 
+  * Added build support for VS2022 and Ninja for LibCarla and osm2odr on Windows
   * Fixed bug causing the TM's unstuck logic to incorrectly remove the vehicles in some situations.
   * Fixed the extra data in Directx textures, so we need to copy row by row on Windows to remove extra bytes on images
   * Fixed sensors to check for the stream to be ready (race condition)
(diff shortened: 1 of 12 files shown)


bernatx (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6778)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alamo3)

bernatx:
Thanks for the addition, that makes sense

PR #7006, opened 2023-12-15
Fix problems of lanemarking generation and tree generation

synkrotron:
#### Description
This is a supplement and fix to my last PR: https://github.com/carla-simulator/carla/pull/6995
To calculate the edge points of lane markings, we need to compute the unit vector perpendicular to the lane line. In this process, if the coordinates of the two corners of the lane are equal, it will cause the error of dividing by 0.
The previous PR verifies if the lane width is 0 to avoid this error. However, even if the lane width is not 0, the two points on the edge of the lane may be the same, as in the picture. So I use "**if (edges.first! = edges.second)** "to avoid mistakes.
After testing for several times, it worked well.
In addition, this judgment needs to be made when generating the TreePosition.
lane1

  * **Platform(s):** Ubuntu 20.04 + Windows 10
  * **Python version(s):** Python 3.7.16
  * **Unreal Engine version(s):** 4.26-carla

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7006)

### DIFF
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -1232,6 +1232,7 @@ namespace road {
             while(s_current < s_end){
               if(lane->GetWidth(s_current) != 0.0f){
                 const auto edges = lane->GetCornerPositions(s_current, 0);
+                if (edges.first == edges.second) continue;
                 geom::Vector3D director = edges.second - edges.first;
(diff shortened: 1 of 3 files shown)


Blyron (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7006)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @synkrotron)

PR #7112, opened 2024-02-02
Update semantic labels in doc (changed from version 0.9.13 -> 0.9.14)

### LINKED ISSUE #6902: Semantic segmenation raws labels are not coherent with the documentation
_CARLA version_: 0.9.14
_Platform/OS_: Windows
_Problem you have experienced:_
By saving the semantic segmentation label into npy, I didn't get the label value written in the documentation: https://carla.readthedocs.io/en/latest/ref_sensors/#semantic-segmentation-camera but those from https://github.com/carla-simulator/carla/blob/master/LibCarla/source/carla/image/CityScapesPalette.h 

Example: label of the road is 1 and not 7 in my case.

These when, I use a basic function labels_to_cityscapes_palettes in python (ex: https://github.com/carla-simulator/data-collector/blob/master/carla/image_converter.py) the colors are wrong

_Steps to reproduce:_
np_image = np.reshape(np.copy(data.raw_data), (data.height, data.width, 4))
label = np_image[:, :, 2] # class Id is stored in B channel
np.save(npy_file, label)

Then, do we agree that the labels of the raw data are not those of the cityscape palette or did I miss something ?

dorn-gerhard:
#### Description

I updated the semantic labels used since 0.9.14 in the docs

Fixes #6902   

  * **Platform(s):** ...
  * **Python version(s):** ...

### DIFF
--- a/Docs/ref_sensors.md
+++ b/Docs/ref_sensors.md
@@ -711,34 +711,39 @@ This raw [carla.Image](python_api.md#carla.Image) can be stored and converted it
 raw_image.save_to_disk("path/to/save/converted/image",carla.ColorConverter.cityScapesPalette)
 ```
 
-The following tags are currently available:
+The following tags are currently available (Note, tags changed from version 0.9.13 to 0.9.14):
 
 | Value          | Tag            | Converted color  | Description      |
 | ----------------------------------- | ----------------------------------- | ----------------------------------- | ----------------------------------- |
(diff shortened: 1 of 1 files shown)


Blyron:
Please reopen de PR and put it against dev
We do not merge PRs against master

dorn-gerhard:
ah sorry, sure ;)

PR #7134, opened 2024-02-08
Implement Ubuntu Build

xavisolesoft:
- Set clang16 as default toolchain.
- Add GTest as optional dependency.
- Make SUMO, xercesc and proj, really optional via OSM2ODR flag.
- Link with Unreal libc++ instead listdc++.

### DIFF
--- a/Configure.py
+++ b/Configure.py
@@ -59,24 +59,21 @@ DEFAULT_C_COMPILER = FindExecutable([
   'cl',
   'clang-cl',
 ] if os.name == 'nt' else [
-  'clang',
-  'gcc',
+  'clang-16',
 ])
 
 DEFAULT_CPP_COMPILER = FindExecutable([
   'cl',
   'clang-cl',
 ] if os.name == 'nt' else [
-  'clang++',
-  'g++',
+  'clang++-16',
 ])
 
 DEFAULT_LINKER = FindExecutable([
   'link',
   'llvm-link',
 ] if os.name == 'nt' else [
-  'lld',
-  'ld',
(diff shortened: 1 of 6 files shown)


Configure.py
```diff
@@ -291,9 +291,17 @@ def SyncArgs():
 LIBCARLA_INSTALL_PATH = WORKSPACE_PATH / 'Install' / 'LibCarla'
 # Language options
 C_COMPILER = ARGV.c_compiler
+if not C_COMPILER:
+  print("Error C Compiler not found")
```
xavisolesoft, line 295: This should be sys.exit("message")

Configure.py
```diff
@@ -970,9 +985,19 @@ def BuildSQLite():
...
         '-march=native',
         '-O2',
+        '-Wno-error=int-conversion',
+        '-fPIC',
+        '-nostdinc++',
+        '-nostdlib++',
+        '-isystem',
+        f'{UNREAL_ENGINE_PATH}/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v22_clang-16.0.6-centos7/x86_64-unknown-linux-gnu/usr/include/',
+        '-L', 
+        f'{UNREAL_ENGINE_PATH}/Engine/Extras/ThirdPartyNotUE/SDKs/HostLinux/Linux_x64/v22_clang-16.0.6-centos7/x86_64-unknown-linux-gnu/usr/lib64',
```
xavisolesoft, line 998: Check if - L flags are needed, probably libc can be get from the system.

Issue #7136, opened 2024-02-08
waypoint.next()/waypoint.previous() issue when connecting two lanes in opposite direction


matejm42:
**CARLA version:** 0.9.15
**Platform/OS:** Ubuntu 20.04
**Problem you have experienced:** When calling `wp.next(150)` it returns a list with a waypoint at the same position as `wp`
**What you expected to happen:** Return the waypoint 150 meters from the `wp` or an empty list `[]`
**Steps to reproduce:** Call `wp.next(150)` from road 25, lane -8 on map Town06

<details> 
<summary> Code that causes the bug  </summary>

```python
import carla
adress = "10.35.95.3"
port = 2084

client = carla.Client(adress, port)
client.set_timeout(30)
world = client.load_world("Town06")
map = world.get_map()


wpev: carla.Waypoint = map.get_waypoint(
                carla.Location(215, 54, 0),
                project_to_road=True,
                lane_type=carla.LaneType.Driving)

wpcurb = wpev.get_right_lane().get_right_lane()
print(wpcurb.transform.location, f"{wpcurb.road_id=}, {wpcurb.lane_id=}")

wpnext1 = wpcurb.next(1)[0]
print(wpnext1.transform.location, f"{wpnext1.road_id=}, {wpnext1.lane_id=}")

wpnext2 = wpcurb.next(12)[0]
print(wpnext2.transform.location, f"{wpnext2.road_id=}, {wpnext2.lane_id=}")


wpnext3 = wpcurb.next(14)[0]
print(wpnext3.transform.location, f"{wpnext3.road_id=}, {wpnext3.lane_id=}")

wpnext4 = wpcurb.next(16)[0]
print(wpnext4.transform.location, f"{wpnext4.road_id=}, {wpnext4.lane_id=}")

wpnext5 = wpcurb.next(152.188)[0]
print(wpnext5.transform.location, f"{wpnext5.road_id=}, {wpnext5.lane_id=}")

```

> Location(x=214.900406, y=54.925457, z=0.000000) wpcurb.road_id=25, wpcurb.lane_id=-8
Location(x=215.930038, y=54.931492, z=0.000000) wpnext1.road_id=25, wpnext1.lane_id=-8
Location(x=221.930862, y=55.283535, z=0.000000) wpnext2.road_id=745, wpnext2.lane_id=1
Location(x=220.301590, y=54.936768, z=0.000000) wpnext3.road_id=745, wpnext3.lane_id=1
[...]

PR #7137, opened 2024-02-08
Fix waypoints next and previous when connecting two lines with opposite directions

matejm42:
#### Description

Fixing issue #7136 - detailed description there

Fix waypoints next and previous when connecting two lines with opposite directions.

Probably not the best solution, but it works for our case. The waypoint that would cause conflict when calling next from it is not returned.

  * **Platform(s):** Ubuntu 20.24
  * **Python version(s):** 3.8 and 3.11.6

#### Possible Drawbacks

Loading maps and getting waypoints from long distances can be a bit slower. However I did not experience any decrease in performance in my case. 

You can not get the next waypoint, if calling next from that waypoint would cause a loop. (We found this situation only on curbs connecting two lanes in opposite directions. In this case, it makes sense for us to behave as if the curbs are not connected together. Hopefully, there are no opposite-direction driving lanes connected together. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7137)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+  * Fixed waypoint.next and previous causing loops when connecting two opposite-direction lanes.
```
wentasah, line 1: Isn't this better?
```suggestion
  * Fixed waypoint.next and .previous causing loops when two opposite-direction lanes follow each other in the map.
```

Blyron:
Please point this to UE4_DEV branch pls

matejm42:
> Please point this to UE4_DEV branch pls

Ah mb; changed

Blyron:
Could you resolve CHANGELOG.md conflicts?

Blyron (commented):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7137)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matejm42)

Blyron (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7137)* status: all files reviewed, 1 unresolved discussion (waiting on @matejm42)

PR #7153, opened 2024-02-16
Javiergr cs/get bones transform

javiergrCS:
#### Description

Added and exposed in the Python API new getter functions to obtain transforms and names of bones, components and sockets of actors

Fixes #  


Util/BuildTools/Setup.sh
```diff
@@ -166,8 +181,12 @@ else
 
   log "Retrieving rpclib."
 
+  start_download_time=$(date +%s)
+
   git clone -b ${RPCLIB_PATCH} https://github.com/carla-simulator/rpclib.git ${RPCLIB_BASENAME}-source
 
+  end_download_time=$(date +%s)
+  echo "Elapsed Time downloading rpclib: $(($end-$start)) seconds"
```
xavisolesoft, line 189: Seems this line should be like this:

```bash
echo "Elapsed Time downloading rpclib: $(($end_download_time-$start_download_time)) seconds"
```

As end and start variables have different names here.

Util/BuildTools/Setup.sh
```diff
@@ -92,15 +92,30 @@ for PY_VERSION in ${PY_VERSION_LIST[@]} ; do
     BOOST_PACKAGE_BASENAME=boost_${BOOST_VERSION//./_}
 
     log "Retrieving boost."
+
+    start=$(date +%s)
```
xavisolesoft, line 96: Note that there is a simpe way in bash to measure and print the execution time of a command. Just writing `time <command_to_be_timmed>`:
```bash
time wget "https://archives.boost.io/release/${BOOST_VERSION}/source/${BOOST_PACKAGE_BASENAME}.tar.gz" || true
```

This command will print the output of the timmed command followed by this output:
```
real    0m0.015s
user    0m0.000s
sys     0m0.031s
```

The output of time command can be also modified by arguments:
https://man7.org/linux/man-pages/man1/time.1.html

xavisolesoft (changes requested):

Issue #6705, opened 2023-08-10
carla.DebugHelper visualizations polluting Sensor Data


saippua:
Our university is using Carla for testing various vehicle controller algorithms. In our testing we are looking for a way to visualize certain things in Carla while simulating, for example the intended path of the vehicle, various detected points of interest, etc.

The carla.DebugHelper draw functions are basically what we are looking for, however we find that these helpers also show up on RGB sensor data. The sensor data is used as input for the controller, and corrupting the sensor data with debug visualizations breaks the controllers.

This issue seems to be present on all the draw functions except `draw_string`. I'm assuming because the text rasterization is done in a separate pass from the main render pass which renders the server-side view as well as sensor datas. I wonder if this functionality could be extended to the other draw functions as well, perhaps through an optional argument, to retain current functionality for those that need it.

PR #7168, opened 2024-02-20
Debug drawing extension to allow drawing primitives on HUD layer

saippua:
#### Description

This PR adds the ability to draw the `DebugHelper` primitives on the HUD layer as well. This allows drawing debug primitives which do not show up in sensor data. New PythonAPI functions are implemented with the naming `DebugHelper.draw_hud_<shape>()` alongside existing draw functions `DebugHelper.draw_<shape>()`. The python function signatures are otherwise equivalent to their existing non-HUD counterparts.

Here is an example. I am drawing the old primitives (red) and new primitives (blue) side by side. The first image shows the server-side window where all the primitives are visible, and the second image shows an `sensor.camera.rgb` sensor output. The new primitives do not show up in the sensor output, which means they can be used for server-side visualization without polluting sensor data. Old primitives are drawn with default properties, and the HUD primitives are drawn with `thickness=10` to match the old primitives. A `thickness` of 1 equates to 1 pixel wide on the new HUD primitives.
Server-side window screenshot

RGB Sensor output

This PR is a proposed solution to the issue #6705.

  * **Python version(s):** 3.7
  * **Unreal Engine version(s):** 4.26 (Carla fork)

#### Possible Drawbacks

Very simple feature, so there isn't much that could go wrong.


Blyron:
Hello saipua, could you merge dev branch into yours so I can use our jenkins to test everything works fine?

saippua:
Hi, I've updated my branch to dev

Blyron (approved):

PR #7174, opened 2024-02-22
Named and type-hint supporting detection results for the python agents.

Daraan:
#### Description

When I worked with the agent functions at first I had some trouble understanding the semantics of the return types of the agent's detection functions (obstacle, traffic lights).

e.g `return (False, None, -1)`

This PR adds Python <3.6 ( including Python2.7 ) compatible named tuples as return values for these two functions. Adding support for type-hints and improving semantics. Aside from the wrapping of return values into namedtuples **this PR changes no functionality of the existing code.**

For Python <3.6 it uses `collections.namedtuple` for Python 3.6+ it uses the typed version `typing.NamedTuple` . Type checkers running on 3.6+ can successfully pick up the typed version even if the non-typed version is used during runtime.

I think the agents are itself not Python2.7 compatible, so I think it might be overkill.
**If stating that the PR is allowed to only support Python3.6+ it could be made much cleaner.**

---

  * **Platform(s):** Ubuntu 22.04 (Only python code so not relevant)
  * **Python version(s):** 2.7, 3.8, 3.10

#### Possible Drawbacks

Seeing None, despite adding one additional file containing the named tuples. They could be added to the BasicAgent.py themselves. In my opinion it would add unnecessary contents the agent file itself.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7174)


Daraan:
A comment if license or copyright information and in which form should be added to the new file is appreciated.

glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7174)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

glopezdiest:
Approved, this does make thing more readable, so nothing against it. No need to add copyright or license information

PR #7194, opened 2024-02-29
Implement UE5 ROS2 native build with dynamic linking

xavisolesoft:
### ROS2
- Implement ROS2 native build for UE5 Ubuntu.
- Changing ROS2 native build from static library to shared library to prevent mismatch between c++ library versions, as ROS2 is compiled with g++ old abi and Unreal is compiled with clang abi.
- ROS2 native library renamed from `libcarla_fastdds.a` to  `libcarla-ros2-native.so`.
- boost_asio and tinyXML libs are now downloaded and builded by fast-dds cmake internally.
- boost was reintroduced inside ROS2 native code to remove code from newer standards than C++11. Pending to see if it's possible to build ROS2 dependencies with a newer version of the standard.

### CARLA Configure.py
- Added log for source download step.

### DIFF
--- a/LibCarla/cmake/ros2-native/CMakeLists.txt
+++ b/LibCarla/cmake/ros2-native/CMakeLists.txt
@@ -1,0 +1,50 @@
+cmake_minimum_required (VERSION 3.20.0)
+
+project (carla-ros2-native)
+
+set (CARLA_WORKSPACE_PATH ${CMAKE_CURRENT_SOURCE_DIR})
+set (LIBCARLA_SOURCE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../source)
+
+file (
+    GLOB
+    LIBCARLA_ROS2_SOURCES
+    ${LIBCARLA_SOURCE_PATH}/carla/ros2/publishers/*.cpp
+    ${LIBCARLA_SOURCE_PATH}/carla/ros2/subscribers/*.cpp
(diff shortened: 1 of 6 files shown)


LibCarla/cmake/ros2-native/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    ${FASTDDS_INCLUDE_PATH}
+    ${BOOST_INCLUDE_PATH}
+)
+
+target_compile_definitions (
+    carla-ros2-native PUBLIC
+    BOOST_ASIO_ENABLE_BUFFER_DEBUGGING
+)
+
+target_link_libraries(carla-ros2-native /home/xsole/programing/carlaUE5/Install/fast-dds-install/lib/libfastrtps.so /home/xsole/programing/carlaUE5/Install/fast-dds-install/lib/libfastcdr.so)
```
LuisPoveda, line 44: can you change these paths to be relative instead?

LuisPoveda (changes requested):
Good job!
This new pipeline looks groovy!

LibCarla/cmake/ros2-native/CMakeLists.txt
xavisolesoft: Sure. I'm working on it, thank you!

LuisPoveda (approved):
:+1:

PR #7226, opened 2024-03-08
Check if clients are listening

Blyron:
Check if sensors are being listened by a client to avoid sending data if not

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/PixelReader.h
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/PixelReader.h
@@ -121,7 +121,8 @@ void FPixelReader::SendPixelsInRenderThread(TSensor &Sensor, bool use16BitFormat
       if (!Sensor.IsPendingKill())
       {
         FPixelReader::Payload FuncForSending =
-          [&Sensor, Frame = FCarlaEngine::GetFrameCounter(), Conversor = std::move(Conversor)](void *LockedData, uint32 Size, uint32 Offset, uint32 ExpectedRowBytes)
+          [&Sensor, Frame = FCarlaEngine::GetFrameCounter(), Conversor = std::move(Conversor)]
+          (void *LockedData, uint32 Size, uint32 Offset, uint32 ExpectedRowBytes)
           {
             if (Sensor.IsPendingKill()) return;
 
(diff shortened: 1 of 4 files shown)


MarcelPiNacy (approved):

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/PixelReader.h
```diff
@@ -218,19 +219,25 @@ void FPixelReader::SendPixelsInRenderThread(TSensor &Sensor, bool use16BitFormat
                 }
                 #endif
 
-                // network
-                SCOPE_CYCLE_COUNTER(STAT_CarlaSensorStreamSend);
-                TRACE_CPUPROFILER_EVENT_SCOPE_STR("Stream Send");
-                Stream.Send(Sensor, BufView);
+                if (Sensor.AreClientsListening())
```
xavisolesoft, line 222: Note that if is redundant because `FuncForSending` `PayLoad` is only called if `WritePixelsToBuffer` is called.

I suggest remove this if and keep the ode as it was.

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/Sensor.h
```diff
@@ -63,6 +63,11 @@ class CARLA_API ASensor : public AActor
     return Stream.IsStreamReady();
   }
 
+  bool AreClientsListening()
```
xavisolesoft, line 66: The implementation should be moved tho the .cpp file.

Implementations on .hpp files increase the compilation time and are forcely inlined by the compiler.

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/PixelReader.h
```diff
@@ -218,19 +219,25 @@ void FPixelReader::SendPixelsInRenderThread(TSensor &Sensor, bool use16BitFormat
...
               }
             }
           };
 
-          WritePixelsToBuffer(
-              *Sensor.CaptureRenderTarget,
-              carla::sensor::SensorRegistry::get<TSensor *>::type::header_offset,
-              InRHICmdList,
-              std::move(FuncForSending));
+          if (Sensor.AreClientsListening())
```
xavisolesoft, line 233: **Suggested changes**
Mi recommendation is to move this if to the line 121 of this file.

** Resulting Code**
Line 121 will result like this:
```c++
if (!Sensor.IsPendingKill() && Sensor.AreClientsListening())
```

**Benefit**
- More readable code.
- A bit of performance improvement due we are avoiding more code to be executed when there are no clients listening.
- Will be easier to modify the ode in the future.

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/WheeledVehicleAIController.cpp
```diff
@@ -109,7 +109,10 @@ void AWheeledVehicleAIController::OnPossess(APawn *aPawn)
   Vehicle = Cast<ACarlaWheeledVehicle>(aPawn);
   check(Vehicle != nullptr);
   MaximumSteerAngle = Vehicle->GetMaximumSteerAngle();
-  check(MaximumSteerAngle > 0.0f);
+  if (MaximumSteerAngle > 0.0f) {
```
xavisolesoft, line 112: Note that in CarlaUE5 `ACarlaWheeledVehicle::GetMaximumSteerAngle()` is always 0, so `MaximumSteerAngle > 0.0f` is always false.

xavisolesoft (changes requested):

xavisolesoft (approved):

PR #7229, opened 2024-03-08
Rework ros2 support

berndgassmann:
#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7229)


berndgassmann:
[CARLA-ROS2-Integration.pptx](https://github.com/carla-simulator/carla/files/14539728/CARLA-ROS2-Integration.pptx)

I've created a basic PowerPoint describing a bit what is implemented...

berndgassmann:
I got it compiled and linked under windows now; had to rename actor -> actors as windows file system is ... after that long time still not able to distiguish between LARGE letters and small letters ... incredible. I also cannot test if ROS2 works on windows... because I had to find out that: oh windows 11 is not supported by ROS2, yet. Theoretically, it should work though given that not some firewall stuff is interfering with the ports required by ROS2/DDS. But crashing when starting up on creating DdsDomainParticipantImpl() with access violation is pointing so some stuff like that. Maybe we should leave the debugging of this under Windows to someone who actually needs that there and has experience with DDS/ROS under Windows.

berndgassmann:
Maybe some comment here on what's actually missing on this now is mainly:
- transfer the synchonization from ros-bridge to here in the sense that we don't just tick by some single instance, but rather every vehicle can decide it it wants to participate in "ticking"; but the idea would be rather a windowing (every client tells what future timestamp he will accept) than a simple tick -- that is what a simulation of multiple vehicles stacks actually require. A vehicle stack knows that while it is evaluating the sensor data there is multiple sensor data upcoming; therefore a let's day 40ms delta time into the future is for most controllers rather usual. Like this the overall simulation speed-up is massive, because the simulator can run at the maximum speed, the hardware is able to provide mostly. Even with massiv multi-vehicle setup.
 At some point the synchoniszation will come, but that portion can be merged and used before.
 The bridging parts of the old ROS bridge can then be removed.
 The supporting/scripting/controlling parts though will NOT BE REPLACED by this. The internal CARLA ROS support will not refine the topic data to shiny perfectly data. That has to be performed on ROS side in each case.
 One open of this ROS2 support -- which has to be clarified over time, I guess -- is the threading: do we need to call the publisher in an async way to not block the server, or is DDS already taking care for such. If DDS is able to block on the publish() call, then a std::async() must be placed to not block the carla server from continuing.
 I know: quite some things to be taken into account, in addition. But I won't be the one to look into these details. I implemented the initial version of the "new" ROS bridge in a C++ like python; now I had the chance to remove the loose python contracts by strong C++ implementation. The rest is up to the community. Thanks.

Blyron:
Hey Bernd,

We are sorry from being late on your PR. We really like your PR and we think it takes very interesting things about ROS2 functionality, but we think we need to focus the PR on ROS2. We are aware that you have introduced quite useful functionality in CARLA, but in order to be tested properly. In order to provide better feedback, first we would like to focus on ROS2 to test it and then after the merge of ROS2 on the introduction of new classes or other stuff. This other functionality could go into another PR so we can test it properly too. 

Would be possible to have a PR focused in ROS2?  

Apart from that we had some questions and comments about your code;

Questions:
- Why are you using git submodules?
- Why have you changed gcc-7? Do you think changing gcc version can be risky?
- Are we mixing coordinate systems with Quaternions class?
- Why have you renamed CarlaTopics from Carla...Publisher to Ue...Publisher?

Comments:
- We are not gonna use a single clang format for the whole proyect, LibCarla is using google coding standard, but in Unreal we are gonna use Unreal's coding standard. So we think that if we want to include clang formatters they should go into a different PR.
- We would like to know the reason of changing namespaces and some files' path.
- The changes from MakeSafeUnitVector to MakeUnitVector needs to be reverted, we always want to make sure the unit vector is safely  normalized as the location can be (0,0,0)
- What is the purpouse of RightHandedVector?.
- In build.cs we think we need to take care of Mac by adding IsLinux() 
- What is the purpouse in ASyncDataStreamImpl Transform RelativeTransform and Rotation?
- Why are we making FCarlaServer inherit from ros2 interface? Does it work if we do not want ROS2 in the build? Or Are we making ROS2 mandatory dependency? We want to keep ROS2 as optional

Kind regards;,

CARLA team

berndgassmann:
I guess I did all cleanup possible. And merged the current compile branch (which is on top of dev) into this.

Blyron:
Could you remove .gitsubmodule?

Blyron:
Hey Bernd, can you answer my message on discord so we can have a chat about the PRs?

berndgassmann:
This PR is replaced by a step-wise (stacked) approach:
1.) #7919 
2.) https://github.com/berndgassmann/carla/pull/1
3.) https://github.com/berndgassmann/carla/pull/2
the later both will be switched to merge into dev when the previous one is merged

PR #7242, opened 2024-03-13
Fix the misuse of the traci module in line 304 of sumo_simulation.py

nwpuhkp:
#### Description

Dear developers, when I tried to run the content related to the SUMO co-simulation in the official documentation, I encountered an error stating that "module 'traci' has no attribute 'sumolib.'" Upon further investigation, I found that 'sumolib' is not a submodule of 'traci,' which is the cause of this error. The solution is to remove 'traci.' before the reference to 'sumolib.'

  * **Platform(s):** Ubuntu22.04(x86_64)
  * **Python version(s):** 3.7.16
  * **Unreal Engine version(s):**


xavisolesoft (approved):
I checked it with python console

PR #7245, opened 2024-03-13
Fixing Lidar for UE5 and be able to save its data to a .ply file using SensorSpawnerActor.

JoseM98:
#### Description

- The ParallelLineTrace method has been added to the engine, so it has been used in ShootLaser(...) method.
- Chaos ReadLock is done instead of PhysX ReadLock in SimulateLidar(...) method.
- Added the option to save .ply files with the PointCloud of the Lidar.
- To be able to access the PointCloud outside SimulateLidar(...) method, the points have been stored in an FVector. As they are stored in a CarlaLib class that has no getters.

Fixes #  

  * **Platform(s):** Windows and Linux.
  * **Unreal Engine version(s):** UE 5.3

#### Possible Drawbacks

- This change has to be added with the engine change, so that the ParallelLineTraceSingleByChannel(...) method does not give an error when compiling the ParallelLineTraceSingleByChannel(...) method.


Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastLidar.cpp
```diff
@@ -150,16 +150,34 @@ ARayCastLidar::FDetection ARayCastLidar::ComputeDetection(const FHitResult& HitI
...
     LidarData.ResetMemory(PointsPerChannel);
+    PointCloudResetMemory();
 
     for (auto idxChannel = 0u; idxChannel < Description.Channels; ++idxChannel) {
       for (auto& hit : RecordedHits[idxChannel]) {
         FDetection Detection = ComputeDetection(hit, SensorTransform);
         if (PostprocessDetection(Detection))
+        {
           LidarData.WritePointSync(Detection);
+          PointCloudWritePointSync(Detection);
```
Blyron, line 161: Would be nice if this operations is surrounded with an if condition and #if WITH_EDITOR

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastLidar.cpp
```diff
@@ -150,16 +150,34 @@ ARayCastLidar::FDetection ARayCastLidar::ComputeDetection(const FHitResult& HitI
       PointsPerChannel[idxChannel] = RecordedHits[idxChannel].size();
 
     LidarData.ResetMemory(PointsPerChannel);
+    PointCloudResetMemory();
```
Blyron, line 153: Would be nice if this operations is surrounded with an if condition and #if WITH_EDITOR

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastLidar.cpp
JoseM98: Fixed.

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Sensor/RayCastLidar.cpp
JoseM98: Fixed.

Blyron (approved):

PR #7267, opened 2024-03-19
Implement Environment Setup.sh

xavisolesoft:
#### Description

This script pretends to make easier the installation process and firts build for carla.

  * **Platform(s):** Ubuntu 20.04 & 22.04
  * **Python version(s):** 3.8 & 3.10
  * **Unreal Engine version(s):** 5.3

### DIFF
--- a/Setup.sh
+++ b/Setup.sh
@@ -1,18 +1,60 @@
 #!/bin/bash
+
+
+
+# ==================================================================================================
+# -- FUNCTIONS -------------------------------------------------------------------------------------
+# ==================================================================================================
+
+satisfies_minimum_version() {
+    CMAKE_VERSION="$(cmake --version | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+')"
+    CMAKE_MINIMUM_VERSION=$1
+    MAJOR="${CMAKE_VERSION%%.*}"
+    REMAINDER="${CMAKE_VERSION#*.}"
+    MINOR="${REMAINDER%.*}"
+    REVISION="${REMAINDER#*.}"
+    MINIMUM_MAJOR="${CMAKE_MINIMUM_VERSION%%.*}"
+    MINIMUM_REMAINDER="${CMAKE_MINIMUM_VERSION#*.}"
+    MINIMUM_MINOR="${MINIMUM_REMAINDER%.*}"
+
+    if [ $MAJOR -gt $MINIMUM_MAJOR ] ||
+        ( [ $MAJOR -eq $MINIMUM_MAJOR ] &&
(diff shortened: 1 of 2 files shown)


Setup.sh
```diff
@@ -0,0 +1,23 @@
...
+
+pip3 install --user numpy
+pip3 install --user -Iv setuptools==47.3.1
+pip3 install --user distro
+pip3 install --user wheel auditwheel
+
+git clone -b ue5-dev https://github.com/carla-simulator/carla.git carlaUE5
+cd carlaUE5
+
+curl -L -O https://github.com/Kitware/CMake/releases/download/v3.28.3/cmake-3.28.3-linux-x86_64.tar.gz
```
xavisolesoft, line 16: Think about a better way to ensure minimum cmake verison.

### Benefits of current implementation
- User no need to care about cmake version.
- User can change cmake version if desired.
- We are not changing global bash environment.

### Critics of this implementation
- cmake binary is non-git-versioned so `git clean -fdx` will remove it forcing to re-setup.

### Alternatives
- Upload cmake binary to carla repo.
- Add an external folder for cmake command like UnrealEngine5_carla one.

Setup.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -e
+
+y | sudo apt-add-repository "deb http://archive.ubuntu.com/ubuntu focal main universe"
+sudo apt-get update
+sudo apt-get install build-essential retry g++-7 ninja-build libvulkan1 python3 python3-dev python3-pip libpng-dev libtiff5-dev libjpeg-dev tzdata sed curl libtool rsync libxml2-dev git git-lfs
```
xavisolesoft, line 6: Check if we need the command `retry`.

Setup.sh
```diff
@@ -0,0 +1,23 @@
...
+
+pip3 install --user numpy
+pip3 install --user -Iv setuptools==47.3.1
+pip3 install --user distro
+pip3 install --user wheel auditwheel
+
```
xavisolesoft, line 12: Add commands to clone the Content.

Setup.sh
```diff
@@ -0,0 +1,23 @@
...
+git clone -b ue5-dev https://github.com/carla-simulator/carla.git carlaUE5
+cd carlaUE5
```
xavisolesoft, line 14: Add commands to clone and build Unreal5.

Setup.sh
xavisolesoft: Yes; we need it

Setup.sh
xavisolesoft: DOne.

Setup.sh
xavisolesoft: Done.

PR #7294, opened 2024-03-26
Add check for CarlaUnreal Content presence.

MarcelPiNacy:
This PR adds a check for whether CarlaUnreal Content is there.
It also changes the conditions under which a warning is issued due to the CMake build type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,9 +68,9 @@ include (${CARLA_WORKSPACE_PATH}/CMake/CarlaOptions.cmake)
 include (${CARLA_WORKSPACE_PATH}/CMake/Warnings.cmake)
 include (${CARLA_WORKSPACE_PATH}/CMake/Errors.cmake)
 
-if (NOT CMAKE_BUILD_TYPE STREQUAL "Release")
+if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   carla_warning (
-    "CARLA is not set to be built in Release mode. This may cause issues when building CarlaUnrealEditor."
+    "CARLA is set to be built in Debug mode. This may cause issues when building CarlaUnrealEditor."
   )
 endif ()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -68,9 +68,9 @@ include (${CARLA_WORKSPACE_PATH}/CMake/CarlaOptions.cmake)
 include (${CARLA_WORKSPACE_PATH}/CMake/Warnings.cmake)
 include (${CARLA_WORKSPACE_PATH}/CMake/Errors.cmake)
 
-if (NOT CMAKE_BUILD_TYPE STREQUAL "Release")
+if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   carla_warning (
-    "CARLA is not set to be built in Release mode. This may cause issues when building CarlaUnrealEditor."
+    "CARLA is set to be built in Debug mode. This may cause issues when building CarlaUnrealEditor."
```
xavisolesoft, line 73:
```suggestion
    "CARLA is not set to be built in Debug mode. This may cause issues when building CarlaUnrealEditor."
```

The message is inconsistent with the if check.

Unreal/CMakeLists.txt
```diff
@@ -55,6 +55,12 @@ set (
   ${CARLA_UNREAL_ENGINE_PATH}/Engine/Binaries/${UE_SYSTEM_NAME}/UnrealEditor${EXE_EXT}
 )
 
+if (IS_DIRECTORY ${CARLA_UE_PATH}/Content/Carla)
```
xavisolesoft, line 58:
```suggestion
if (IS_DIRECTORY ${CARLA_UE_PATH}/Content/Carla/Blueprints)
```

This will also check that `Content/Carla` is not empty

xavisolesoft, line 58: Makes sense to move this check to the `launch` target?

Content is not needed for build, is needed just for launch. Sometimes I would like to build CARLA without launching.

MarcelPiNacy: We could look into it, but the idea here is that if the variable BUILD_CARLA_UNREAL is set to ON, we expect content to be downloaded. Alternatively we could use a warning instead of an error.

xavisolesoft: On my understanding, is good to have this error for `launch` because with out this error is difficult for the users to know why Unreal Engine is chrashing when the content is missing.

But in the other hand, users should be able to build without content.

That is why I propose move this check to the `launch` target instead of CARLA build.

Do you think it will be difficult to move this check to the `launch` target?

xavisolesoft (approved):

PR #7300, opened 2024-03-27
Remove boost::numpy

MarcelPiNacy:
This PR removes boost numpy, as it is deprecated.
See https://github.com/ndarray/Boost.NumPy

### DIFF
--- a/Unreal/CMakeLists.txt
+++ b/Unreal/CMakeLists.txt
@@ -71,7 +71,6 @@ set (
   libsqlite3
   Boost::asio
   Boost::python
-  # Boost::numpy
   Boost::geometry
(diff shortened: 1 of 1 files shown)


Unreal/CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ set (
   libsqlite3
   Boost::asio
   Boost::python
-  Boost::numpy
+  # Boost::numpy
```
xavisolesoft, line 74:
```suggestion
```

I suggest to delete instead comment, for havving a clearner code base.

xavisolesoft (approved):

Issue #7341, opened 2024-04-09
[Python agents] `set_destination` has unused parameter and `_tailgating` bugged


Daraan:
Carla 0.9.15 and before:

### The problem:

The `set_destination` function currently is illogical as it only uses the `start_location` for boolean evaluation and not as a location anymore.

https://github.com/carla-simulator/carla/blob/cead6d1b05e89ad84df80984f62daba57bb50e55/PythonAPI/carla/agents/navigation/basic_agent.py#L141
https://github.com/carla-simulator/carla/blob/cead6d1b05e89ad84df80984f62daba57bb50e55/PythonAPI/carla/agents/navigation/basic_agent.py#L151-L156

### Manifestation:

The only native example where this is used is in the BehaviorAgent tailgating evasion maneuver:

https://github.com/carla-simulator/carla/blob/cead6d1b05e89ad84df80984f62daba57bb50e55/PythonAPI/carla/agents/navigation/behavior_agent.py#L119-L120

In my experiments this is broken as the agent makes a U-Turn to reach a weaypoint that is far behind the current location.

Daraan:
I think this is partially fixed with #7342, the BehaviorAgent will now not longer append the current location to the queue, to which it tries to come back later.

However as the queue is cleaned there is no plan beyond this. This might be similar to the original behavior when `set_destination` and `_tailgating` where still in sync but I have not researched this.

PR #7342, opened 2024-04-09
Agent's `set_destination` needs update in their logic and fix broken tailgating, but how?

Daraan:
#### Description

Addresses #7341:   
[`set_destination`](https://github.com/carla-simulator/carla/blob/cead6d1b05e89ad84df80984f62daba57bb50e55/PythonAPI/carla/agents/navigation/basic_agent.py#L141) does not really use the `start_location`, only for boolean evaluation instead of allowing users to pass a location.

In my experiments this sometimes creates troubles when passing a starting point, e.g. for [tailgating](https://github.com/carla-simulator/carla/blob/cead6d1b05e89ad84df80984f62daba57bb50e55/PythonAPI/carla/agents/navigation/behavior_agent.py#L119-L120) the agent performs U-Turns to an old way point.

Note: Python code only

  * **Python version(s):** 3.7, 3.10

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7342)

### DIFF
--- a/PythonAPI/carla/agents/navigation/basic_agent.py
+++ b/PythonAPI/carla/agents/navigation/basic_agent.py
@@ -139,30 +139,37 @@ class BasicAgent(object):
     def get_global_planner(self):
         """Get method for protected member local planner"""
         return self._global_planner
-
-    def set_destination(self, end_location, start_location=None):
+        
+    def set_destination(self, end_location, start_location=None, clean_queue=True):
+        # type: (carla.Location, carla.Location | None, bool) -> None
         """
         This method creates a list of waypoints between a starting and ending location,
         based on the route returned by the global router, and adds it to the local planner.
-        If no starting location is passed, the vehicle local planner's target location is chosen,
-        which corresponds (by default), to a location about 5 meters in front of the vehicle.
+        If no starting location is passed and `clean_queue` is True, the vehicle local planner's 
+        target location is chosen, which corresponds (by default), to a location about 5 meters 
+        in front of the vehicle.
+        If `clean_queue` is False the newly planned route will be appended to the current route.
 
             :param end_location (carla.Location): final location of the route
             :param start_location (carla.Location): starting location of the route
+            :param clean_queue (bool): Whether to clear or append to the currently planned route
         """
         if not start_location:
-            start_location = self._local_planner.target_waypoint.transform.location
-            clean_queue = True
-        else:
-            start_location = self._vehicle.get_location()
-            clean_queue = False
-
+            if clean_queue and self._local_planner.target_waypoint:
+                # Plan from the waypoint in front of the vehicle onwards
+                start_location = self._local_planner.target_waypoint.transform.location 
+            elif not clean_queue and self._local_planner._waypoints_queue:
+                # Append to the current plan
+                start_location = self._local_planner._waypoints_queue[-1][0].transform.location
+            else:
+                # no target_waypoint or _waypoints_queue empty, use vehicle location
+                start_location = self._vehicle.get_location() 
         start_waypoint = self._map.get_waypoint(start_location)
         end_waypoint = self._map.get_waypoint(end_location)
-
+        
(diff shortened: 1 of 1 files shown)


Daraan:
@glopezdiest could you give it a look when you have the chance and share your thoughts. Thanks a lot.

Daraan:
Thank you for your reply.

As summary, we have these cases and `end_location` is always `not None`:

1)
> The first one is the more **generic one**. Both `start_location` and `end_location` are given, because a specific path is desired. In this case, the function should clean the previous queue, and just add the new route. In this case, the `start_location` can make the controller fail if it is a weird position, but that is acceptable and the fault relies on the inputs, not the function itself.

* `start_location` is `not None`
* `clear_queue=True`

2)
> The second one is the **rerouting one**, where you just want to, for example, stop moving randomly, and start moving in a specific way.

* `start_location` is `None`
*  `clear_queue=True`

3)
> The third one is the **non-interference one**, where you don't want to change the current behavior, and just add to it once it has finished.

* `start_location` is `None`
*  `clear_queue=False`

Theoretically this gives a fourth case similar to 3) with the problems of 1) and up to the input.
* `start_location` is `not None`
*  `clear_queue=False`

---

I'll look again over it in the coming days/week

glopezdiest:
That breakdown looks good to me
> Theoretically this gives a fourth case similar to 3) with the problems of 1) and up to the input.
>    start_location is not None
>    clear_queue=False
And yes, that's true, but I don't see an issue with that. It is a weird use-case from my point of view, but maybe someone can find a reasonable on later on

glopezdiest:
@Daraan Would you be interested in changing the `set_destination` function as discussed? I think this would be a good fix

Daraan:
Yes, I am just held up with something else currently.
I'll look again in the next week (s)

Am 31.05.2024 13:13 schrieb "glopezdiest - ***@***.***" ***@***.***>:

External Sender: Achtung bei Links & Anhängen innerhalb dieser E-Mail // Careful with links & attachments inside this email!



@Daraan<https://github.com/Daraan> Would you be interested in changing the set_destination function as discussed? I think this would be a good fix

Daraan:
@glopezdiest 

I've finished and updated the rework. Comments are currently left in for you and can/should be cleaned.

Case 2 `start_location=None, clean_queue=True`:
* Default A same behavior: However, it also checks if the `local_planner.target_waypoint` is set. This is currently always the case, but I think it could make sense in the future to set it to None in `LocalPlanner.run_step` when the queue is empty instead of keeping the last waypoint that was targeted -> fallback to vehicle location if it is not available. 

Case 1 :  `start_location not None, clean_queue=True`:
* Default B - New behavior : `start_location` is actually used instead of disregarding it for the vehicle location. For me this fixes the BehaviorAgents buggy tailgaiting behavior of randomly turning around because the queue, was not cleaned (#7341 )

Case 3:  `start_location=None, clean_queue=False`:
* If the queue is not empty takes the *last* waypoint from the local planner's waypoint queue as starting location so that the new route is appended. In my opinion it makes no sense to plan from the current location, right?. If empty -> fallback to vehicle location.  
NOTE: This last waypoint will then be twice in the queue. Do you think this could cause a problem? Could also pop it at that location instead.

Case 4:  `start_location not None, clean_queue=False`:
* New theoretical option as discussed above.

glopezdiest:
Looks good, if you remove the comments we can merge it

Daraan:
> Looks good, if you remove the comments we can merge it
> 

Take a look again. Are you find with the cleaned comments and type annotation or should I fully remove them?

glopezdiest:
The small comment is fine, as a further clarification. I'll merge it ASAP

glopezdiest (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7342)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

PR #7400, opened 2024-04-11
Fix windows package generation

xavisolesoft:

  * **Platform(s):** Windows 10, Ubuntu 22.04
  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** 5.3


xavisolesoft:
After this concept test, we determinet that the approach followed by this PR does not work always due the undetermenistic execution order of the cmake targets.

We implemented a new approach to solve the same issue without flacky bugs:
https://github.com/carla-simulator/carla/pull/7427

PR #7450, opened 2024-04-16
Removing HoudiniEngine plugin and main references

JoseM98:
#### Description
- It is necessary to get latest Carla Content 
- Removed Houdini from the Carla build system to avoid downloading its plugin.
- Removed Houdini modules from CarlaTools.build.cs.

Fixes #  

  * **Platform(s):** Windows 10 and Ubuntu 22.04
  * **Python version(s):** 3.10.12
  * **Unreal Engine version(s):** UE4.26 Carla


Unreal/CarlaUE4/Plugins/CarlaTools/Source/CarlaTools/Private/HoudiniImportNodeWrapper.cpp
```diff
@@ -5,7 +5,6 @@
 // For a copy, see <https://opensource.org/licenses/MIT>.
 
 #include "HoudiniImportNodeWrapper.h"
```
xavisolesoft, line 7: Do we have plans to remove `HoudiniImportNodeWrapper` and `HoudiniImporterWidget`.
On my understanding, does not make sense to keep this classes in the code base after removing HoudiniEngine.
There is any reason to keep this classes?

Blyron: This classes are still referenced in Digital Twins Tool, names are unproper due to those are used for more stuff than Houdni

xavisolesoft (approved):

Issue #6632, opened 2023-07-13
Setting up carla with unreal engine 4.26 on Linux, the git clone link for unreal engine not working. it requires username for github which :Support for password authentication was removed on August 13, 2021. And when I try to use the ssh link to clone it says: ERROR: Repository not found. fatal: Could not read from remote repository.  Please make sure you have the correct access rights and the repository exists.


Bo-Yu-Columbia:

skilh:
I have solved this issue with a personal access token. You can generate it and then use it as a password in this case.

siddharthjain38:
@Bo-Yu-Columbia Have you linked github account with unreal engine account. use - https://www.unrealengine.com/en-US/ue-on-github

Bo-Yu-Columbia:
yes i have

dscherrenburg:
I was running into the same issue just a minute ago. I got the same error messages both for following the tutorial and using my ssh-key. 
However, it is possible to clone unreal engine using your ssh-key. The mistake I made that lead to the same error (Please make sure you have the correct access rights and the repository exists.) was pressing enter to the following question: "Are you sure you want to continue connecting (yes/no/[fingerprint])?". After following the instructions by typing 'yes', the package cloned perfectly without any issues. 
So it is possible that you have made the same mistake and that the workaround would be just typing 'yes' to the question.

l-yerim23:
hi my name is yerim lee.
I like d

PR #7504, opened 2024-04-25
Use Github Access Token instead of password in build_docker_unreal.md

KianBahasadri:
Change docs from using Github password to use Github Access Token

#### Description
The docs for building Unreal from docker uses a password for authentication
```EPIC_PASS=<GitHubPassword>```
This is outdated, changing it to 
```EPIC_PASS=<GitHubAccessToken>```

Mentioned in #6632


Blyron:
Hello! All pull requests must be against dev please

PR #7609, opened 2024-05-07
Introduce several helper macros and utilities for improved usability.

MarcelPiNacy-CVC:
This PR introduces the following helper macros:
 - carla_add_executable
 - carla_add_library
 - carla_add_custom_target
These all wrap the corresponding CMake commands, with the goal being to add extra documentation code in a similar fashion than seen in carla_option and carla_string_option.
This PR also adds two new options to print all relevant targets and options at configure time and to save it to /Help.md.

### DIFF
--- a/Unreal/CMakeLists.txt
+++ b/Unreal/CMakeLists.txt
@@ -8,7 +8,8 @@ project (
     "Open-source simulator for autonomous driving research."
 )
 
-option(ENABLE_DIRECTORY_CLEAN
+carla_option (
+  ENABLE_DIRECTORY_CLEAN
   "Enable ADDITIONAL_CLEAN_FILES on directory level"
   ON
 )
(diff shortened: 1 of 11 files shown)


Unreal/CMakeLists.txt
```diff
@@ -337,9 +340,11 @@ add_dependencies (
   ${UE_DEPENDENCIES_ORDER_ONLY}
 )
 
-function(add_carla_ue_package_target TARGET_NAME_SUFFIX UE_BUILD_CONFIGURATION)
-  add_custom_target (
+function (add_carla_ue_package_target PACKAGE_CONFIGURATION UE_BUILD_CONFIGURATION)
+  set (TARGET_NAME_SUFFIX -${PACKAGE_CONFIGURATION})
```
xavisolesoft, line 344: In the case that `PACKAGE_CONFIGURATION` is empty, target suffix will be `-` and this will generate the target `pacakge-` instead `package`.

We should generate and empty suffix when `PACKAGE_CONFIGURATION` is empty to maintain the target name `package`.

xavisolesoft (approved):

PR #7639, opened 2024-05-13
Add functions and macros to aid in documentation and general usability.

MarcelPiNacy-CVC:
This PR adds several macros to aid in documentation and usability.


CMake/Util.cmake
```diff
@@ -17,26 +17,104 @@ endfunction ()
...
+    "- ${NAME}\n"
+    "\t- Description: ${DESCRIPTION}\n"
+    "\t- Default: \"${VALUE}\"\n"
+  )
+  set_property (GLOBAL PROPERTY CARLA_OPTION_DOCS ${DOCS})
 endmacro ()
+
+
+
+function (carla_get_target_docs OUT_VAR)
```
xavisolesoft, line 72: It looks good.

Just I want to add that Cmake ExternalProjects are also considered targets but are not documented. So we have Easter egg targets like:
```
cmake --build Build --target fastdds
```

Or this rare one added by Eigen:
```
cmake --build Build --target blas
```

What do you think about it? Makes sense to add also ExternalProject to the target doc list?
Probably, ExternalProject targets would add a lot of noise as there is no way at cmake to make targets private or not accessible.

xavisolesoft (approved):

PR #7893, opened 2024-07-02
Added C++ client to be compiled with other targets

Blyron:
Add support to compile C++ Example using default cmake process

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7893)


Examples/CMakeLists.txt
```diff
@@ -1 +1,73 @@
-add_subdirectory (CppClient)
+set (
+  CARLA_EXAMPLE_CPP_CLIENT_SOURCE_PATH
+  ${CARLA_WORKSPACE_PATH}/Examples/CppClient
+)
+
+set (
```
MarcelPiNacy-CVC, line 6: Do we know if only passing carla-client as a dependency is enough? In theory carla-client should already be including all of these.

Examples/CMakeLists.txt
```diff
@@ -1 +1,73 @@
...
+  Boost::asio
+  Boost::python
+  Boost::geometry
+  Boost::gil
+  RecastNavigation::Recast
+  RecastNavigation::Detour
+  RecastNavigation::DetourCrowd
+)
+
+if (BUILD_EXAMPLES)
```
MarcelPiNacy-CVC, line 21: This if can be removed: we check for BUILD_EXAMPLES being on in the top CMakeLists.txt file, in which case all subprojects in Examples/ will be skipped.

Examples/CMakeLists.txt
```diff
@@ -1 +1,73 @@
...
+    ${CARLA_EXAMPLE_CPP_CLIENT_SOURCE_PATH}/*.h
+  )
+  
+  carla_add_executable (
+    carla-example-client
+    "Build the CARLA C++ client example."
+    ${CARLA_EXAMPLE_CPP_SOURCES}
+  )
+    
+  add_dependencies (
```
MarcelPiNacy-CVC, line 52: This add_dependencies is not needed, as target_link_libraries already implicitly adds the dependency.

Examples/CMakeLists.txt
```diff
@@ -1 +1,73 @@
...
+    LANGUAGES
+      CXX
+    VERSION
+      ${CARLA_VERSION}
+  )
+
+  file (
+    GLOB_RECURSE 
+    CARLA_EXAMPLE_CPP_SOURCES
+    CONFIGURE_DEPENDS
```
MarcelPiNacy-CVC, line 35: Is CONFIGURE_DEPENDS needed here?

Examples/CMakeLists.txt
```diff
@@ -1 +1,73 @@
...
+  )
+
+  file (
+    GLOB_RECURSE 
+    CARLA_EXAMPLE_CPP_SOURCES
+    CONFIGURE_DEPENDS 
+    ${CARLA_EXAMPLE_CPP_CLIENT_SOURCE_PATH}/*.cpp
+  )
+
+  file (
```
MarcelPiNacy-CVC, line 39: CARLA_EXAMPLE_CPP_HEADERS is not referenced in carla_add_executable.

Examples/CMakeLists.txt
MarcelPiNacy-CVC: Nevermind, this is indeed correct. We should update other source-related GLOB calls in CARLA.

Examples/CMakeLists.txt
Blyron: updated

Examples/CMakeLists.txt
Blyron: Updated

Examples/CMakeLists.txt
Blyron: Okay

Examples/CMakeLists.txt
Blyron: Updated

MarcelPiNacy-CVC (approved):

PR #7919, opened 2024-07-08
Cleanup and harmonize Carla build

berndgassmann:
Ensure that actually only the unreal sysroot is deployed for all C as well C++ compilations including dependencies. That ensures the server is NOT anymore linked against the system glibc (which is in Ubuntu22.04 incompatible with the Unreal one). As a consequence no local clang installation required anymore to build under Linux.

Harmonized naming split into client/server (instead of libcpp and libstdcpp) to ensure nothing mixed up (and there were things mixed up before!)

Refactor C# Unreal build using CarlaRules base class to provide common functionality. Fixing windows build without ROS2 and added intitial windows build with ROS2

Restrict to one DDS-Domain-Participant for carla-server (Fixes bind address already in use exception on with ROS2).

Due to continued segfaults in reallocations of MallocBinned2 in conjunction with ROS2 enforce using AnsiMalloc calls under Linux (see https://forums.unrealengine.com/t/dealing-with-allocator-mismatches-with-external-libraries/1416830)

Use C++20 to allow for more robust struct initialization (see https://en.cppreference.com/w/cpp/language/aggregate_initialization), added Eigen C++20 patches and required CARLA changes to cope with C++20 compilation errors
Be aware: This change requires the Unreal PR
(https://github.com/CarlaUnreal/UnrealEngine/pull/23)

Move forward to g++-13

Minors:
- Fix make clean call
- Update osm2odr commit with build fix (https://github.com/carla-simulator/sumo/pull/6)
- Update ad-rss to 4.4.5 (supports boost 1.80)
- Fix windows install scripts and forward arguments for building libcarla
- Adapt codeformat.py to python3 using clang-format-14

#### Description

Fixes #  

  * **Platform(s):** ...
  Ubuntu 22.04, 20.04, compilation on Windows (some time ago on the other PR #7229)
  * **Python version(s):** ... 3.10


berndgassmann:
Finally replaced by #9478

PR #7982, opened 2024-07-24
Documentation fixes

Daraan:
#### Description

Included in this PR are

- Rolled back two changes from #7763:
  1. Readded "identifiers" in this docstring, as the sentence is incomplete currently.
        ```diff
        - Defines the of each GBuffer texture (See the method 
        + Defines the identifiers of each GBuffer texture (See the method 
        ```
  2. Removed the readded reference to the removed `carla.Sensor.is_listening` property. See #7439

- fixes of spelling mistakes (code and documentation)
- Corrected wrong literal wrapping instead of linking in
  > [old:] Defines the of each GBuffer texture (See the method `[carla.Sensor.listen_to_gbuffer](#carla.Sensor.listen_to_gbuffer)`).  
- Added space after # Heading when docs are generated `#Python API references` -> `# Python API references` to be correct markdown. **I am not sure if this is intentionally or an old oversight**, but haven't seen anything where this could be relevant

#### Where has this been tested?

  * **Platform(s):** Ubuntu
  * **Python version(s):** 3.7 & 3.10 (should be irrelevant)
  * **Unreal Engine version(s):** 4.26

#### Note:

Currently pending changes to snippets from `make PythonAPI.docs` not included

PR #7999, opened 2024-07-31
Added inverse transform

glopezdiest:
#### Description

Exposed the C++ InverseTransformPoint() into the API. Docs are missing

  * **Python version(s):** 3.8
  * **Unreal Engine version(s):** CARLA

#### Possible Drawbacks

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/7999)

### DIFF
--- a/PythonAPI/carla/source/libcarla/Geom.cpp
+++ b/PythonAPI/carla/source/libcarla/Geom.cpp
@@ -237,6 +237,10 @@ void export_geom() {
       self.TransformPoint(location);
       return location;
     }, arg("in_point"))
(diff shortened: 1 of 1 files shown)


Blyron (approved):

Daraan:
Added docs in #7982

```
    - def_name: inverse_transform
      params:
      - param_name: in_point
        type: carla.Vector3D
        doc: >
          Location in the space to which the inverse transformation will be applied.
      doc: >
        Applies the inverse of `transform` by translating a 3D point in place from global to local coordinates using the current transformation as frame of reference.  
```

Maybe you can comment whether you think its fitting.

PR #8072, opened 2024-08-23
Typing and modernisation of Agent code

Daraan:
#### Description

This PR adds:
- TypeComments for the GlobalRoutePlanner, and barried code by `TYPE_CHECKING` granting good type coverage for pyright: basic
- Updated some <0.9.13 code to use `Vector3D` methods that came in 0.9.13, e.g. location.
  this deprecates `agents.tools.misc compute_distance and vector`.  
  Compatibility is granted with some slight improvements in performance by using a global constant for the `eps` value.
- Removed some deprecated/unused fragments, e.g. inheritance from object, unnecessary class parentheses.
- Some spelling mistakes also addressed in #7982 

  * **Python version(s):** 3.7 & 3.10

#### Possible Drawbacks

- This uses `location.distance(location2)` in favor of calculating the distance over `numpy`, the results are `np.close` but not identical. I do not see an issue here.
- This changes `class BasicAgent(object): -> class BasicAgent:` which results in slightly different type behavior in Python 2 only.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8072)


PythonAPI/carla/agents/navigation/behavior_agent.py
```diff
@@ -15,7 +15,7 @@
 from agents.navigation.local_planner import RoadOption
 from agents.navigation.behavior_types import Cautious, Aggressive, Normal
 
-from agents.tools.misc import get_speed, positive, is_within_distance, compute_distance
```
Daraan, line 18: functions were unused

Blyron:
@Daraan can you point to target ue4-dev?

Daraan:
> @Daraan can you point to target ue4-dev?

Yes and done.

Blyron (approved):

PR #8117, opened 2024-09-09
Fix invalid comparisions with string literals in RSS example

Daraan:
#### Description

This PR addresses some string comparisons in the *rss_visualization.py* that are written like:  
`ad.rss.state.RssStateEvaluator.LongitudinalDistanceOppositeDirection == "LongitudinalDistanceOppositeDirection"` which is always `False` as the enum is an int and has no support for string comparison.

In the example some arrows on the HUD were therefore always or never printed.

---

For verification try this statement which **returns `False` but is intended to be `True`**.

```python
from carla import ad
print(ad.rss.state.RssStateEvaluator.LongitudinalDistanceOppositeDirection == "LongitudinalDistanceOppositeDirection")
# False
```

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8117)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
  * Importing from carla.command is now possible
  * carla.ad subpackages are now directly importable and are not directly importable anymore (e.g. import ad)
  * Fixed segfault in traffic manager when trying to access not available vehicles
+ * Fixed invalid comparission in python examples/rss
 
 ## CARLA 0.9.15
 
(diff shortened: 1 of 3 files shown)


Daraan:
Where did the rss/examples go in the new branches?

Daraan:
@PabloVD  rebased to dev branch

PabloVD:
Merged, thanks!

PR #8125, opened 2024-09-11
Instance segmentation

AreopagX:
#### Description

Previously, the instance segmentation camera only allows displaying the internal ids from the Unreal Engine. This pr allows to display the actor ids when they are available, otherwise falling back to ids from UE.

This pr builds on the #5452 , which makes introduces the option to switch between only internal or only actor ids for the instance segmentation. However, it was found that this results in static actors (e.g., parked vehicles) missing in the segmentation images, since they just do not have an actor id. 

Furthermore, in order to allow for mapping ids from the segmentation camera to bounding boxes, this pr adds the actor id to the bounding boxes retrieved from the "get_all_lvl_bbs()" call. That way, bounding boxes for static actors can also be matched with their pixels in the segmentation images.

#### Demo

The images show the instance segmentation with static and dynamic actors. The first two channels contain the id, the third channel contains the class label and the last channel is unused. 
In the first image, the mouse points at the parked vehicle, yielding a pixel value of (159, 49, 14, 255). This means that the actor id of this vehicle is $159*256 + 49=40753$. The second image highlights the id of the rider which is 24.
Parked Vehicle (Static Actor)
Actor

  * **Platform(s):** Ubuntu 23.04
  * **Python version(s):** 3.8.18

#### Possible Drawbacks

- The instance segmentation still only covers the last two bytes of the UE actor ids. Therefore, id collisions could happen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8125)


Blyron:
We will review this as soon as we can

Blyron:
Could you change the base branch to ue4-dev, please?

AreopagX:
@Blyron I reopened this pr with ue4-dev as base branch in #8334 . I didn't find an option to change the base branch without opening a new pr.

PR #8129, opened 2024-09-12
Adding sensor lightning values and adding render variables to fix vsm…

Blyron:
Fix for virtual shadow maps and change default values to camera sensors to look as expected

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8129)


LuisPovedaCano (approved):
Nice

PR #8130, opened 2024-09-12
Porting the changes done to UE5 to fix the recording leak in UE4

jorge-kabuto:
The slowdown is considerably more noticeable here since the engine runs much smoother. This makes evident that this is a stopgap measure, and should be looked into further down the line.

#### Description

Removed parallelization of the Write functions in ServerSession and Client by stopping the use of boost post, which led to desync between client and server, and ultimately caused the leak.

Fixes #  

  * **Platform(s):** Ubuntu 22
  * **Python version(s):** ...
  * **Unreal Engine version(s):** Unreal Engine 4

#### Possible Drawbacks

Noticeable slowdown while recording with the Python API. On my tests, both the client and the server slowed to about 6 fps while saving to disk in sync mode.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8130)

### DIFF
--- a/LibCarla/source/carla/streaming/detail/tcp/Client.cpp
+++ b/LibCarla/source/carla/streaming/detail/tcp/Client.cpp
@@ -15,7 +15,6 @@
 #include <boost/asio/connect.hpp>
 #include <boost/asio/read.hpp>
 #include <boost/asio/write.hpp>
-#include <boost/asio/post.hpp>
 #include <boost/asio/bind_executor.hpp>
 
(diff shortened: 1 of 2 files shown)


Blyron (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8130)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jorge-kabuto)

PR #8152, opened 2024-09-17
Hack to fix ad_rss's import namespaces

### LINKED ISSUE #8093: `libcarla.py` messy import of `ad` packages as top-level modules.
CARLA version: 0.9.15 dev (but not latest); using .egg file.
Platform/OS: Ubuntu 22.04

### Problem you have experienced:

I realized that the integration of the `ad` subpackage is messy: Recusively all packages are added as top-level packages to `sys.modules` after the call to
`spec = importlib.util.spec_from_file_location(__name__,__file__)` in `libcarla.py`

**Does not work; but should**
```
import carla
import carla.ad
```

**Does work**
```
import carla
from carla import ad
carla.ad
# Below should not work
# all the subpackages of ad can be imported directly after importing carla
import ad
import map
import rss
...
```

I do not know if this is an issue of [ad-rss-lib](https://github.com/intel/ad-rss-lib) or from CARLA.

Daraan:
#### Description

Fixes #8093

This is a hack that fixes the messy import namespacing of the `ad_rss` library. Importing `carla` which in turns imports `ad_rss as ad` puts all its subpackages as top-level pckages/modules.

For example this is currently valid:

```py
import carla
import ad
import core # ad.rss.core
import world 
...
```

This raises an ImportError

```py
import carla.ad
```

Also it creates ambiguous import behavior where the order of imports matter, e.g. if I had a subpackage/module names like *any* of ad_rss subpackages the following code will also fail as it imports the wrong module

```py
# import world # <-- would import my local world.py
import carla  # assume with RSS
import world, core  # This imports ad subpackages instead of mine.

```

This is non-standard and confusing behavior.

See also: https://github.com/intel/ad-rss-lib/issues/258

---

**Note: This issue should actually be addressed by ad_rss lib or somewhere in the packaging process before creating the `.so` file**

**This PR manually adjust `sys.modules["ad..."]` to `sys.modules["carla.ad..."]` so that correct and normal import behavior is possible.**

---

#### Possible Drawbacks

- Code that relies on the strange import behavior needs updates.
- When the ad_rss requirement is updates the adjustments need to be checked again. I checked that it aligns with the currently latest master branch of https://github.com/intel/ad-rss-lib (4.5.3)
- This is only a hack.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8152)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
  * Synchronized actor BoundingBox between server and client
  * Add actor_id to bounding boxes
  * Fixed invisible terrain in instance segmentation
+ * carla.ad subpackages are now directly importable and are not directly importable anymore (e.g. import ad)
 
 ## CARLA 0.9.15
 
--- a/PythonAPI/carla/source/carla/__init__.py
+++ b/PythonAPI/carla/source/carla/__init__.py
@@ -6,3 +6,59 @@
 
 # pylint: disable=W0401,import-self
 from .libcarla import *
+
+def __fix_ad_namespaces():
+    """
+    When using PythonAPI.rss the ad_rss subpackage is imported, however
+    this package puts all its subpackages into sys.modules.
+    This creates ambiguous modules and a wrong hierarchy.
+    
+    This is a # HACK to fix this issue.
+    
+    When updating the RSS version, check if this is correct or ad_rss has fixed
+    the import behavior.
+    """
+    import sys
+    import inspect
+    ad_submodules = {}  # dict of modules
+    def get_submodules(mod, parent):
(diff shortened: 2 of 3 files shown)


PabloVD:
Thanks for the PR. Could you please modify the PR to the ue4-dev branch rather than the dev branch? We are no longer maintaining that one.

PabloVD (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8152)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Daraan)

PabloVD:
Thanks for PR, merged.

PR #8161, opened 2024-09-18
Correct "libcarla.command" namespace and allow `from carla.command import ...`

### LINKED ISSUE #6414: CARLA 0.9.14 - PythonAPI - ModuleNotFoundError: No module named 'carla.libcarla.command'
Hi,

While writing our own setup script to spawn cars and pedestrians in the simulator, we encountered this issue when trying to import the `SpawnActor`, or any module under `carla.command`

Our setup script is based on the `/carla/PythonAPI/examples/generate_traffic.py` one.

Any suggestion?

CARLA version:  0.9.14
Platform/OS: Ubuntu 20.04
Problem you have experienced: ModuleNotFoundError: No module named 'carla.libcarla.command'
What you expected to happen: to import the module
Steps to reproduce: 

Install carla python from pypi: and try to import a command

```
pip3 install carla==0.9.14
python3
> from carla.command import SpawnActor
```

Daraan:
Fixes  #6414  

Currently `import carla.command` or `from carla.command import ...` is not possible.

This will raise:
```python
# command.py
      7 # pylint: disable=W0401
----> 8 from .libcarla.command import *

ModuleNotFoundError: No module named 'carla.libcarla.command'
```

This is because the `command` module is registered as `libcarla.command` in `sys.modules`. This is not really correct.

---

This PR adjusts this:
- renames the module to `carla.libcarla.command` 
- provides a duplicate `carla.command` that can be imported.

Note / Up to discussion:
This PR deletes `command.py` as otherwise `carla.command is not carla.libcarla.command`. Both modules would be equivalent but not the same. I think this is the cleaner option.

### Alternatives

This PR:
- PRO: Compatibility is granted and extended to be more user friendly
- CON: Duplicate module in sys.modules

1) Instead of putting `command` into `libcarla` we could put it into `carla` directly without creating a duplicate `carla.libcarla.command`.
   - PRO: More user friendly, no duplication of the module in `sys.modules`.
   - CON: Not part of `libcarla`
2) Do not allow `import carla.command` and only `from carla import command`
   - PRO: No duplication in `sys.modules`
   - CON: Less user friendly

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8161)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
  * Synchronized actor BoundingBox between server and client
  * Add actor_id to bounding boxes
  * Fixed invisible terrain in instance segmentation
+ * Importing from carla.command is now possible
 
 ## CARLA 0.9.15
 
--- a/PythonAPI/carla/source/carla/__init__.py
+++ b/PythonAPI/carla/source/carla/__init__.py
@@ -6,3 +6,8 @@
 
 # pylint: disable=W0401,import-self
 from .libcarla import *
+
+# Allow from carla.command import ...
+import sys
+sys.modules["carla.command"] = command
+del sys
--- a/PythonAPI/carla/source/carla/command.py
+++ b/PythonAPI/carla/source/carla/command.py
@@ -1,8 +1,0 @@
-# Copyright (c) 2019 Computer Vision Center (CVC) at the Universitat Autonoma de
-# Barcelona (UAB).
-#
-# This work is licensed under the terms of the MIT license.
-# For a copy, see <https://opensource.org/licenses/MIT>.
-
-# pylint: disable=W0401
-from .libcarla.command import *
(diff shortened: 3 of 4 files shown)


PabloVD:
Thanks for the PR. Could you please modify the PR to the ue4-dev branch rather than the dev branch? We are no longer maintaining that one.

Daraan:
Rebased and changed base

PabloVD:
Tested and works fine, this will allow to write more user friendly code, so we are merging it. Thanks for PR!

qhaas:
Any plans to integrate this into ue5-dev?  Per commit 4a6622d15dc4f1c5247e470882dc639743e17e14
`AttributeError: module 'carla' has no attribute 'SpawnActor'`

PR #8340, opened 2024-11-05
Aaron/programaticvehiclefactory

Blyron:
Make Vehicle factory programatic. 
Vehicles are written into json file and the file is read when simulation starts and the vehiclefactory is populated


LuisPovedaCano (approved):
Good Job!

PR #8381, opened 2024-11-18
Fix Windows build errors.

MarcelPiNacy-CVC:
This PR fixes multiple build errors on Windows systems.
This is done by introducing header guard files for the engine. That is, whenever you include a UE header, it must be surrounded by  ue-header-guard-begin/end.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8381)


ali-0-1:
Later, modified according to suggestion at #8893 

Tag issue seems resolved but,

This time got syntax error which seems version issue of compiler but our compiler version is recent.

various errors like:

ROS2CallbackData.h(59): error C2061: syntax error: identifier 'variant' is not a member of std

PR #8401, opened 2024-11-22
Compile ROS2 dependencies using Unreal's toolchain

joel-mb:
#### Description

ROS2 dependencies are now compiled using Unreal's toolchain.

  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** 5.5

#### Possible Drawbacks

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8401)


MarcelPiNacy-CVC (approved):
Looks good. At some point we should rewrite Carla.Build.cs to get the libraries paths using CMake instead of hardcoding the library names, which is pretty fragile.

PR #8436, opened 2024-12-03
Update to new version of boost

PabloVD:
Update from 1.84.0 to 1.86.0


KieranRatcliffeInvertedAI:
Hello I am a developer from Inverted AI and one of our dependencies is in the Carla package. According to conversation with @PabloVD, when this `boost` dependency is upgraded, it will allow for the `numpy` version to be upgraded as well since `boost` 1.87 is compatible with `numpy` 2.x. This will allow us to upgrade our version of the `numpy` dependency as well. That is a long path to get to my request which is that we would like to do this upgrade on our side as soon as possible so please let me know the status of testing this PR and when it might be merged and the `numpy` version can be incremented as well.

Blyron:
We need to check before that if upgrading the boost version hit us in some way.

MarcelPiNacy-CVC:
Currently this PR is _very_ broken: more work needs to go into this (and this only applies to the UE5 branch). For example, if you open the UE editor and attempt to run a CARLA python script, it seems to get stuck.

MarcelPiNacy-CVC:
Also, we could attempt to switch directly to boost 1.88 since we're at it.

Issue #8510, opened 2024-12-27
The output of RGB camera and instance segmentation camera in carla0.9.14 is different


DarkerSun6:
Hello, I have placed an RGB camera and an instance segmentation camera in the same location on the TOWN 01 map in version 0.9.14 of Carla.
There is a problem here, the mountains that can be seen in the RGB camera are missing in the instance segmentation camera.
And the vehicles hidden behind the mountain cannot be seen in the RGB camera, but are seen in the instance segmentation camera.
May I ask what the problem is
![2 0_3D_box](https://github.com/user-attachments/assets/f4c2e678-bc46-4b32-aae8-ee20231c97ba)
![2 0](https://github.com/user-attachments/assets/5315caae-1afb-4b6b-b107-f8506def9f5b)

lynrookie:
The semantic segmentation camera renders elements in scene with a different color depending on how these have been tagged. The tags are created by the simulator depending on the path of the asset used for spawning. For example, meshes tagged as Pedestrians are spawned with content stored in "Unreal/CarlaUE4/Content/Static/Pedestrians".
So, back to your problem, the mountains that can be seen in the RGB camera are missing in the instance segmentation camera, 
And the vehicles hidden behind the mountain can not be seen in the RGB camera, but are seen in the instance segmentation camera. 
That is because the mountains asset is not tagged in right way, You may check it in the [CARLA documentation](https://carla.readthedocs.io/en/latest/tuto_G_retrieve_data/#semantic-segmentation-camera)

Blyron:
Right now, the CARLA UE4 version does not support landscapes

tdegraaff:
@jorge-kabuto Can now be closed (https://github.com/carla-simulator/carla/pull/8528).

Issue #6905, opened 2023-11-13
Instance Segmentation does not show terrain


yoakim82:
CARLA version: 0.9.14 (custom build with included custom objects)
At least Town01 and Town03 affected, probably all maps.

Platform/OS: Ubuntu 22.04 LTS

Problem you have experienced:

When creating snapshots of a scene with both depth, rgb and segmentation sensors, there is an unexpected behavior in the segmentation view of the scene. Note I'm referring to the instance segmentation sensor here.
We use the instance segmentation to capture tight bounding boxes for our objects of interest, but when an object is spawned behind a hill or piece of terrain, the segmentation view will show the object even though it is 100% occluded in the rgb view. This totally ruins the purpose of having semantic segmentation. 

![terrain_traffic_light_rgb](https://github.com/carla-simulator/carla/assets/6775811/04c2492d-0df2-4dbe-9b43-d0c44af1e634)
![terrain_traffic_light_inst_seg](https://github.com/carla-simulator/carla/assets/6775811/f156fd70-d16a-49f4-96a7-316862808e38)

What you expected to happen:
I expect the objects being occluded by the terrain to not show up in the segmentation view, which seem to be the case for other occluding objects (e.g. trees/vegetation).
To clarify, the bounding boxes shown in the rgb image are generated only from information in the instance segmentation image.

Steps to reproduce:
Setup a triplet of sensors and navigate to a place in the map like in our enclosed scene.

nikky4D:
did you have this solved?

yoakim82:
No, I did not. My guess is it's related to how the town is created and how the terrain is labeled semantically.

But I have no clue as to how to fix this. Anyone with deeper UE4 skills is welcome to comment on this.

Yongzhaooo:
I met the same situation and it is still not fixed now : (
![image](https://github.com/carla-simulator/carla/assets/133014490/a5852675-ca8e-4736-9dc7-f6756e094898)

AreopagX:
related: #5367

tdegraaff:
@jorge-kabuto Can now be closed (https://github.com/carla-simulator/carla/pull/8528).

Issue #5367, opened 2022-04-15
instance segmentaion bug in town01


zxiaomzxm:
CARLA version: 0.19.3
Platform/OS: ubuntu 20.04
Problem you have experienced: 
I use instance segmentation camera in **town01**, but capture strange results:
rgb image
![图片](https://user-images.githubusercontent.com/5425522/163555064-3b97daeb-d3b9-475b-9237-e33efc505a18.png)
instance image
![图片](https://user-images.githubusercontent.com/5425522/163555207-efaa83e2-9477-4df0-8424-ed685eb9a923.png)
It seems that the instance camera can not see the mound in the right side and ray casting just pass through it.
Other maps has no such phenomenon.

XGodina:
Hi @zxiaomzxm, You wrote "CARLA version: 0.19.3", do you mean the " CARLA version 0.9.13", right?. I will take a look. Thank you for letting us know

zxiaomzxm:
> Hi @zxiaomzxm, You wrote "CARLA version: 0.19.3", do you mean the " CARLA version 0.9.13", right?. I will take a look. Thank you for letting us know

yep, it's 0.9.13. And I found town04 have similar problem.
![000545](https://user-images.githubusercontent.com/5425522/171339306-a5a58fd3-4055-49a1-8721-e43d23eff0f3.jpg)
![000545](https://user-images.githubusercontent.com/5425522/171339317-fde69ea0-6b7d-4995-907e-62f9272a52de.png)

ghh116553:
> Hi @zxiaomzxm, You wrote "CARLA version: 0.19.3", do you mean the " CARLA version 0.9.13", right?. I will take a look. Thank you for letting us know

I meet the same issue. InstanceCamera can not show the landscape correctly.

JosenJiang:
@zxiaomzxm Hi, you seem to have implemented an instance segmentation fisheye camera in Carla. Would you mind sharing your methods? I had some problems implementing this feature.

AreopagX:
@Blyron Can you give hints how this might be resolvable?

tdegraaff:
@jorge-kabuto Can now be closed (https://github.com/carla-simulator/carla/pull/8528).

PR #8528, opened 2025-01-06
[Fix] Invisible terrain in instance segmentation (UE4)

tdegraaff:
- Fixes:
  - #5367 
  - #6905
  - #8510 
- Requires one minor change to UnrealCarla: https://github.com/CarlaUnreal/UnrealEngine/pull/36
- For the latest UE4.26 CARLA version

#### Description

The current implementation of `TagActor` in [Tagger.cpp#L104](https://github.com/carla-simulator/carla/blob/ue4-dev/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Game/Tagger.cpp#L104) only considers `UStaticMeshComponents` and `USkeletalMeshComponents`. This causes "terrain" to be invisible in the instance segmentation, as also pointed out in #5367 and #6905. The terrain is only visible in the semantic segmentation, since the Landscapes already have the custom depth pass enabled in the UE editor (in contrast to all other actors, where the depth pass in enabled in `Tagger.cpp`).

Thus, the solution is to extend the `Tagger.cpp` implementation to also tag `ULandscapeComponents`, which make up the terrain. While I thought this would be an easy fix, I was faced with lots of problems. Landscapes are not really designed to be modified during runtime. But after learning a lot about the internals of the UE especially regarding Landscapes and their rendering system, I think I came up with a (more or less) good solution.

#### Side Note

The custom depth stencil buffer ID set in the UE editor for the Landscapes is `22` (`other`), while the actual ID should be `10` (`terrain`). I think this was overlooked in [this change](https://github.com/carla-simulator/carla/commit/dda82223dbc74a530d68a89e831ca4c700fc27cc), where the IDs were adjusted to reflect the Cityscapes labeling. This PR actually fixes this issue silently by correctly setting the stencil values during runtime, as it is also done for all other actors. To be fully coherent, if this PR is merged, the custom depth pass could be disabled on the Landscapes + removing the old IDs from the custom depth stencil buffer.

#### Visualization

The following images show a before/after comparison. 

**Before PR:**
before
* mountains in the background are invisible, as is the small hill in the right
* landscape terrain has wrong color in semantic segmentation (can directly be seen since ground level terrain (not a landscape) has correct color)

**After PR:**
after
* above mention problems are fixed

* **Platform(s)**: Windows 11
* **Python version(s)**: 3.8
* **Unreal Engine version(s)**: 4.26
* `make check` not tested, due to developing on Windows

#### Concluding Remarks

I did not have the chance yet to get hands-down experience with the current state of the UE5 version of CARLA. But having a look into the code of the `ue5-dev` branch, I noticed that there were major changes to the instance segmentation, which now also uses a post processing material instead of this `TaggedComponent` mechanisms. But since the `Tagger` still exists and also does not cover `ULandscapeComponents`, this part of the PR may also be interesting for the `ue5-dev` branch.

And besides of the UE5 version, I think this PR is worthy for people working with the latest UE4 version of CARLA.

#### Implementation details (for reviewers)

Let me "briefly" explain some design decisions:
* I wanted to keep the same idea of attaching a `UTaggedComponents` and creating a `SceneProxy` using the parent while updating the material.
  * `FLandscapeComponentSceneProxy` is not part of the exported `LANDSCAPE_API` of UE, thus requiring a minor modification of Unreal: https://github.com/CarlaUnreal/UnrealEngine/pull/36
* UE uses an optimized rendering system for landscapes, which caches the created `SceneProxies`. When creating a new `SceneProxy` it keeps the ID of the parent, resulting in a collision in this cache's hash map. Thus, the `FGuid` is modified prior to the creation of the new `SceneProxy` and reset afterwards, essentially creating a secondary render system for these new `SceneProxies`.
  * This is also documented at the corresponding code.
* The material of a landscape must be a `ULandscapeMaterialInstanceConstant`, which derives from `UMaterialInstanceConstant`, thus it is not possible to assign a `UMaterialInstanceDynamic` as new material at runtime (as it is done for the other objects). Thus I tried to create a `ULandscapeMaterialInstanceConstant` during runtime from a prepared `UMaterialInstanceConstant` but this does not work in the packaged version, since landscapes require some additional shaders, which are not cooked, when a `MaterialInstanceConstant` is not assigned to a landscape.
  * Instead, the implementation now loads a small `UWorld` during runtime, which contains only a Landscape with the prepared annotation material assigned. The material of this Landscape is retrieved and applied to the `FTaggedLandscapeComponentSceneProxy`.
  * Since the material must be prepared before cooking, I created a Material Instance of the `AnnotationColor` material for the tagging of landscapes, having the color `10, 0, 0` (`terrain` with no ID).
    * Having no ID in the green and blue channel might be a downside, but still better than being invisible.
* Since all of this follows the same scheme of `UTaggedComponents` but is not compatible in the details (especially the material), I created complementary classes specific for this landscape "edge case".

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8528)


tdegraaff:
@Blyron First of all, thank you for merging the secondary PR and starting the CI build.
I noticed that the build failed, but I cannot see the log. Thus, I tried to build CARLA on a Linux server inside a Docker container.
During this, I noticed the following problems:
1) Setup.sh fails when retrieving Xerces-C in [Line 428](https://github.com/carla-simulator/carla/blob/ue4-dev/Util/BuildTools/Setup.sh#L428)
  - The problem is that archive.apache.org seems to have an issue, resulting in a `Unable to establish SSL connection.` error by wget, killing the script.
  - When simply commenting this line, thus retrieving the backup from your servers, Setup.sh runs fine.
2) Performing `make check ARGS="--all --xml" sometimes results in Segmentation Faults, sometimes runs through but some of the last tests are failing.
  - I also performed a build of the current `ue4-dev` branch and observed the same behavior, so this likely does not originate from this PR.
  - Is `make check` known to have problems or maybe this problem originates from building in Docker?
3) Package.sh fails when copying RSS examples in [Line 171](https://github.com/carla-simulator/carla/blob/ue4-dev/Util/BuildTools/Package.sh#L171)
  - I think this originates from this change, where RSS was dropped: https://github.com/carla-simulator/carla/commit/238f84aa4fc1858fad33da8aec7a750c016012e6
  - Maybe it has been overlooked to remove this copy statement in Package.sh

When fixing 1) and 3) and ignoring 2), I successfully built and ran CARLA headlessly on the server. The instance segmentation looked as in the provided screenshots in this PR.

Blyron:
Hi @tdegraaff 
We are aware of the issues so we are manually testing the PR and it will be merged soon. 
After that we have a issue to take care and then we will look at jenkins file. We need to fix it in order to get other PRs merged faster

jorge-kabuto (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8528)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tdegraaff)

Issue #8529, opened 2025-01-07
Fatal Error When Spawning Vehicle with Semantic Lidar in Mine_01


VitaminQRQ:
CARLA version: 0.10.0  

Platform/OS: Windows 11 and Ubuntu 22.04
  
Problem you have experienced: 
Unable to spawn a vehicle with semantic lidar in Mine_01. When I add a semantic lidar sensor, CARLA encounters a fatal error and exits. The same issue occurs on both Windows and Ubuntu. However, the same sensor configuration works fine in Town10.  

What you expected to happen:  
Successfully spawn a vehicle with a semantic lidar sensor.  

Steps to reproduce:  
Replace the `stack.json` file in the ROS2 example with the attached content and run `ros2_native.py`:

```json
{
    "type": "vehicle.lincoln.mkz",
    "id": "ego",
    "sensors": [
        {
            "type": "sensor.camera.rgb",
            "id": "rgb",
            "spawn_point": {"x": -4.5, "y": 0.0, "z": 2.5, "roll": 0.0, "pitch": 20.0, "yaw": 0.0},
            "attributes": {
                "image_size_x": 400,
                "image_size_y": 200,
                "fov": 90.0
            }
        },
        {
            "type": "sensor.lidar.ray_cast",
            "id": "lidar",
            "spawn_point": {"x": 0.0, "y": 0.0, "z": 2.6, "roll": 0.0, "pitch": 0.0, "yaw": 0.0},
            "attributes": {
                "range": 100,
                "channels": 64,
                "points_per_second": 600000,
                "rotation_frequency": 20,
                "upper_fov": 10,
                "lower_fov": -30,
                "atmosphere_attenuation_rate": 0.004,
                "dropoff_general_rate": 0.45,
                "dropoff_intensity_limit": 0.8,
                "dropoff_zero_intensity": 0.4
            }
        },
        {
            "type": "sensor.lidar.ray_cast_semantic",
            "id": "lidar_semantic",
            "spawn_point": {"x": 0.0, "y": 0.0, "z": 2.6, "roll": 0.0, "pitch": 0.0, "yaw": 0.0},
            "attributes": {
                "range": 85,
                "channels": 64,
                "points_per_second": 600000,
                "rotation_frequency": 20,
                "upper_fov": 10,
                "lower_fov": -30
            }
        },
        {
            "type": "sensor.other.gnss",
            "id": "gnss",
[...]

PR #8568, opened 2025-01-14
Fix crash in Mine map when activating semantic lidar

PabloVD:
As stated by [this issue](https://github.com/carla-simulator/carla/issues/8529), activating the semantic lidar in the Mine map in Carla UE5 led to a crash in the server. This was due to the landscapes lacking tags. Hence, when the semantic lidar sensor tried to access the tag of the object, it got a null value, leading to a crash.

In order to fix that, this PR includes a condition to check whether the actor is a landscape, and in that case, asign on the fly a terrain tag.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8568)


jorge-kabuto (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8568)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @PabloVD)

PR #8572, opened 2025-01-15
Fix unpopulated variable in CarlaSetup.sh

### LINKED ISSUE #8571: ./CarlaSetup.sh --interactive stops before downloading content
CARLA version: Carla Unreal 5 (Commit 902108b, currently HEAD)
Platform/OS: Ubuntu 22.04

Problem you have experienced: 

When running `./CarlaSetup.sh --interactive` the scripts exits with an error when trying to create the folder for the Carla Content:

```
Could not find CARLA content. Downloading...
mkdir: cannot create directory ‘/Unreal’: Permission denied
```
This happens because the scripts tries to execute `mkdir -p $cd/Unreal/CarlaUnreal/Content`, but fails because `$cd`. It seems this happened while syncing CarlaSetup.sh and CarlaSetup.bat in 902108b. On Windows, %CD% is always defined in a command prompt, but it seems this was not respected when porting to CarlaSetup.sh

What you expected to happen:

The content directory should be created.

Steps to reproduce:

1. Checkout commit 902108b (currently HEAD)
2. Run `./CarlaSetup.sh --interactive`

Latios96:
When syncing CarlaSetup.bat to CarlaSetup.sh in 902108b, it was not respected that %CD% is always defined in a Windows command prompt. On Linux, $cd is never defined causing the subsequent steps to fail.

Fixes #8571

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8572)

### DIFF
--- a/CarlaSetup.sh
+++ b/CarlaSetup.sh
@@ -65,6 +65,7 @@ else
 fi
 
 # -- CLONE CONTENT --
+cd=$(pwd)
 if [ -d $cd/Unreal/CarlaUnreal/Content ]; then
     echo "Found CARLA content."
 else


MarcelPiNacy-CVC:
Thank you for the PR, we'll use a slightly different method though.

PR #8573, opened 2025-01-15
Fix CarlaSetup.sh invalid variable.

MarcelPiNacy-CVC:
This PR fixes CarlaSetup.sh using "$cd" instead of the correct variable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8573)


MarcelPiNacy-CVC:
This PR fixes #8571.

Issue #8559, opened 2025-01-10
Incorrect semantic segmentation in mine map, 0.10.0


PabloVD:
In the mine map in 0.10.0, rocks and vegetation are not visible in the semantic segmentation. Moreover, the terrain color does not seem to be the proper one. May be due to the usage of semantic segmentation in landscapes.

Note that these problems are not present apparently in instance segmentation cameras.

![Captura de pantalla 2025-01-10 123659](https://github.com/user-attachments/assets/a1bfcf58-873e-4a22-84c8-140c6f2ce516)

yuanhaorannnnnn:
Yes, you'are right, models from landscape isn't compatible with segmentation camera.

utkuoguzman:
So what is the expected time to fix this? Or how can one overcome this? why "add rock/stone tag" is relevant to this?

yuanhaorannnnnn:
> So what is the expected time to fix this? Or how can one overcome this? why "add rock/stone tag" is relevant to this?

you can update the code and content and check the update

PR #8598, opened 2025-01-22
add rock/stone tag

yuanhaorannnnnn:
#### Description

Fixes
Add Rock/Stone asset tag, so anymap with the rock/stone assets will visible on segementation camera, including Mine01, related isseue:
https://github.com/carla-simulator/carla/issues/8559

  * **Platform(s):Linux 22.04.1-Ubuntu GNU/Linux
  * **Python version(s):Python 3.8.19
  * **Unreal Engine version(s):5.5.0

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8598)


utkuoguzman:
Can you do the same for the roads (splines of the landscape object) and the plants as well? They are also looking black in the semantic camera. I only see the car and the sky.

PR #8619, opened 2025-01-30
Invertedai waypoints ue4

rf-ivtdai:
#### Description

Hi there, I am an ML engineer from Inverted AI. I've been using the invertedai_traffic PythonAPI examples, and noticed a few issues need fixing before users can run Carla simulations with our APIs more smoothly.

1. The script spawns actors into Carla based on agent states and properties returned by IAI, but fails for most agents because their carla transform locations has z = 0, using something like z = 0.1 improves things a lot.
2. With the above fix, there are still occasional failures in actor spawning, resulting in downstream errors when `--iai-log` is enabled, this is also fixed in this PR.
3. We introduced a new way to drive IAI agents better: by passing carla waypoints to our API. The command line options and code are included in this change.
4. Carla pedestrians have orientations in degrees but need to be converted into radians prior to being passed to Inverted AI.
5. The default map Town10 is not loaded into Carla world, which causes downstream errors especially when getting waypoints from Carla.
6. We noticed setting a `max_speed` for IAI agents improves driving behaviour.

Fixes #  

  * **Platform(s):*Ubuntu*
  * **Python version(s):*Python 3.8 & 3.10*
  * **Unreal Engine version(s):*UE4*

This PR is applicable to the same example for `ue5-dev` branch, as the logic and resources used calling inverted AI APIs are the same, and are independent of the engine and assets.

#### Possible Drawbacks

No known drawbacks. Tested with IAI cars only or Carla pedestrians + IAI cars.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8619)


PythonAPI/examples/invertedai_traffic.py
```diff
@@ -223,56 +245,56 @@ def initialize_pedestrians(pedestrians):
...
-    world.tick()
-
-    for controller in world.get_actors().filter('controller.ai.walker'):
-        controller.start()
-        dest = world.get_random_location_from_navigation()
-        controller.go_to_location(dest)
-        controller.set_max_speed(0.5 + random.random())
-
-    return pedestrians
+#         loc = world.get_random_location_from_navigation()
```
KieranRatcliffeInvertedAI, line 256: @rf-ivtdai Should this not be outright deleted or kept in instead of being commented?

rf-ivtdai: I'm actually hoping to still run this with IAI cars and the rest of the change.

Blyron (approved):

PR #8714, opened 2025-02-27
[Fix] Invisible spline meshes in instance segmentation (UE4)

tdegraaff:
#### Description

* Requires one minor change to UnrealCarla: https://github.com/CarlaUnreal/UnrealEngine/pull/38
* For the latest UE4.26 CARLA version

I noticed that spline meshes (e.g. the cables on PowerPoles) appear invisible in the instance segmentation (showing the decoded R channel of the instance segmentation):
before

This PR extends the implementation in TaggedCommponent to explicitly handle SplineMeshSceneProxies. SplineMeshComponents are derived from StaticMeshComponents, thus the components are technically already handled by the existing tagging code. But without explicitly deriving a FTaggedSplineMeshSceneProxy from FSplineMeshSceneProxy, they seem to not be rendered correctly. FSplineMeshSceneProxy is currently not public in UE4, thus a minor modification to UnrealCarla is required: https://github.com/CarlaUnreal/UnrealEngine/pull/38

After these changes, spline meshes are rendered correctly in the instance segmentation:
after

  * **Platform(s):** Windows 11
  * **Python version(s):** 3.8

#### Possible Drawbacks

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8714)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
  * carla.ad subpackages are now directly importable and are not directly importable anymore (e.g. import ad)
  * Fixed segfault in traffic manager when trying to access not available vehicles
  * Fixed invalid comparission in python examples/rss
+ * Fixed invisible spline meshes in instance segmentation
 
 ## CARLA 0.9.15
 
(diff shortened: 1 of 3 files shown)


jorge-kabuto:
Thank you for your contribution! Great work as always.

PR #6922, opened 2023-11-17
Javiergr cs/recorder doors

javiergrCS:
#### Description

Added vehicle doors to the recorder / replayer.

Fixes #  

#### Where has this been tested?

  * **Platform(s):** Windows and Ubuntu
  * **Python version(s):** 3.7.9
  * **Unreal Engine version(s):** 4.26

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/6922)

PR #8779, opened 2025-03-23
fix a recorder crash and replay hang for 0.9.15 log issue 

JiayuanZhu:
Fix the recording & replaying issue introduced by this pull request https://github.com/carla-simulator/carla/pull/6922 

- The vehicle door animation event was introduced by pull 6922 and introduced two issues
- The recorder starts to look for CarlaActors before they are registered and ends up with an accessing null pointers crash
- The VehicleDoor enum is not appended at the end of the CarlaRecorderPacketId and thus makes the old log file screw up when replaying by latest runtime built from the ue4-dev branch.  e.g. a uint8_t value of 21 in an old log file should be interpreted as an AnimVehicleWheels even but is no incorrectly interpreted as a VehicleDoor event.  This leads to a hang. 

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8779)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Game/FrameData.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Game/FrameData.cpp
@@ -553,12 +553,18 @@ void FFrameData::AddTriggerVolume(const ATrafficSignBase &TrafficSign)
   {
     return;
   }
+
+  auto CarlaActor = Episode->GetActorRegistry().FindCarlaActor(&TrafficSign);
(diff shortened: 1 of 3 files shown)


glopezdiest (approved):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8779)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @JiayuanZhu)

PR #8823, opened 2025-04-09
Fix DVS camera in ue5-dev branch

### LINKED ISSUE #8724: DVS Camera removed in 0.10.0 / UE5?
While [still documented in 0.10.0](https://github.com/carla-simulator/carla/blob/0.10.0/Docs/ref_sensors.md#dvs-camera), the [example was removed](https://github.com/carla-simulator/carla/commit/e2dc71c39aa70b1ec4e632860df3531215bc1ef9): `Removed DVS and added semantic lidar to mcwt`

A grep for `sensor.camera.dvs` in commit ada75f920642e18cace0a9f85ecf9d4077ddb531 only returns the documentation and some tests.  Was the DVS camera removed with the move to UE5?
```
$ git log -1
commit ada75f920642e18cace0a9f85ecf9d4077ddb531 (HEAD, tag: 0.10.0)
Author: Joel Moriana <joel.moriana@gmail.com>
Date:   Wed Dec 18 15:07:43 2024 +0100

    Release dockerfile updated
$ grep -lIr sensor.camera.dvs .
./Docs/bp_library.md
./Docs/ref_sensors.md
./PythonAPI/test/API/prop_control.py
./PythonAPI/test/smoke/test_sensor_tick_time.py
```

LJFYC007:
#### Description

Fixes #8724  

Fixed a logic error in the DVS Camera and added a corresponding test in `examples/manual_control.py` to verify the fix.

  * **Platform(s):** Windows
  * **Python version(s):** 3.12.7
  * **Unreal Engine version(s):** 5.5.4

#### Possible Drawbacks

Currently, this version triggers an assertion error in `D3D12RenderTarget.cpp` (belong to UE5):

```c++
check( ConvertTypelessToUnorm((DXGI_FORMAT)FormatInfo.PlatformFormat) == ConvertTypelessToUnorm(DXGIFormat) );
```

This is caused by one side using srgb while the other does not. However, for the DVS Camera, this error is non-critical and can be safely ignored.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8823)

### DIFF
--- a/PythonAPI/examples/manual_control.py
+++ b/PythonAPI/examples/manual_control.py
@@ -1131,6 +1131,7 @@ class CameraManager(object):
             ['sensor.camera.instance_segmentation', cc.Raw, 'Camera Instance Segmentation (Raw)', {}],
             ['sensor.lidar.ray_cast', None, 'Lidar (Ray-Cast)', {'range': '50'}],
             ['sensor.lidar.ray_cast_semantic', None, 'Semantic Lidar (Ray-Cast)', {'range': '50'}],
+            ['sensor.camera.dvs', cc.Raw, 'Dynamic Vision Sensor', {}],
             ['sensor.camera.rgb', cc.Raw, 'Camera RGB Distorted',
                 {'lens_circle_multiplier': '3.0',
                 'lens_circle_falloff': '3.0',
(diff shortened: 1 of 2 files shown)


qhaas:
Thanks for fixing this!  We had been holding off upgrading to UE5 given we need the event camera in CARLA.

We likely need to double check units of measure and update documentation accordingly... when comparing the timestamp reported by the event and the elapsed_seconds reported by a snapshot, looks like the events are reporting nanoseconds instead of microseconds.

Update:  Seems to work!

LJFYC007:
Sounds great—thank you all for your hard work!

qhaas:
At some point buildings, trees, etc. (i.e. non-actors) no longer appear in the events, they were there back the commits dated from August.  Vehicles and traffic lights are still there. Odd

qhaas:
Somewhere between the following commits and current, certain types of objects in the scene started disappearing from the dvs camera data.  It appears to be related to the updating of the CARLA fork of UE5.  These older commits work, but are by no means the latest working commits:
* ue5 carla commit ce6eb585aac4d1278b0cfa911fc3942b8eb5b3e9
* carla content (i.e. assets) commit fbfe49a16ae180a2c3557f4d6ab5709b2436d0ef 
* carla ue5 commit eea6a8e677e3509a3e56122d69e3d380bb624bbf

qhaas:
The patch in the OP still works, just have to disable the post processing effects in the dvs camera now, then the events for the other objects (e.g. sky, buildings, etc.) return, e.g.

UPDATE:  Disabling PP causes shadows to not generate events

```
dvs_bp = bp_library.find('sensor.camera.dvs')
dvs_bp.set_attribute("enable_postprocess_effects", 'False')
```

qhaas:
Looks like this fix was rolled into PR 9697

PR #8795, opened 2025-03-27
[Feature] Support of masked materials in instance segmentation (UE4)

tdegraaff:
#### Description

Currently, the instance segmentation replaces all materials on all meshes simply using an opaque material, which can be colorized with an RGB encoding of the type and the object ID. Secondary SceneProxies of all meshes are then rendered with these tagged materials resulting in the instance segmentation output. The current solution has the drawback, that masked materials are simply ignored. Thus, partial transparent surfaces in the RGB image appear fully opaque in the instance segmentation, resulting in very coarse annotations, which also do not match to the semantic segmentation.

This PR addresses this issue by introducing additional logic, that tackles masked materials for the instance segmentation. The following images show a before/after comparison of the semantic channel in the instance segmentation.

|         Town01_Opt          |          Town10HD_Opt           |
| :-------------------------: | :-----------------------------: |
| Town01-1 | Town10HD-1 |
| Town01-2 | Town10HD-2 |

In general, the instance segmentation on vegetation and fences is now as fine-grained as the semantic segmentation. Note, that the main problem with the sky and the lane markings in Town01 (and Town02) are fixed. Also, trees in Town10HD in the background (with LOD4) were previously invisible (since it uses a very special material) and are now visible. The only remaining difference between the semantic segmentation and the instance segmentation is, that the semantic segmentation looks through translucent surfaces, while the instance segmentation does not.

#### Implementation

* The core of this PR is the `UTaggedMaterialsRegistry`, which is now responsible to handle tagged materials for the `UTaggedComponents`
  * It holds a mapping from original materials to modified materials, which are duplicates of the originals, but introduce some additional expressions into the material graph, so that they render with the tag color
  * This keeps the relevant aspects of the original material intact (especially `OpacityMask` and `WorldPositionOffset`), so that the duplicated material behaves like the original, except that it renders with the tag color
* If a `UTaggedComponent's` parent uses a masked material, a tag-injected corresponding material is requested from the registry
  * Otherwise, the default opaque tagged material is used
* When playing in editor (PIE), these tagged materials are created and compiled on demand
* For the packaged version, a commandlet creates a `UTaggedMaterialsRegistry` for the current package, which then will be cooked and copied to the package

#### Where has this been tested?

* **Platform(s):** Windows 11, Ubuntu 18.04
* **Python version(s):** 3.8
* **Unreal Engine version(s):** 4.26
* Tested in Editor (Windows) and Packaged (Windows + Linux)

#### Possible Drawbacks

[...]

PR #8909, opened 2025-05-15
[Feature, UE4] Configurable translucency in depth and semantic + instance segmentation

tdegraaff:
#### Description

Currently, the semantic segmentation and the depth sensors are looking through translucent materials, always detecting the first opaque material along a ray. In contrast, the instance segmentation is ignoring translucency entirely and detects always the first hit along a ray. This leads to inconsistencies between the different ground truth annotations for cameras. Additionally, in monocular depth estimation, the ground truth depth typically does not return the depth behind glass (c.f. [Ramirez et al.](https://arxiv.org/pdf/2206.04671)), yet they state that having both (first and last hit of a ray) could be desirable. Lastly, in datasets providing semantic segmentation masks, they typically do not spare out windows (e.g. from cars; c.f. [Cityscapes](https://www.cityscapes-dataset.com/examples/)), as it is currently the case in CARLA. This currently looks like this:

| sem. seg | inst. seg | depth |
| ---------- | --------- | ------ |
| 1s1 | 1i0 | 1d1 |

This PR addresses these issues, while making it configurable during runtime, which behavior is desired by the user (via newly added `world.set_annotations_traverse_translucency`). To be consistent with the literature and public datasets, the default is set to `false`, i.e. annotations by default do not traverse translucent materials. With these changes, it can now look like this:

| w.s_a_t_t       | sem. seg          | inst. seg         | depth             |
| --------------- | ----------------- | ----------------- | ----------------- |
| False (Default) | 1s0 | 1i0 | 1d0 |
| True            | 1s1 | 1i1 | 1d1 |

#### Implementation

This feature extends on https://github.com/carla-simulator/carla/pull/8795:

* the default `AnnotationColor` material is extended to support being entirely masked, based on two parameters `IsTranslucent` and `TraverseTranslucency`
  * only if `IsTranslucent` and `TraverseTranslucency` are `true` on a tagged material, it is rendered invisible in the TaggedSceneProxies
* the `TaggedMaterialRegistry` is now additionally checking, if a material uses a translucent blend mode
  * if yes, `IsTranslucent` it set to `true` on the tagged material, else the default `false` is kept
* `TraverseTranslucency` on the other side is a parameter from a MaterialParameterCollection, whose value can be changed via the `world.set_annotations_traverse_translucency` API function
* `DepthCamera` and `SemanticSegmentationCamera` now also use the tagged scene proxies to render their result instead of the default scene proxies for the RGB cameras
  * This way, all ground truth annotations for the camera are consistent in their behavior

* **Platform(s):** Windows 11
* **Python version(s):** 3.8

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8909)


Blyron (approved):

Blyron:
@tdegraaff Once again. Thanks for your job!

PR #8937, opened 2025-05-24
Working Windows ROS2 Native Build

rbellerobotics:
#### Description

This merge requires updates the build pipeline to function on Windows when ROS2 is enabled. The build script has been updated to account for different types of VS code as wel.

This is known to work on Windows 11 with the most recent SDK. This was also verified on Linux.

This has not tested spawning a vehicle and sensors, as well as ensuring their outputs are correct. It only went as far as hitting play in the editor and verifying the clock outputs time with `ros2 topic echo /clock` in WSL.

#### Possible Drawbacks

Windows 10 SDKs not tested. Sensors and vehicles not tested.


CMake/Options.cmake
```diff
@@ -53,7 +53,7 @@ carla_option (
 carla_option (
   ENABLE_ROS2
   "Enable ROS2."
-  OFF
+  ON
```
rbellerobotics, line 56: Note that I defaulted ROS2 to be on. I wasn't sure if getting Windows and Linux both working was enough to consider it a solid default.

LibCarla/source/carla/ros2/ROS2.cpp
```diff
@@ -94,7 +94,7 @@ void ROS2::SetFrame(uint64_t frame) {
       if (_controller->HasNewMessage()) {
         auto it = _actor_callbacks.find(actor);
         if (it != _actor_callbacks.end()) {
-          VehicleControl control = _controller->GetMessage();
+          VehicleControl control = _controller->GetCarlaMessage();
```
rbellerobotics, line 97: This was because Windows kept failing to compile this and would return an error where it said "GetMessageA()" couldn't be found. Changing the name worked so my best guess is that there is some other `GetMessage` function that interfered with it. Could be wrong on that one though.

Ros2Native/CMakeLists.txt
```diff
@@ -9,67 +9,136 @@ include (ExternalProject)
...
+      -DCOMPILE_EXAMPLES=OFF
+      -DCOMPILE_TOOLS=OFF
+      -DTHIRDPARTY_Asio=FORCE
+      -DTHIRDPARTY_fastcdr=FORCE
+      -DTHIRDPARTY_TinyXML2=FORCE
+      -DSQLITE3_SUPPORT=OFF
+      -DOPENSSL_FOUND:BOOL=ON
+      -DOPENSSL_INCLUDE_DIR:FILEPATH=${UE_OPENSSL_INCLUDE}
+      -DOPENSSL_SSL_LIBRARY:FILEPATH=${UE_OPENSSL_LIBS}/libssl.a
+      -DOPENSSL_CRYPTO_LIBRARY:FILEPATH=${UE_OPENSSL_LIBS}/libcrypto.a    DEPENDS foonathan_memory
```
rbellerobotics, line 100: Need to fix this syntax. Depends is on the same line as the arg.

PR #8945, opened 2025-05-29
Fix map names in smoke tests

joel-mb:
#### Description

This PR updates the smoke tests to use a specific set of maps instead of getting all available maps using `get_available_maps`.

`get_available_maps` could return maps without an OpenDRIVE file, which causes the tests to fail.

To make sure all maps are tested, new maps should be added manually to the list.

  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** CARLA fork

#### Possible Drawbacks

None

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8945)

### DIFF
--- a/PythonAPI/test/smoke/test_map.py
+++ b/PythonAPI/test/smoke/test_map.py
@@ -19,17 +19,14 @@ class TestMap(SmokeTest):
 
     def test_load_all_maps(self):
         print("TestMap.test_load_all_maps")
-        map_names = list(self.client.get_available_maps())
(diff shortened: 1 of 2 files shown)


Blyron (approved):

glopezdiest (commented):
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8945)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #8951, opened 2025-05-30
Added Left Hand Traffic support

glopezdiest:
### Description

This PR adds left-handed traffic support. Left-handed traffic is defined in two stages, the road and the traffic elements.

For the roads, the OpenDrive is used, specifically the `rule` road attribute, like the example below

```xml
<road name="Road 0" length="1.3310253693587601e+1" id="0" junction="-1" rule="LHT">
    <link>
        <predecessor elementType="road" elementId="3" contactPoint="end" />
        <successor elementType="road" elementId="10" contactPoint="start" />
    </link>
...
</road>
```
The *rule* attribute works as follows:
- _LHT_ for left-handed traffic
- _RHT_ for right-handed traffic
- Any other values, or if the _rule_ attribute is missing, defaults to right-handed traffic

As for the traffic elements (traffic lights, stops, yield, speed limits...), the creators of the maps are responsible for placing them in the correct spot so that they make sense in a LHT environment. As such, this doesn't mean that available maps, such as  Town10HD_Opt, can now be used as a LHT.

Having maps with mixed RHT and LHT is possible, though limited testing has been done in that area.

The TrafficManager has been modified to automatically use the direction set by the road. The `keep_right_rule_percentage` parameter has been renamed to `keep_slow_lane_rule_percentage`, as the right rule didn't make sense in LHT. For left-handed roads, vehicles controlled by the TM will move to the left instead (if the parameter is active). **THIS BREAKS COMPATIBILTIY WITH PREVIOUS VERSIONS**.

Waypoints now have an additional **is_rht** attribute to determine if they are part of a right-handed road

  * **Python version(s):** 3.8

### Possible Drawbacks

Theoretically none as this is an addition, but some RHT code has changed to accommodate for LHT.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/8951)


utkuoguzman:
Is this in a new 20GB content download? because I pulled Carla code now and  it gives errors about not finding these 2 BPs in my content folder.

Issue #9018, opened 2025-06-30
Add custom blueprint to Carla blueprints library/actors registry Carla 0.9.15


astralmaster:
I have a custom blueprint that I created in Carla's Unreal Editor 4 that is based on static mesh I imported to Carla using `make import`. This blueprint has some logic/code to it, I can drag and drop it to the map and it works fine, but I want to be able to spawn it in Carla using python API. What is the right way to make Carla 'see' my custom blueprint so that it is available for spawning in the blueprints library?   Carla 0.9.15 

The closest issue to this I could find was [https://github.com/carla-simulator/carla/issues/3580](https://github.com/carla-simulator/carla/issues/3580) , however, my blueprint is not a Sensor, but rather a moving object (let's say its a rolling ball)

astralmaster:
> That is not possible in 0.9.15

Is it possible in another version?

Blyron:
My bad I thought it was implemented but it is not. For that purpouse you will need to create a new factory in Ue5 dev branch exists
Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/Factory/BlueprintActorFactory.h
Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/Factory/BlueprintActorFactory.cpp

astralmaster:
That looks promising! Thank you, I will try porting this to UE4, if successful, I will do a pr, hopefully it will help others too.

Blyron:
Okay thanks. Remember to add a doc if you want to have more exposition!

uuauto2024:
Hi, I am facing the same issue, have you found any solution?

CorneliusHsiao:
Thanks @astralmaster for your implementations. I want to mention several more steps that are necessary to make blueprint import function work on my pull of ue4-dev:
1. In `Unreal\CarlaUE4\Plugins\Carla\Source\Carla\Actor\ActorBlueprintFunctionLibrary.h`, add `#include "Carla/Actor/BlueprintParameters.h"`.
2. Copy `BlueprintParameters.h` from `ue5-dev` branch to `Unreal\CarlaUE4\Plugins\Carla\Source\Carla\Actor\BlueprintParameters.h`.
3. In `Unreal\CarlaUE4\Plugins\Carla\Source\Carla\Game\CarlaGameModeBase.cpp`, add the following in the implementation of `ACarlaGameModeBase::ACarlaGameModeBase`:
```
  ...
[...]

PR #9026, opened 2025-07-03
Custom blueprint import ability added

astralmaster:
#### Description

Adds the ability to import custom blueprints created in UE4 to Carla's blueprint library, spawnable via Python API. Ported from Carla ue5-dev branch. In order to add your blueprint, create `BlueprintParameters.json` file in `/CarlaUE4/Content/Carla/Config/` with following format:

```
{
  "Blueprints": [
    {
      "Name": "SomeBlueprintName",
      "Path": "/Game/Path/To/SomeBlueprintName"
    },
	{
      "Name": "SomeOtherBlueprintName",
      "Path": "/Game/Path/To/SomeOtherBlueprintName"
    }
  ]
}
```

If, for example, in your UE4 Content Browser you have a blueprint with path `"Content -> Blueprints -> SomeBlueprintName"`, then you would modify the `Path` in `BlueprintParameters.json` as `"Path": "/Game/Blueprints/SomeBlueprintName"`.

Then, from Python API:

```
world = client.get_world()
blueprint_library = world.get_blueprint_library()
blueprint_name = 'SomeBlueprintName'
bp = blueprint_library.find(blueprint_name)
```

Fixes https://github.com/carla-simulator/carla/issues/9018

  * **Platform(s):** Windows 10 x64
  * **Python version(s):** 3.7
  * **Unreal Engine version(s):** UE 4.26.2

#### Possible Drawbacks

The code uses `FindFunction()` to determine whether `PostProcessBlueprint` has been used/implemented in UE4, should probably check if there is an easier, more robust way. This was needed to correctly determine whether the return value of `PostProcessBlueprint` should be taken into account when setting the `SpawnResult.Status` in `BlueprintActorFactory.cpp`'s `SpawnActor()`. In the original `BlueprintActorFactory.cpp` from UE5, the code checks if `PostProcessBlueprint `returns true/false and sets the  `SpawnResult.Status` accordingly, but if `PostProcessBlueprint ` is not used in UE4, then the default return value is `false` and that wrongly (I assume) sets the  `SpawnResult.Status` to `EActorSpawnResultStatus::UnknownError`.

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
@@ -1699,6 +1699,45 @@ void UActorBlueprintFunctionLibrary::MakePropDefinitions(
   FillActorDefinitionArray(ParameterArray, Definitions, &MakePropDefinition);
 }
 
+
+void UActorBlueprintFunctionLibrary::MakeBlueprintDefinition(
+  const FBlueprintParameters &Parameters,
+  bool &Success,
+  FActorDefinition &Definition)
+{
+  FillIdAndTags(Definition, TEXT("blueprint"), Parameters.Name);
+  AddRecommendedValuesForActorRoleName(Definition, {TEXT("blueprint")});
+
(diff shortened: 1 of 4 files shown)


Blyron:
Could you add a small .md on how have you done it to help people in the community?

astralmaster:
That should do it!

PR #9089, opened 2025-07-28
Close proc if its already running before launch it

Blyron:
Add logs if recast is already running when loading new episode.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9089)


MarcelPiNacy-CVC (approved):

LuisPovedaCano (approved):
:+1:

PR #9091, opened 2025-07-29
Aaron/addgeolocation to transform

Blyron:
Added Geolocation inverse operation and test for geolocation

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9091)


glopezdiest:
The code LGTM. Two minor things though:

- The geolocation test has a 0.1 latitutude tolerance, which is about 10kms of distance. It should be reduced to a much lower value.
- Docs are missing.

joel-mb (approved):

PR #9095, opened 2025-07-30
Relax asserts

Blyron:
#### Description

Relax conditions of asserts when getting road predecessors and successors to avoid crashes in Largemaps.

Fixes #  

  * **Platform(s):** Windows/Ubuntu22

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9095)


MarcelPiNacy-CVC (approved):
Looks good, but maybe we should emit warnings.

LuisPovedaCano (approved):
:+1:

PR #9217, opened 2025-09-01
Allow rotation (pitch) of all vehicle wheels via API

lrstttl:
#### Description

Added two additional API calls to set/get a vehicle's wheel pitch angle.  
The setter only works with disabled physics, but can be useful for co-simulations to update the vehicle animation.

Both calls are very similar to the existing calls `carla::client::Vehicle:.SetWheelSteerDirection` and `carla::client::Vehicle:.GetWheelSteerAngle` (see https://github.com/carla-simulator/carla/blob/ue5-dev/LibCarla/source/carla/client/Vehicle.h#L78-L84).

  * **Platform(s):** Windows 10/11, Ubuntu 20.04, Ubuntu 22.04
  * **Python version(s):** Python 3.10, Python 3.8
  * **Unreal Engine version(s):** Unreal Engine 4.26, 4.27

#### Possible Drawbacks

Setter only works with disabled physics. However, existing calls `carla::client::Vehicle:.SetWheelSteerDirection` and `carla::client::Vehicle:.GetWheelSteerAngle` already exist and have the same drawback.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9217)

### DIFF
--- a/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp
+++ b/Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp
@@ -966,6 +966,38 @@ float ACarlaWheeledVehicle::GetWheelSteerAngle(EVehicleWheelLocation WheelLocati
   }
 }
 
+void ACarlaWheeledVehicle::SetWheelPitchAngle(EVehicleWheelLocation WheelLocation, float AngleInDeg) {
+
+  if (bPhysicsEnabled == false)
+  {
+    check((uint8)WheelLocation >= 0)
+    UVehicleAnimInstance *VehicleAnim = Cast<UVehicleAnimInstance>(GetMesh()->GetAnimInstance());
+    check(VehicleAnim != nullptr)
+    VehicleAnim->SetWheelPitchAngle((uint8)WheelLocation, AngleInDeg);
+  }
+  else
+  {
+    UE_LOG(LogTemp, Warning, TEXT("Cannot set wheel pitch angle. Physics are enabled."))
+  }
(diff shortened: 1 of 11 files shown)


Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp
```diff
@@ -966,6 +966,38 @@ float ACarlaWheeledVehicle::GetWheelSteerAngle(EVehicleWheelLocation WheelLocati
...
+  if (bPhysicsEnabled == false)
+  {
+    check((uint8)WheelLocation >= 0)
+    UVehicleAnimInstance *VehicleAnim = Cast<UVehicleAnimInstance>(GetMesh()->GetAnimInstance());
+    check(VehicleAnim != nullptr)
+    VehicleAnim->SetWheelPitchAngle((uint8)WheelLocation, AngleInDeg);
+  }
+  else
+  {
+    UE_LOG(LogTemp, Warning, TEXT("Cannot set wheel pitch angle. Physics are enabled."))
```
Blyron: Please remove LogTemp and use proper Log category such us LogCarla

Blyron (changes requested):

Unreal/CarlaUE4/Plugins/Carla/Source/Carla/Vehicle/CarlaWheeledVehicle.cpp
lrstttl: Updated it

Blyron:
Checking using our workflows.

Blyron (approved):

PR #9333, opened 2025-10-08
Added Geo Projections 

glopezdiest:
#### Description

Right now, CARLA only parses the reference latitude and longitude, ignoring the projection and parameters (supposing a TransverseMErcator with spherical Earth). **This PR adds the capability of using different types of projections** with several parameters. The supported projections are:

- TranverseMercator
- UniversalTransverseMercator (UTM)
- WebMercator
- Lamber Conformal Conic 2SP

Different types of ellipsoid are now also supported, as well as indicating specific ellipsoid paramters. However, datums aren't supported, and all units are supposed to be meters and degrees.

Explanation of the changes:
1) Transform functions have been move from **GeoLocation** to a new **GeoProjection** class.
2) This has caused the **GNSS** and **V2X** sensors to need the geoprojection instead of the georeference, modifying several files responsible for sending the data.
3) Projection parameters are defined in **GeoProjectionParams**.

The projections have been exposed onto the **PythonAPI** so that the transformation functions can be used with other projections, not just the maps one:
- `carla.Map.transform_to_geolocation()` and `carla.Map.geolocation_to_transform()` now accept an optional argument `projection`. Providing no projection will use the maps one (the one defined in the OpenDRIVE geoReference).
- Added new classes `carla.GeoEllipsoid`, `carla.GeoProjectionTM`, `carla.GeoProjectionUTM`, `carla.GeoProjectionWebMerc` and  `carla.GeoProjectionLCC2SP` to define custom projections.
- Added new functions `carla.Map.get_georeference()` which returns the geo reference used by the map as a `carla.GeoLocation()`
- Added new functions `carla.Map.get_geoprojection()` which returns the geo projection used by the map as a `carla.GeoProjection...()` (the exact class depends on the projection)

**New smoke tests** have been added which test the transformation using both the maps projection as well as the custom ones.

While the OpenDRIVE uses PROJ string, **CARLA doesn't exactly parse them as PROJ** does. Here are the main differences:
- Last-value wins policy is ignored. Specifically for the ellipsoid, first a specific name is checked, then the `a`, `b`, `f` and `f_inv`, regardless of place in the string.
- Default ellipsoid is a WGS84.
- All parameters have default values, throwing a warning instead of an error if one is missing.

The OpenDrive files for all maps will be updated shortly to maintain compatibility with previous data.

  * **Python version(s):** 3.10

### Possible Drawbacks

None, as the new implementation keeps backwards compatibility with the old parsing, and the rest are additions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9333)


LibCarla/source/carla/geom/GeoProjection.h
```diff
@@ -0,0 +1,105 @@
...
+    using ProjectionParams = boost::variant2::variant<
+        TransverseMercatorParams,
+        UniversalTransverseMercatorParams,
+        WebMercatorParams,
+        LambertConformalConicParams>;
+
+    struct GeoProjection {
+
+        template <typename T>
+        static auto Make(T&& args)
```
jorge-kabuto: I think it's standard to avoid using auto in function definitions

LibCarla/source/carla/geom/GeoProjectionsParams.h
```diff
@@ -0,0 +1,156 @@
...
+        double y_0 = 0.0f;
+        Ellipsoid ellps = Ellipsoid();
+
+        bool operator==(const TransverseMercatorParams &rhs) const {
+            return (lat_0 == rhs.lat_0) && (lon_0 == rhs.lon_0) && (k == rhs.k)
+                && (x_0 == rhs.x_0) && (y_0 == rhs.y_0) && (ellps == rhs.ellps);
+        }
+    
+        bool operator!=(const TransverseMercatorParams &rhs) const {
+            return !(*this == rhs);
```
jorge-kabuto: Is there a reason to use different comparison logic to the == operator above?

LibCarla/source/carla/geom/GeoProjectionsParams.h
```diff
@@ -0,0 +1,156 @@
...
+    class Ellipsoid {
+    public:
+
+        Ellipsoid() = default;
+
+        Ellipsoid(double a, double f_inv):
+            a(a), f_inv(f_inv) {}
+
+        double a = 6378137.0;
+        double f_inv = std::numeric_limits<double>::infinity();  // Default to sphere
```
jorge-kabuto, line 31: Are these data members supposed to be publicly accesible after initialization?

LibCarla/source/carla/geom/GeoProjection.h
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        /// Get the paramaters.
+        const ProjectionParams& getParams() const {
+            return params;
+        }
+
+        /// Set the Proj raw string.
+        void setRawReference(std::string s) {
+            raw_reference = std::move(s);
```
jorge-kabuto: The raw_reference variable is not a reference in the C++ meaning, but it does move the string data to avoid the copy overhead. Does the name mean something else?

PythonAPI/carla/source/libcarla/Map.cpp
```diff
@@ -53,6 +54,17 @@ static auto GetTopology(const carla::client::Map &self) {
   return result;
 }
 
+static auto GetGeoReference(const carla::client::Map &self) {
```
jorge-kabuto: Avoid auto on function definitions

glopezdiest: Done, keeping the auto at GetGeoProjection as it can be any of the 4 projections

LibCarla/source/carla/geom/GeoProjection.h
glopezdiest: Ah I see, I didn't know that, change to setProjString

LibCarla/source/carla/geom/GeoProjectionsParams.h
glopezdiest: Yes, they are available in the Python API and should be modifiable

LibCarla/source/carla/geom/GeoProjectionsParams.h
glopezdiest: Done

LibCarla/source/carla/geom/GeoProjection.h
glopezdiest: Done

jorge-kabuto (dismissed):
> @jorge-kabuto reviewed 15 of 20 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9333)* status: all files reviewed, 5 unresolved discussions (waiting on @glopezdiest)

Blyron (approved):

PR #9384, opened 2025-10-24
Improved large vehicles at junctions

glopezdiest:
### Description

This PR adds a new behavior for large vehicles at junctions, more closely matching their real behavior. This means that now the large vehicles take longer to turn but do so in a wider turn, exiting the intersection in a much more straight trajectory. This results in a reduction of collisions with other vehicles in the road as well as a lesser invasion of the sidewalk while turning.

This has been done by adding a dynamic offset to the vehicle's buffer. The **HandleLargeVehicleJunction** is responsible for calculating and saving the length of the incoming junction, as well as knowing the state of the vehicle (entering the junction / inside it / exiting it).

That information is used by the **CalculateBaseOffset** function at the MotionPlanStage.cpp to calculate the offset. The offset keypoints are as follows:
- At the start of the intersection, the offset is 0, to ensure no sudden changes to the vehicle's buffer.
- At 30% of the junction, the offset reaches its maximum value of 1.0m.
- The offset then gradually decreased, reaching 0m at 50% and it's minimum value of -1.0m at 70% of the junction
- The offset goes up to 0 at the exit for the same reason as the entry.
- The offset sign depends on the type of turn (right or left). Straight motions through the junction don't have this offset.

The **LARGE_VEHICLES_JUNCTION_OFFSET** and **LARGE_VEHICLES_JUNCTION_POINT** constants are the ones that fix these keypoints. Turns must high a minimum turn radius for the offset to be applied.

Note that this doesn't completely solve all issues with large vehicles at junctions, but it smooths things out in many situations

  * **Python version(s):** 3.10

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9384)


LibCarla/source/carla/trafficmanager/MotionPlanStage.cpp
```diff
@@ -305,6 +312,59 @@ bool MotionPlanStage::SafeAfterJunction(const LocalizationData &localization,
...
+  }
+
+  float junction_length = large_vehicles[actor_id].first;
+  bool turn_flag = large_vehicles[actor_id].second;
+  float max_offset = LARGE_VEHICLES_JUNCTION_OFFSET;
+  float max_offset_point = LARGE_VEHICLES_JUNCTION_POINT;
+
+  // From offset to -offset, but making sure the entries and exits have offset 0 for smooth transition.
+  // i.e the vehicles opens up at the entry to perform a wider turn later on, exiting in a straighter trajectory.
+  float t = cg::Math::Clamp(junction_missing_length / junction_length, 0.0f, 1.0f);
```
LuisPovedaCano, line 352: Try to use more meaningful names for the variables

LuisPovedaCano (approved):
Good job.
But please try to change what I commented.

PR #9450, opened 2025-11-21
Extend ROS2 support Step 2: Fine grained ServerSynchronization

berndgassmann:
#### Description
This is one of the second steps of the ROS2 extension series.
This one requires #9431 to be merged; until then it's listed as a draft.

Introduced a fine grained ServerSynchronization mechanism, where each
synchonization participant is treated independently and interacts
with the synchronization of the carla-server individually. If a client
is disconnected (or dies) the synchronization state of all participants
registered via that client are dropped, i.e. the server will continue
running in case the participants of that client were the only ones
demanding synchronous mode.

The synchronization interface provides means of a time window, up to
which the server is allowed to run. Like this, every client can prevent
the carla-server to run too fast depending on their individual speed.
There is no sync-master anymore. Every client decides for its own if
it requires synchronization or not.

Drawback of this change: some existing code might have to be changed
(see removal of synchronous_master in generate_traffic.py).

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9450)

### DIFF
--- a/Util/BuildTools/Setup.sh
+++ b/Util/BuildTools/Setup.sh
@@ -217,7 +217,7 @@ unset BOOST_BASENAME
 # -- Get rpclib and compile it with libc++ and libstdc++ -----------------------
 # ==============================================================================
 
-RPCLIB_PATCH=v2.2.1_c5
+RPCLIB_PATCH=carla-callbacks
 RPCLIB_BASENAME=rpclib-${RPCLIB_PATCH}-${CXX_TAG}
 
 RPCLIB_LIBCXX_INCLUDE=${PWD}/${RPCLIB_BASENAME}-libcxx-install/include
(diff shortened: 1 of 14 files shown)


Util/BuildTools/Setup.sh
```diff
@@ -217,7 +217,7 @@ unset BOOST_BASENAME
 # -- Get rpclib and compile it with libc++ and libstdc++ -----------------------
 # ==============================================================================
 
-RPCLIB_PATCH=v2.2.1_c5
+RPCLIB_PATCH=carla-callbacks
```
berndgassmann, line 220: Here we should enter v2.2.1_c6 after a corresponding tag at the head of the branch 
https://github.com/carla-simulator/rpclib/tree/carla-callbacks
is created.

berndgassmann:
Rebased, fixed windows build and smoke test runs.

berndgassmann:
Sorry, forgot to rebuild of PythonAPI after changing the client code and overlooked that the client has at least to wait for the next tick in non-sync-mode to receive the updated settings value...
Now it actually should work.

Util/BuildTools/Setup.sh
MarcelPiNacy-CVC: Will do, thank you.

MarcelPiNacy-CVC (approved):

PR #9451, opened 2025-11-21
Extend ROS2 support Step 3: Rework ROS2 handling

berndgassmann:
#### Description
Extend ROS2 support Step 3: Rework ROS2 handling

This commit is extending the built-in ROS2 support of CARLA:
- the internal interface is simplified and object oriented to allow
  easier ROS2 extensions in future and reduce code duplication
- the sensor/actor stream existing for the TCP counterparts are used
  in a shared manner, which results in lower processing overhead and
  requires little individual code changes for ROS2 within UE-sensors
- enable Boost and Asio exceptions for FastDDS to ensure  properly
- additional service interfaces are provided to control CARLA directly
  from ROS2 (Map, Blueprint and Episode handling, Spawn/DestroyObject)
- convenient interfaces previously available via carla_ros_bridge are
  provided (e.g. pseudo sensors, object messages, traffic light/sign,
  actor/sensor lists)
- new interfaces like e.g. CarlaVehicleTelemetryData, V2X,
  SynchronizationWindow,WeatherControl
- get detailed vehicle meshes as exact ground truth for e.g. LiDAR
  perception
- Vehicle publisher implements a variety of individual publishers to
  mimick also old ROS-bridge convenient publisher:
  + CarlaVehicleInfo
  + CarlaVehicleControlStatus
  + Speed
  + Odometry
  + Object
  + ObjectWithCovariance
  + VehicleTelemetryData
- Vehicle subscribers:
  + VehicleControlSubscriber
  + AckermannControlSubscriber,
  + SetTransformSubscriber
- use of c++17 to support std::sample() usage
- Dynamic switching on ROS visibility is possible for all actors now
- Default startup behavior is selectable by ROS2TopicVisibility
  parameter in DefaultGame.ini: If true, then all and every topic that
  is currently offered by the implementation is visible from the
  beginning. If false, then only the sensors/actors created via the
  Client- or ROS-Interface are visible by defaults. Others can be
  activated via EnableForRos() calls (this allows for disabling
  interfaces e.g. for leaderboard)
- ensure multiple service calls at once are working:
  + qos history kind eprosima::fastdds::dds::KEEP_ALL_HISTORY_QOS
  + qos reliablility kind
    eprosima::fastdds::dds::RELIABLE_RELIABILITY_QOS
  + qos durability kind
    eprosima::fastdds::dds::TRANSIENT_LOCAL_DURABILITY_QOS
    for reader, writer, topic
- add set_transform() call to RGB Cameras
- add odometry publisher to walkers
- filter out invalid bounding boxes observed on a few traffic signs
using bounding box transmission within object data

  * **Platform(s):** ...Ubuntu 22.04 and 24.04  (with some 24.04 adaptions which come on a separate PR)
  * **Python version(s):** ...3.10, 3.12
  *  **ROS2 version(s):** humble (22.04), kilted (24.04)

#### Possible Drawbacks

Some features of the old ros-brigde which are client only code are not yet implemented (lane invasion sensor, enable_autopilot) at server side, but are kept available by the carla_ros_client node which deploys a tiny subset of the former carla_ros_bridge to provide the ros2 interfaces via the carla python client.
- the set_velocity interface (set movement via Twist) is currently not supported; is that used at all?
Did not recognize any other interface that might be missing yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9451)


berndgassmann:
Rebased to the (hopefully) fixed step2 commit.
And reverted some unnecessary interface changes on ROS side to allow maximum compatibility to the previous ros_bridge behavior.
Currently mainly some features implemented at CARLA client side are missing e.g. lane invasion, set_autopilot (because traffic manager is handled by client).

berndgassmann:
When this is merged, also the updated carla_msgs should be merged into the dev branch
https://github.com/carla-simulator/ros-carla-msgs/pull/24

Then we should create also dev branch within the ros-bridge repository that we don't mess up the master with the development changes.

After the carla_msgs have been merged I can update the ros-bridge PR with the correct carla_msgs commit here:
https://github.com/carla-simulator/ros-bridge/pull/762

berndgassmann:
The latest commits were switching to a more stable FastDDS version (the latest 2.x release from last week) which is fully compatible with Humble, Jazzy, Iron, Kilted. The above is running at our site stable under Humble (Ubuntu 22.04) as well as Kilted (Ubuntu 24.04 -- needs an additional commit -- see draft #9478). Should also run without any problems under Jazzy (see #9551) and most probably also Iron (at least I used Iron with 2.14.4 quite some time ago). 

In addition, the latest commits were mainly targeting stable and deterministic runs even under heavy system load by
fine tuning FastDDS settings, polishing publishing order, reducing load on /tf by moving sensors to /tf_static because most sensors might be static ones attached at fixed positions to moving vehicles or within the environment (maybe on could make it configurable at some point in time e.g. by refining the publish_tf parameter from boolean to some enum { Disable, Static, Dynamic }?) and some smaller fixes.

Blyron:
Sorry Bernd, but this PR has too many files; So we are closing it. Look at Jesus Anaya work, it is smoother to land. 
Thanks for your hard work.

berndgassmann:
Sure, yes I was looking into how to combine the two somehow, but due to the many added datatypes in our banch this will be not straight forward, I fear.

PR #9523, opened 2026-01-19
Fix deadlock in RPC server shutdown

berndgassmann:
#### Description

If there are pending calls which are not yet processed while the CARLA server is stopped, the respective waiting for the future of the sync call result to become ready can block the worker thread from finishing and therefore blocks the join() call on server.stop(). This leads to a deadlock.
With the current change, the thread checks every second if a shutdown was requested and returns an empty response on shutdown. 

Maybe the issue was introduced by the new server synchronization handling, but might have been existing also for longer now. Reproduction on my end was easy.
Start CARLA server, start generate_traffic.py, stop CARLA server which was then blocking...

Fixes #  

  * **Platform(s):** ...Ubuntu 24.04
  * **Python version(s):** ...

#### Possible Drawbacks
None.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9523)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Latest Changes
+ * Fix possible RPC Server deadlock on shutdown
  * Change foonathan memory vendor branch to v1.3.1.
  * Improved the way the TrafficManager controls large vehicles at junctions, reducing the frequency of collisions with other elements in the simulation.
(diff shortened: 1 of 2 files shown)


Blyron:
Bernd the CI/CD seems not to be working, could you recreate the PR but this time targeting the proper branch from beginning?

berndgassmann:
Yes, sure. Accidentally it was first UE5 branch...

PR #9529, opened 2026-01-22
Fix RPC Server deadlock on shutdown

berndgassmann:
#### Description

If there are pending calls which are not yet processed while the CARLA server is stopped, the respective waiting for the future of the sync call result to become ready can block the worker thread from finishing and therefore blocks the join() call on server.stop(). This leads to a deadlock.
With the current change, the thread checks every second if a shutdown was requested and returns an empty response on shutdown. 

Maybe the issue was introduced by the new server synchronization handling, but might have been existing also for longer now. Reproduction on my end was easy.
Start CARLA server, start generate_traffic.py, stop CARLA server which was then blocking...

Fixes #  

  * **Platform(s):** ...Ubuntu 24.04
  * **Python version(s):** ...

#### Possible Drawbacks
None.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9529)

### DIFF
--- a/LibCarla/source/carla/rpc/Server.h
+++ b/LibCarla/source/carla/rpc/Server.h
@@ -17,6 +17,7 @@
 #include <rpc/server.h>
 #include <rpc/this_session.h>
 
+#include <atomic>
 #include <future>
 
 namespace carla {
@@ -80,7 +81,7 @@ namespace rpc {
     boost::asio::io_context _sync_io_context;
 
     ::rpc::server _server;
-    bool _shutdown_in_progress{false};
+    std::atomic_bool _shutdown_in_progress{false};
   };
 
   // ===========================================================================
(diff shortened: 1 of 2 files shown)


LibCarla/source/carla/rpc/Server.h
```diff
@@ -79,6 +80,7 @@ namespace rpc {
     boost::asio::io_context _sync_io_context;
 
     ::rpc::server _server;
+    bool _shutdown_in_progress{false};
```
MarcelPiNacy-CVC: One question: shouldn't this be atomic_bool?
We can then potentially lower the delay in the wait_for call.

berndgassmann: Yes, sure. What delay do you have in mind. 100ms/200ms? But since it's only relevant on shutdown I didn't want to create overhead within standard runs.

berndgassmann: I don't know why the CI build failed on some smoke tests, but I don't believe that this is actually related to this change, might be error that occurs from here to then. So maybe we give the CI another chance with the mentioned changes.

MarcelPiNacy-CVC (approved):

PR #9533, opened 2026-01-26
Quality of Life Improvements to IAI Integration

KieranRatcliffeInvertedAI:
#### Description

The changes in this PR fall into 4 different categories:

1. Log/Birdview GIF Addition
Added the GIF generator from the IAI Logger tool as a default when setting the --iai-log flag. Visualizing the paths of the vehicles from a top-down perspective is useful to pair with the log for analysis.

2. IAI Performance Features Integration
Implemented the IAI Waypoint Manager for use with the newest and highest performing IAI models. This replaced the old waypoint finding functions in this example.

3. Ego Co-Simulation Framework
Added some functions that will more easily allow integration of ego-vehicles from any source (SIL, logs, etc.) and seamlessly integrate with the script. This will be useful since it is assumed most users who would be interested in this example script would like to simulate their own ego vehicle among the IAI and Carla agents.

4. Carla Video Capture Framework
A simple but flexible framework is added to add different types of camera sensors to the simulation. Following the format of the framework, other types of cameras can be added, whether static (e.g. CCTV cameras) or attached to an agent of the users choosing (e.g. dash camera). The camera capture framework will save each frame as an image to a default directory (or a given path), compile these images into a video, then delete the images for disk-space efficiency. 

Fixes #  

  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** UE4 4.26.2

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9533)

### DIFF
--- a/PythonAPI/examples/.gitignore
+++ b/PythonAPI/examples/.gitignore
@@ -1,1 +1,3 @@
 cache
+*.json
+*.gif
--- a/PythonAPI/examples/invertedai_traffic.py
+++ b/PythonAPI/examples/invertedai_traffic.py
@@ -1077,8 +1077,7 @@ def main():
 if __name__ == '__main__':
 
     try:
-        for _ in range(100):
-            main()
+        main()
     except KeyboardInterrupt:
         pass
     finally:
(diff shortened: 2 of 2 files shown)


Blyron:
Can we have also a PR for ue4-dev branch?
Thanks for contribution, we really appreciate it.

PR #9552, opened 2026-02-13
Fix geo latitude inversion unreal coordinates

### LINKED ISSUE #9530: CRITICAL❗Incorrect geocoordinates and wrong north orientation in Digital Twin map generation (9.16 - ue4-dev branch)
Setup
CARLA version: 0.9.16 (Digital Twin Map Generator)
Branch: ue4-dev
Platform: Ubuntu Linux 20.04 / 22.04
Python version: 3.8 / 3.10
GPU: NVIDIA (e.g. RTX / GTX)
GPU Drivers: NVIDIA proprietary drivers (e.g. 535.xx)

In the ue4-dev branch, the geocoordinates generated in the CARLA Digital Twin map are incorrect and the map is not oriented with the correct geographic north.

After importing real-world geographic data (e.g. IGN data, WGS84, or projected coordinate systems such as Lambert93), the resulting CARLA map shows a significant mismatch between real-world positions and CARLA coordinates.
Additionally, the map orientation is rotated: real north does not correspond to CARLA north, leading to axes misalignment (e.g. real north appears as west or south in CARLA).

This prevents accurate alignment of real GPS trajectories or satellite imagery with the CARLA simulation.

Steps to reproduce

Use the Digital Twin Map Generator from the ue4-dev branch.
Generate a map from real-world geographic data (IGN / WGS84 / projected CRS).
Define the <geoReference> in the generated OpenDRIVE file.
Load the map into CARLA.
Compare known real-world GPS points or satellite imagery with CARLA actor positions.
Observe position offsets and incorrect north orientation.

Expected behavior

Generated CARLA maps should correctly match real-world geocoordinates.
CARLA north should be aligned with true geographic north.
GPS trajectories converted from WGS84 or projected CRS should overlay correctly on the CARLA map without rotation or offset.

csonthomisi:
#### Description

This PR fixes a bug in the geo-coordinate conversion system where latitude values changed incorrectly (inverted sign) when moving north/south in maps using Transverse Mercator (tmerc) or Universal Transverse Mercator (UTM) projections without an explicit `<offset>` tag in the OpenDRIVE file.

**Root Cause:**
Unreal Engine uses a left-handed coordinate system while geographic projections (Transverse Mercator, UTM) assume a right-handed system. The projection inverse transforms did not account for this handedness mismatch, causing the northing (Y) component to have an inverted sign relative to latitude changes.

**Changes Made:**

1. **GeoProjection.cpp**:
   - Modified `TransformToGeoLocationTransverseMercator()` to negate the Y-coordinate: `double y = -((location.y - p.y_0) / p.k);`
   - Modified `TransformToGeoLocationUniversalTransverseMercator()` to negate the Y-coordinate: `double y = -((location.y - y_0) / k);`
   - Added comments documenting the fix and the reasoning (Unreal left-handed coordinate system correction, applied 2026-02-13)

**Behavior Change:**
- **Before:** Moving an actor +Y in CARLA local coordinates produced decreasing latitude (incorrect)
- **After:** Moving an actor +Y in CARLA local coordinates produces increasing latitude (correct)
- Map origin (local 0,0) continues to map correctly to the expected geolocation

  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** 4.26 (CARLA ue4-dev branch)
  * **Test method:** GNSS sensor output validation; confirmed latitude increases when actor moves north

#### Possible Drawbacks

- Maps that explicitly use an `<offset>` transform in OpenDRIVE may need validation to ensure the combined effect (offset negation + projection negation) is correct. However, the offset's OffsetTransform already applies a Y negation, so this change complements existing behavior.
- Any downstream code that relied on the incorrect (inverted) latitude behavior will need adjustment, but the fix aligns the system with correct geographic coordinate semantics.

#### Files Changed

1. `LibCarla/source/carla/geom/GeoProjection.cpp`

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9552)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Latest Changes
+ * Fixed North/South latitude inversion in geo-coordinate conversion for Transverse Mercator and UTM projections
  * Fix possible RPC Server deadlock on shutdown
  * Change foonathan memory vendor branch to v1.3.1.
  * Improved the way the TrafficManager controls large vehicles at junctions, reducing the frequency of collisions with other elements in the simulation.
--- a/LibCarla/source/carla/geom/GeoProjection.cpp
+++ b/LibCarla/source/carla/geom/GeoProjection.cpp
@@ -220,7 +220,11 @@ namespace geom {
         double e6 = e4 * e2;
 
         double x = (location.x - p.x_0) / p.k;
-        double y = (location.y - p.y_0) / p.k;
+        // Fixed: negate Y to account for Unreal left-handed coordinate system
+        // Reason: Unreal uses a left-handed coordinate system; negating the northing
+        // corrects latitude sign for maps without an explicit <offset> in OpenDRIVE.
+        // Applied: 2026-02-13
+        double y = -((location.y - p.y_0) / p.k);
 
         double M = a * ((1.0 - e2 / 4.0 - 3.0 * e4 / 64.0 - 5.0 * e6 / 256.0) * lat_0
             - (3.0 * e2 / 8.0 + 3.0 * e4 / 32.0 + 45.0 * e6 / 1024.0) * std::sin(2.0 * lat_0)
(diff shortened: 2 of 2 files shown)


csonthomihaly:
It seems that this is maybe not the best place and solution for this problem. I will try to implement a better solution next week. Until that, this should not be merged! Thanks for your patience!

berndgassmann:
Even the OffsetTransform didn't do all things in a correct manner, so I tried to solve it finally by using an extra class doing the conversion. You might want to test the proposed solution in #9590 if it works for you...

PR #9556, opened 2026-02-19
Added recorder to weather + simultaneous record and replay

glopezdiest:
#### Description

This PR has two additions to the recorder / replayer.

First of all, the weather is now added to the log file, and can be replayed. A "replay_weather" argument has been added to "client.replay_file()" to replay the weather saved in the recorder. It is False by default to match previous behavior.

Secondly, it is not possible to record and replay at the same time. Previously starting a recorder automatically stopped the replayer. Now a new argument "stop_replayer" has been added to "client.start_recorder()", to decide whether to stop it or not. It is set to True by default.

Also added the "offset" and "map_override" arguments to the start recorder docs, as they were missing

  * **Python version(s):** 3.11

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9556)


LuisPovedaCano (approved):
Good Job! 👍

PR #9580, opened 2026-03-10
Make traffic manager more robust on varying delta times

berndgassmann:
The traffic manager PID controller used fixed delta time values. In case the simulation is not running at exact the fixed_delta_seconds of 0.05, the PID controller over-reacted or under-reacted massively. Using the actual delta time between the control steps improved the PID controller behaviour on different timings.
To ensure that the controller is not degenerating because of the inverse of the delta time on time steps close to zero (which should rarely happen in practice -- if at all), just the previous signals are returned.

#### Description

Fixes #  

  * **Platform(s):** Ubuntu 22.04, 24.04
  * **Python version(s):** 3.10 , 3.12

#### Possible Drawbacks

The question would be: was there maybe a good reason to fix the delta time values for the PID controller back in time to the constants which we don't see? For us, the changes work even with running CARLA e.g. at 10Hz instead of 20Hz.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9580)


glopezdiest:
Hey @berndgassmann. I don't think there was a reason why the delta time was fixed to 0.05, not have any reason to believe this breaks anything. We'll test it ASAP

glopezdiest (approved):

PR #9590, opened 2026-03-16
GeoProjection: Fix the handling of left handed Unreal location values

### LINKED ISSUE #9389: Carla GPS location mismatch for Carla 0.9.15 and 0.9.16
**Setup**
  * CARLA version: 0.9.16
  * Platform: linux
  * Python version: 3.11
  * GPU: RTX A1000
  * GPU Drivers: CUDA12

**Describe the bug**
I am using bench2drive leaderboard. For exactly the same code, the Carla 0.9.15 get gps location at 'gps': (8909, array([ 0.00293298,  0.00160249, -0.05002071])), while  Carla 0.9.16 get the gps location at 'gps': 'gps': (12397, array([-0.00293595,  0.00160185, -0.05002071])). The actual game time at this frame is exactly the same. There is a plus minus sign difference of the first axis in the gps data. Can we fix this bug?

**Steps to reproduce**
Just run simlingo repo code on two 0.9.15 and 0.9.16 version of Carla can produce this problem. https://github.com/RenzKa/simlingo

### LINKED ISSUE #9530: CRITICAL❗Incorrect geocoordinates and wrong north orientation in Digital Twin map generation (9.16 - ue4-dev branch)

berndgassmann:
Use a helper class LocationRightHanded to allow the implemenation of the GeoProjections to perform their projections in a right handed coordinate frame.

The explicit handling using LocationRightHanded revealed error in OffsetTransform: result Location is now transformed back to Unreal coordinates and Rotation is fixed.

Allow OpenDRIVE offsets for any type of projection

Fix UniversalTransverseMercatorParams::operator!=()

This fix should contain the fixes which #9552 tried to tackle by using a more general solution. Especially this solution only applies the left-handed/right-handed conversion to the carla::geom::Location values, because all offset values provided by the OpenDRIVE map itself have to be considered to be in right-handed notation. 

#### Description

Fixes #9530
Fixes #9389
And implements #9552 in a more complete manner.

   

  * **Platform(s):** ... Ubunut 24.04
  * **Python version(s):** ...3.12.

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9590)

### DIFF
--- a/PythonAPI/carla/source/libcarla/Geom.cpp
+++ b/PythonAPI/carla/source/libcarla/Geom.cpp
@@ -317,12 +317,34 @@ void export_geom() {
   class_<cg::TransverseMercatorParams>("GeoProjectionTM")
     .def(init<double, double, double, double, double, cg::Ellipsoid>(
       (arg("lat_0")=0.0, arg("lon_0")=0.0, arg("k")=1.0, arg("x_0")=0.0, arg("y_0")=0.0, arg("ellps")=cg::Ellipsoid())))
+    .def(init<double, double, double, double, double, cg::Ellipsoid, boost::optional<cg::OffsetTransform>>(
+      (arg("lat_0")=0.0, arg("lon_0")=0.0, arg("k")=1.0, arg("x_0")=0.0, arg("y_0")=0.0, arg("ellps")=cg::Ellipsoid(), arg("offset")=boost::python::object())))
     .def_readwrite("lat_0", &cg::TransverseMercatorParams::lat_0)
     .def_readwrite("lon_0", &cg::TransverseMercatorParams::lon_0)
     .def_readwrite("k", &cg::TransverseMercatorParams::k)
     .def_readwrite("x_0", &cg::TransverseMercatorParams::x_0)
     .def_readwrite("y_0", &cg::TransverseMercatorParams::y_0)
     .def_readwrite("ellps", &cg::TransverseMercatorParams::ellps)
+    .add_property("offset",
+        +[](const cg::TransverseMercatorParams &self) {
+          return OptionalToPythonObject(self.offset); //returns None or OffsetTransform 
+        },
+         +[](cg::TransverseMercatorParams& self, object value) {
+            if (value.is_none()) {
+                self.offset = boost::none;
+            } else {
+                extract<cg::OffsetTransform> ex(value);
+                if (!ex.check()) {
+                    PyErr_SetString(
+                        PyExc_TypeError,
+                        "offset must be OffsetTransform or None"
+                    );
+                    throw_error_already_set();
+                }
+                self.offset = ex();
+            }
+        }
+    )
     .def("__eq__", &cg::TransverseMercatorParams::operator==)
     .def("__ne__", &cg::TransverseMercatorParams::operator!=)
   ;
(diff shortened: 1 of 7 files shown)


PythonAPI/carla/source/libcarla/Geom.cpp
```diff
@@ -359,21 +381,64 @@ void export_geom() {
 
   class_<cg::WebMercatorParams>("GeoProjectionWebMerc")
     .def(init<cg::Ellipsoid>((arg("ellps")=cg::Ellipsoid())))
```
glopezdiest: [Applied to all projections]

There are two inits, one with the offset transfom and one without it. I'd merge them into one.

PythonAPI/carla/source/libcarla/Geom.cpp
```diff
@@ -359,21 +381,64 @@ void export_geom() {
...
+    .def(init<cg::Ellipsoid, boost::optional<cg::OffsetTransform>>((arg("ellps")=cg::Ellipsoid(), arg("offset")=boost::python::object())))
     .def_readwrite("ellps", &cg::WebMercatorParams::ellps)
+    .add_property("offset",
```
glopezdiest: [Applied to all projections]

Is this really the best way to define the offset? It feels convoluted. Why not just pass the default offset? Or None? 

And if that isn't possible, at least try to merge the four offsets into one template to avoid repeated code

glopezdiest:
Hey @berndgassmann, thank you for the PR. The implementation is correct, and it fixes the issue with the offset transform.

I've left some comments regarding the exposure to the PythonAPI. The code itself works fine but It'd be great if a couple of things were changed.

berndgassmann:
Sure I can change...
did you click on publishing of the comments yet?
:-)

glopezdiest:
There, I though I had published them, sorry

PythonAPI/carla/source/libcarla/Geom.cpp
berndgassmann: Now reduced to one single init. Had to refactor the boost optional handling to make the default parameter as well as the read/write access working.

PythonAPI/carla/source/libcarla/Geom.cpp
berndgassmann: Now the code duplication is gone and delegated to some general template functions OptionalFromPythonConverter, OptionalToPythonConverter that can be used for other boost optional python access.

Blyron (approved):

PR #9602, opened 2026-03-24
Fixed sudo issue with UE5 linux build

glopezdiest:
#### Description

The way sudo is used when installing CARLA makes it so that the Build folder and UnrealEngine5 (if installed through the CarlaSetup.sh), have root user. As such, the rest of the pipeline fails due to denied permissions.

This PR changes the sudo to only the install prerequisites, where it is actually used. A `sudo -v` is used to set the password so that the non interactive pipeline still works as intended

  * **Platform(s):** ubuntu 22.04
  * **Python version(s):** 3.11
  * **Unreal Engine version(s):** CARLA's UE5

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9602)

### DIFF
--- a/Docs/build_linux_ue5.md
+++ b/Docs/build_linux_ue5.md
@@ -30,27 +30,21 @@ The setup script will prompt you for your sudo password, in order to install the
 
 __Building in Linux unattended__:
 
-If you want to run the setup script unattended, your git credentials need to be stored in an environment variable. Add your github credentials to your `.bashrc` file:
+If you want to run the setup script unattended, store your sudo password and your git credentials in an environment variable.
 
 ```sh
-export GIT_LOCAL_CREDENTIALS=username@github_token
+sudo -v   # Used to cache the password for 15 minutes
+export GIT_LOCAL_CREDENTIALS=github_username@github_token 
 ```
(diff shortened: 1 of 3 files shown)


Blyron (approved):

Docs/build_linux_ue5.md
```diff
@@ -30,28 +30,22 @@ The setup script will prompt you for your sudo password, in order to install the
 
 __Building in Linux unattended__:
 
-If you want to run the setup script unattended, your git credentials need to be stored in an environment variable. Add your github credentials to your `.bashrc` file:
+If you want to run the setup script unattended, store your sudo password and your git credentials in an environment variable.
```
jorge-kabuto, line 33: Just to clarify, is this something that is usually done or most user will end up not storing the sudo password and just launching interactively? Basically: is it a security risk or am I being paranoid?

glopezdiest: By "storing your password" it means that sudo remembers it. CARLA is not storing it anywhere. As for the vulnerability, that is the same possible one as when someone does `sudo apt-get update`.

PR #9645, opened 2026-04-05
feat(LibCarla/ros2): [7/7] remove legacy FastDDS-generated type files

JArmandoAnaya:
#### Description

Part of the DDS middleware decoupling series (issue #9294). Blocked on #9644.

Removes 125 legacy FastDDS-generated type files from
`LibCarla/source/carla/ros2/types/`:

- 31 type definition pairs (`{TypeName}.h` + `{TypeName}.cpp`)
- 31 PubSubType pairs (`{TypeName}PubSubTypes.h` + `{TypeName}PubSubTypes.cpp`)
- `FastDDSConversions.h`

These files were auto-generated by `fastddsgen` from IDL definitions and are
no longer referenced anywhere in the codebase. `GenericCdrPubSubType<T>`
(introduced in #9643) replaces all 31 hand-generated `*PubSubType` classes
using `CdrSerialization.h` for direct CDR serialization of
`carla::ros2::msg::*` POD structs.

Both `cmake/fast_dds/CMakeLists.txt` and `cmake/cyclone_dds/CMakeLists.txt`
collect type files via `file(GLOB)` patterns (`types/*.h`, `types/*.cpp`),
so no CMake changes are needed. The globs self-adjust on the next configure pass.

After this PR, `LibCarla/source/carla/ros2/types/` contains only:
- `CdrSerialization.h`
- `CdrTopicInfo.h`
- `msg/` (31 vendor-agnostic POD struct headers)

### PR series

This PR is part of the DDS middleware decoupling series ([#9294](https://github.com/carla-simulator/carla/issues/9294)). Each PR in the chain inherits the commits of all prior PRs. The **Commits** column lists only the commits introduced by that PR, and the **Files** column counts only the files changed by those new commits.

| # | PR | Title | Branch | Commits | Files |
|---|----|-------|--------|---------|-------|
| 1/7 | [#9608](https://github.com/carla-simulator/carla/pull/9608) | DDS middleware abstraction layer | `feature/dds-middleware-abstraction-layer` | `0ab35ee6` | 10 |
| 2/7 | [#9612](https://github.com/carla-simulator/carla/pull/9612) | POD message types, FastDDS conversions, and TypeMap specializations | `feature/ros2-pod-types-and-fastdds-conversions` | `e665d2af` | 39 |
[...]

PR #9644, opened 2026-04-05
feat(LibCarla/ros2): [6/7] add CycloneDDS CDR middleware, UE4 runtime selection, and correctness fixes

JesusAnaya:
#### Description

Part of the DDS middleware decoupling series (issue #9294). Blocked on #9643.

This PR completes the CycloneDDS CDR middleware implementation, wires it into the UE4
runtime, and fixes several correctness issues found during integration testing.

### PR series

This PR is part of the DDS middleware decoupling series ([#9294](https://github.com/carla-simulator/carla/issues/9294)). Each PR in the chain inherits the commits of all prior PRs. The **Commits** column lists only the commits introduced by that PR, and the **Files** column counts only the files changed by those new commits.

| # | PR | Title | Branch | Commits | Files |
|---|----|-------|--------|---------|-------|
| 1/7 | [#9608](https://github.com/carla-simulator/carla/pull/9608) | DDS middleware abstraction layer | `feature/dds-middleware-abstraction-layer` | `0ab35ee6` | 10 |
| 2/7 | [#9612](https://github.com/carla-simulator/carla/pull/9612) | POD message types, FastDDS conversions, and TypeMap specializations | `feature/ros2-pod-types-and-fastdds-conversions` | `e665d2af` | 39 |
| 3/7 | [#9619](https://github.com/carla-simulator/carla/pull/9619) | Migrate publishers and subscribers to POD message types | `feature/ros2-publishers-pod-types-migration` | `98b6f7aa` | 27 |
| 4/7 | [#9620](https://github.com/carla-simulator/carla/pull/9620) | CycloneDDS enum, factory, stubs, and tests | `feature/ros2-cyclonedds-enum-factory` | `2a4a8db3` | 5 |
| 5/7 | [#9643](https://github.com/carla-simulator/carla/pull/9643) | Unified CDR serialization + `GenericCdrPubSubType` | `feature/ros2-cdr-serialization` | `5a9dd3b4`, `0a9ca585` | 8 |
| **6/7** | **[#9644](https://github.com/carla-simulator/carla/pull/9644)** | **CycloneDDS CDR middleware, UE4 runtime selection, and correctness fixes (this PR)** | `feature/ros2-cyclonedds-cdr-middleware` | `d3f2a45b`, `2d80046b` | **26** |
[...]

PR #9666, opened 2026-04-10
fix(LibCarla/cmake): install ros2/dds/ headers in server cmake for Windows builds

JesusAnaya:
#### Description

PR #9644 added `#include "carla/ros2/dds/DDSMiddleware.h"` to `ROS2.h` for the new `Enable(bool, DDSMiddleware)` signature, but did not update the server cmake header install list. This breaks the Windows CI build with:

```
carla/ros2/ROS2.h(13): fatal error C1083: cannot open include file:
    'carla/ros2/dds/DDSMiddleware.h': No such file or directory
```

**Why this only affects Windows:** LibCarla headers are installed to `CarlaDependencies/include/` via two different cmake paths:

- **Linux** (`make LibCarla ARGS="--ros2"`): uses the `ros2` build type, which delegates header installation to `cmake/fast_dds/CMakeLists.txt`. That file already installs all `ros2/` subdirectories, including `ros2/dds/`. Working correctly.
- **Windows** (`make LibCarla`, no `--ros2`): uses the `Server` build type, which runs `cmake/server/CMakeLists.txt`. This file contains a manually maintained `foreach` loop that lists every subdirectory whose headers should be installed. It listed `"ros2/"` but not `"ros2/dds/"`.

The UE4 plugin includes `ROS2.h` unconditionally (not behind `#ifdef WITH_ROS2`) from `CarlaEngine.h`, `CarlaEpisode.h`, `ActorROS2Handler.h`, and several sensor `.cpp` files. Before #9644, `ROS2.h` had no subdirectory dependencies, so the missing entry was not a problem.

**Fix:** add `"ros2/dds/"` to the server cmake `foreach` install loop, following the same pattern already used for other nested subdirectories (`"opendrive/parser/"`, `"road/element/"`, `"streaming/detail/tcp/"`, etc.).

**Why this must merge before #9645:** PR [#9645](https://github.com/carla-simulator/carla/pull/9645) (delete legacy FastDDS-generated types) cannot pass Windows CI until this header installation gap is fixed. The error is caused by code already merged in #9644, not by #9645 itself.

- **Platform(s):** Ubuntu 22.04
- **Python version(s):** 3.10-3.12
- **Unreal Engine version(s):** 4.26

Tests:

```sh
make LibCarla ARGS="--ros2"    # builds server, client, ros2 targets
make check.LibCarla            # 114/114 server + 56/56 client tests pass
```

Header installation verified:

```
CarlaDependencies/include/carla/ros2/dds/DDSMiddleware.h  (present after build)
```

#### Possible Drawbacks

None. This is a pure addition to the CMake install list. It cannot affect existing installs, only adds previously missing headers. The Linux build path is unaffected because it uses a separate CMake target (`cmake/fast_dds/CMakeLists.txt`) that already installs these headers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9666)

### DIFF
--- a/LibCarla/cmake/server/CMakeLists.txt
+++ b/LibCarla/cmake/server/CMakeLists.txt
@@ -24,7 +24,7 @@ foreach(dir ""
             "sensor/" "sensor/data/" "sensor/s11n/"
             "streaming/" "streaming/detail/" "streaming/detail/tcp/" "streaming/low_level/"
             "multigpu/"
-            "ros2/")
+            "ros2/" "ros2/dds/")
 
   file(GLOB headers "${libcarla_source_path}/carla/${dir}*.h")
   install(FILES ${headers} DESTINATION include/carla/${dir})


JesusAnaya:
@LuisPovedaCano Hello, I'm creating this PR because I noticed that since PR [#9644](https://github.com/carla-simulator/carla/pull/9644), there has been an issue building the Windows version of Carla. The issue looks like a minor problem, and a small change can resolve it. It is probably better to check whether this code change fixes the Windows build in GitHub CI before merging the PR https://github.com/carla-simulator/carla/pull/9645 to verify everything is correct.

LuisPovedaCano (approved):

PR #9487, opened 2025-12-17
Upgrade boost to 1.89.0

MarcelPiNacy-CVC:
#### Description

This PR upgrades boost to 1.89.0

#### Where has this been tested?

  * **Platform(s):** Windows 10
  * **Python version(s):** 3.13
  * **Unreal Engine version(s):** UE5-CARLA

#### Possible Drawbacks

Needs further testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9487)

PR #9670, opened 2026-04-14
fix(PythonAPI/build): add NumPy 2 compatibility via Boost 1.90 upgrade 

### LINKED ISSUE #9365: Support Numpy 2
We are unable to build the pythonapi client when numpy 2 is in our environment

An upgrade of boost.python or migration to [pybind](https://github.com/carla-simulator/carla/issues/3731) such that numpy 2.x is supported

**Describe alternatives you've considered**
continue using the outdated numpy 1

**Additional context**
[relevant check](https://github.com/carla-simulator/carla/blob/ue5-dev/PythonAPI/CMakeLists.txt#L12)
```
if ("${NUMPY_VERSION_STRING}" VERSION_GREATER "2.0.0")
  carla_error ("Unsupported Numpy version, please downgrade to a version prior to numpy 2.")
endif ()
```

JArmandoAnaya:
#### Description

Upgrades Boost from 1.84.0 to 1.90.0 and, as a direct consequence, makes CARLA's
PythonAPI compatible with both NumPy 1.x (>= 1.18.4) and NumPy 2.x. NumPy 1.x
support is preserved, not dropped.

We first landed the upgrade at Boost 1.89.0, which is the minimum version that
ships the upstream NumPy 2 C ABI fix
([boostorg/python#432](https://github.com/boostorg/python/pull/432)) natively.
Moving from 1.89.0 to 1.90.0 was a small follow-up: 1.90 only added
`[[deprecated]]` annotations to `boost::asio::deadline_timer`, which we migrated
to `boost::asio::steady_timer`. Going straight to the newest stable release
avoids carrying a known-deprecated API into the tree.

---

#### Improvements

- Boost upgraded from **1.84.0 -> 1.90.0** on Linux and Windows. Old cached
  `boost-1.84.0-install/` trees are picked up automatically via the updated
  `BOOST_VERSION` and SHA256 pins in `Util/BuildTools/Setup.sh`,
  `Util/BuildTools/Setup.bat`, `Util/BuildTools/BuildOSMRenderer.bat`, and
  `Util/InstallersWin/install_boost.bat`.
- LibCarla networking migrated from `boost::asio::deadline_timer` to
  `boost::asio::steady_timer` in `Primary`, `Secondary`, `Client`, and
  `ServerSession`. No deprecated Asio API remains in the tree.
- Python requirements widened from `numpy<2.0.0` to `numpy>=1.18.4` in
  `PythonAPI/examples/requirements.txt`, `PythonAPI/carla/agents/requirements.txt`,
  and `PythonAPI/test/requirements.txt`.
- Docker CI test requirements (`Util/Docker/requirements/3.11/test.txt` and
  `3.12/test.txt`) pinned to `numpy>=2.0.0` to exercise NumPy 2.x on CI.
- Vendored patch `Util/Patches/boost-1.84.0/dtype.cpp.patch` and its helpers
  (`verify_patch.cpp`, `verify_patch.sh`) are deleted. Boost ships the fix now.

---

#### Fixes

- **NumPy 2 Python-level removals.** `np.bool` replaced with `np.bool_` in
  `PythonAPI/examples/V2XDemo.py`, `manual_control_chrono.py`, and
  `manual_control_carsim.py` (DVS event dtype). `np.matrix(np.identity(4))`
  replaced with `np.array(np.identity(4))` in `client_bounding_boxes.py`, safe
  because the only consumer is `np.dot()`.

- **NumPy 2 C ABI ([NEP 52](https://numpy.org/neps/nep-0052-python-api-cleanup.html)).**
  Boost 1.89+ makes `libboost_numpy` build against both NumPy 1.x and 2.x
  headers without any vendored patches (`PyArray_Descr` is now opaque).

- **Boost 1.84 -> 1.90 API call-site updates** in LibCarla and the CarlaTools
  UE4 plugin (each removal was introduced between 1.85 and 1.90):

| File | Change |
|------|--------|
| `LibCarla/source/carla/ThreadPool.h` | `io_context::work` -> `executor_work_guard` |
| `LibCarla/source/carla/multigpu/listener.cpp` | `io_context::reset()` -> `restart()` |
| `LibCarla/source/carla/multigpu/router.cpp`, `secondary.cpp` | `address::from_string()` -> `make_address()` |
| `LibCarla/source/carla/rpc/Server.h` | `reset()` -> `restart()` |
| `LibCarla/source/carla/streaming/EndPoint.h` | `resolver::query`/`iterator` -> range-based `resolve()` |
| `LibCarla/source/carla/FileSystem.cpp` | added `#include <boost/filesystem/directory.hpp>` (split in 1.85) |
| `Unreal/.../MapPreviewUserWidget.{h,cpp}` | `io_service` -> `io_context`<br>`buffer_cast<T>()` -> `static_cast` via `streambuf::data().data()`<br>MSVC C4459 pragma guard |

- **`steady_timer` default-construction trap.** `deadline_timer` defaulted to
  `pos_infin`, but `steady_timer` default-constructs with `expiry() == time_point::min()`
  (in the past). `Primary::StartTimer` and `ServerSession::StartTimer` guard on
  `expiry() <= now()` to detect timeouts, so the default would fire `Close()`
  before the socket is set up. `_deadline` is now explicitly initialised with
  `std::chrono::steady_clock::time_point::max()` in both constructors.

- **Call-site API migrations** for the timer rewrite:
  `expires_from_now(d)` -> `expires_after(d)`, `expires_at()` -> `expiry()`,
  `deadline_timer::traits_type::now()` -> `std::chrono::steady_clock::now()`,
  `boost::posix_time` durations -> `std::chrono` durations.

---

#### Added testing

New regression guards added under `LibCarla/source/test/common/` and
`PythonAPI/test/unit/`:

| File | Coverage |
|------|----------|
| `test_boost_deadline_timer.cpp`<br>(test class `BoostAsioTimer`) | 6 tests: `TimerFiresWithinReasonableTime`, `ExpiresAtNowComparison`, `ChronoDurationComposition`,<br>`InitializedWithMaxIsInDistantFuture` (pins the `time_point::max()` fix),<br>`DefaultConstructedExpiryIsInPast` (documents the trap),<br>`CancelTriggersAbortedError` (covers the `CloseNow()` path) |
| `test_boost_rtree.cpp` | 4 tests guarding `geometry::index::rtree` (used by TrafficManager) |
| `test_boost_version.py` | Asserts the Boost version pinned in `Setup.sh` is current (1.90.0) and no stale `boost-1.84.0` patch dir or link exists |
| `test_numpy_compat.py` | 15 tests: Python alias regressions, DVS pipeline, carla C extension load under NumPy 1.x and 2.x |
| `run_numpy_compat_matrix.sh` | Matrix runner: creates venvs for NumPy 1.x and 2.x, installs the wheel, runs `test_numpy_compat.py` in each |

---

#### Commits

1. `d4151ed05` fix(PythonAPI/examples): replace deprecated NumPy aliases
2. `69df4d2a1` fix(build): upgrade Boost to 1.89.0 for NumPy 2 ABI compatibility
3. `a5e4ce0d3` docs(CHANGELOG): note NumPy 2 compatibility
4. `5e68bef2c` fix(build): bump Boost 1.89 -> 1.90 and migrate timers to steady_timer

Fixes #9365

---

- **Platform(s):** Ubuntu 22.04
- **Python version(s):** 3.8, 3.9, 3.10, 3.11, 3.12
- **Unreal Engine version(s):** UE4.26

Results:

- `make check.LibCarla` passes in both debug and release on the server and
  client suites, including all 6 `BoostAsioTimer.*` and 4 `BoostRtree.*` tests.
- `make check.PythonAPI` passes, including the full `test_boost_version.py`
  set (static Boost version pin + no stale 1.84.0 link).
- `python3 -m nose2 -v --start-dir PythonAPI/test/unit test_numpy_compat` on
  Python 3.12 + NumPy 2.4.4 with the carla wheel installed: **15/15 pass.**
- `bash PythonAPI/test/unit/run_numpy_compat_matrix.sh python3.12`:
  NumPy 1.26.4 venv 15/15 pass, NumPy 2.4.4 venv 15/15 pass. The 2.x log
  confirms end-to-end load:
  `carla C extension OK with NumPy 2.4.4, NumPy 2.x (NEP 52, opaque PyArray_Descr)`.
- CI-equivalent package build: `make package ARGS="--python-version=3.10,3.11,3.12 --ros2 --chrono --no-zip --target-wheel-platform=manylinux_2_31_x86_64"`: SUCCESS.
- Smoke tests (Python 3.12, port 3654): all pass.

---

#### Possible Drawbacks

- Users with a cached `Build/boost-1.84.0-c10-install/` or
  `Build/boost-1.84.0-install/` tree from a previous checkout need to remove it
  before building, so the 1.90.0 source is fetched and compiled. `make clean`
  is sufficient.
- Windows: `install_boost.bat` SHA256 has been updated to the 1.90.0 source-zip
  digest. Cached `boost-1.84.0-install/` directories on Windows build agents
  must likewise be removed. The Windows CI path does not build ROS2, so none of
  the Asio/Filesystem call-site changes affect the Windows CI build.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9670)


Blyron:
Why should we merge this?

JesusAnaya:
Fair enough. I expect NumPy 1.x to reach end of life, and downstream libraries are already moving to require NumPy 2 (e.g., opencv-python >= 4.12). Having this compatibility in place before it becomes urgent seems better than scrambling later, but I understand if the priority isn't there yet.

Blyron:
What about using updated version of boost instead of patching it? I do not like patches.

JesusAnaya:
Good point. I went with a patch to avoid a full Boost version bump, which could be quite invasive in this refactoring, since we only needed to change a few things from Boost. But if you’d prefer to upgrade Boost instead, I can look into it and send a new commit.

Blyron:
Look at this https://github.com/carla-simulator/carla/pull/9487/changes#diff-e69ecb8d585c244655338ab3d5b078f10bca6c28d00bf4357e064baf193d0e45

is a PR from a former CARLA collaborator.

JesusAnaya:
Ready! I upgraded Boost to 1.90. I used the example from @MarcelPiNacy-CVC as the first try. But, after that, migrating from 1.89 to 1.90 required only a few extra steps, so I did it in a second commit.

LuisPovedaCano (changes requested):
Great work on this PR!
The API migrations are all correct, the `time_point::max()` initialization fix is a subtle but important catch, and the test coverage is thorough.
Thanks for the effort put into this!

One small observation: `test_no_boost_189_in_linked_libs` in `test_boost_version.py` guards against accidentally linking a stale `boost-1.89.0` install (the intermediate version used during development of this PR), but a user upgrading directly from the previous `1.84.0` baseline would slip past it undetected.
Since that's arguably the more common stale-cache scenario for most contributors, it might be worth adding a similar check for `boost-1.84.0` there too or widening the assertion to something like "must link against boost-1.90.0" rather than "must not link against boost-1.89.0"

JesusAnaya:
Hello! Thanks for the code review. I made the update.

LuisPovedaCano (approved):
👍

PR #9678, opened 2026-04-14
fix(road): resolve Digital Twin tool crashes and tree misplacement (#9565)

### LINKED ISSUE #9565: Digital Twin Tool generating issue
**Setup**

- CARLA 0.9.16
- OS: Ubuntu 22.04.5 LTS
- Python Version: 3.10.12
- GPU: NVIDIA GeForce RTX 5090
- Driver Version: 580.126.09

**Description**

I am experiencing issues when generating maps using the Digital Twin tool from OSM data.

**Workflow**

1. I create a query in Overpass Turbo using a bounding box of approximately 1.5 km x 1.5 km.
2. I export the result as an .osm file.
3. In CARLA, I open UW_OnRoadMainWidget.
4. I select the option to load OSM from path.
5. I enter the same bounding box coordinates in the CARLA widget that I used in Overpass.
6. The map generation process starts.

**Problem 1 – Failure in metropolitan areas**

In dense metropolitan areas such as:

- Bilbao
- Barcelona

(Spanish cities)

The Digital Twin generation fails completely. The process either:

- Crashes,
- Freezes,
- Or does not generate the map correctly.

However, using the same workflow and similar bounding box size in rural areas works correctly.

**Problem 2 – Incorrect object placement in rural areas**

In rural areas where the map generation succeeds:

- Roads are generated correctly.
- Buildings are generated correctly.

However:

- A large number of trees or static objects are spawned incorrectly.
- Trees appear in the middle of roads.
- Obstacles are generated inside traffic lanes.
- Some objects block vehicle movement.

This makes the generated map unusable for traffic simulation.

**Questions**

- Has anyone experienced Digital Twin generation failures in large metropolitan areas?
- Is there a known limitation related to urban density or mesh complexity?
- Is there a recommended OSM filtering strategy before importing?
- Is there a configuration parameter to prevent vegetation or props from spawning inside road geometries?

I can provide screenshots, logs and the OSM files if needed.

im-hashim:
## Summary

Fixes three bugs in `Map::GetTreesTransform` and `Map::FilterRoadsByPosition` reported in #9565. The Digital Twin Tool was crashing in dense metropolitan areas (Bilbao, Barcelona) and placing vegetation inside driving lanes on rural maps.

### Bug 1 — Null-pointer crash on roads without `maxspeed` tag (metropolitan areas)

`GetInfo<RoadInfoSpeed>()` returns `nullptr` for roads that carry no explicit `maxspeed` OSM tag. Urban streets in Spain (and most of Europe) rely on the country's default speed limit and don't tag `maxspeed` individually, so `roadinfo` is `nullptr` for the majority of urban road segments. The previous code called `roadinfo->GetType()` unconditionally → **SIGSEGV**.

**Fix:** null-check `roadinfo` before dereferencing; fall back to `"Town"` when absent.

### Bug 2 — Trees spawning inside driving lanes (rural maps)

When a lane section has no negative-ID driving lanes (common on road segments whose OSM direction is reversed relative to the OpenDRIVE reference), `min_lane` stays `0`. `GetLane(0)` always succeeds because lane 0 (the OpenDRIVE reference line) is always present. Its polynomial-evaluated width is near-zero but non-zero, passing the `!= 0.0f` guard. `GetCornerPositions` then returns two corners within micrometres of road centre. `MakeUnitVector()` on the resulting near-zero `director` returns `(0,0,0)`, so `treeposition = edges.first − (0,0,0) × dist = edges.first` — placed **on the road surface**.

**Fix (two layers):**
1. `if (outer_lane == 0) continue;` — skip the lane section when no valid outboard driving lane is found.
2. `if (director_squared_length <= EPSILON * EPSILON)` guard before `MakeUnitVector()` as a defensive second layer.

### Bug 3 — One-way streets and complex junctions silently excluded (metropolitan areas)

`FilterRoadsByPosition` queried only lane `−1` (RHT) or `+1` (LHT). One-way urban streets, service roads, and complex junction links frequently lack that exact lane → road silently excluded → missing or incorrect streets in the generated map.

**Fix:** after the primary lane lookup fails, fall back to any non-reference (`id != 0`) driving lane with the smallest absolute ID.

## Files changed

- `LibCarla/source/carla/road/Map.cpp` — `GetTreesTransform` and `FilterRoadsByPosition`
- `LibCarla/source/test/client/test_road_regression_9565.cpp` — new LibCarla GTest regression suite

## Test plan

- [x] LibCarla GTest regression suite (`test_road_regression_9565.cpp`) — 4 tests covering all three bug paths:
  - `filter_roads_includes_positive_only_lanes_9565`
  - `get_trees_transform_no_crash_without_speed_9565`
  - `get_trees_transform_positive_lane_road_gets_trees_9565`
  - `get_trees_transform_trees_outside_road_9565`
- [x] Manual verification: generate a Digital Twin map from a 1.5 km × 1.5 km OSM export of Bilbao or Barcelona (previously crashed) — should complete without crash
- [x] Manual verification: generate a rural map — trees should appear alongside roads, not on driving surfaces

## Related

Closes #9565

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9678)

### DIFF
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -1322,17 +1322,16 @@ namespace road {
                   s_current += distancebetweentrees;
                   continue;
                 }
-                // GetCornerPositions returns (+lateral-offset corner,
-                // -lateral-offset corner). Select the true outer edge from that
-                // ordering based on lane ID sign, then offset farther outward so
-                // trees are always placed away from the driving surface.
+                // GetCornerPositions returns the lane corners in (t_offset + width, t_offset - width) order.
+                // The true outer edge therefore depends on the lane side: for positive lane IDs it is the second corner,
+                // and for negative lane IDs it is the first. Offset farther outward so trees are always placed away from the driving surface.
                 const bool is_positive_lane = (lane->GetId() > 0);
-                const geom::Vector3D positive_lateral_corner = edges.first;
-                const geom::Vector3D negative_lateral_corner = edges.second;
+                const geom::Vector3D first_corner = edges.first;
+                const geom::Vector3D second_corner = edges.second;
                 const geom::Vector3D outer_corner =
-                  is_positive_lane ? negative_lateral_corner : positive_lateral_corner;
+                  is_positive_lane ? second_corner : first_corner;
                 const geom::Vector3D inner_corner =
-                  is_positive_lane ? positive_lateral_corner : negative_lateral_corner;
+                  is_positive_lane ? first_corner : second_corner;
                 const geom::Vector3D outward_direction =
                     (outer_corner - inner_corner).MakeUnitVector();
                 geom::Vector3D treeposition =
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -1312,9 +1312,9 @@ namespace road {
                 // Use double precision for the length check to avoid false
                 // negatives from float cancellation on near-equal corners.
                 const double director_squared_length =
-                    static_cast<double>(director.x) * director.x +
-                    static_cast<double>(director.y) * director.y +
-                    static_cast<double>(director.z) * director.z;
+                  static_cast<double>(director.x) * static_cast<double>(director.x) +
+                  static_cast<double>(director.y) * static_cast<double>(director.y) +
+                  static_cast<double>(director.z) * static_cast<double>(director.z);
                 // Skip degenerate or near-degenerate lane widths; normalising a
                 // near-zero vector produces unstable directions and can place
                 // trees on or very close to the road surface.
@@ -1326,12 +1326,10 @@ namespace road {
                 // The true outer edge therefore depends on the lane side: for positive lane IDs it is the second corner,
                 // and for negative lane IDs it is the first. Offset farther outward so trees are always placed away from the driving surface.
                 const bool is_positive_lane = (lane->GetId() > 0);
-                const geom::Vector3D first_corner = edges.first;
-                const geom::Vector3D second_corner = edges.second;
                 const geom::Vector3D outer_corner =
-                  is_positive_lane ? second_corner : first_corner;
+                  is_positive_lane ? edges.second : edges.first;
                 const geom::Vector3D inner_corner =
-                  is_positive_lane ? first_corner : second_corner;
+                  is_positive_lane ? edges.first : edges.second;
                 const geom::Vector3D outward_direction =
                     (outer_corner - inner_corner).MakeUnitVector();
                 geom::Vector3D treeposition =
--- a/LibCarla/source/carla/road/Map.cpp
+++ b/LibCarla/source/carla/road/Map.cpp
@@ -22,6 +22,7 @@
 
 #include <third-party/marchingcube/MeshReconstruction.h>
 
+#include <limits>
 #include <vector>
 #include <unordered_map>
 #include <stdexcept>
@@ -38,6 +39,7 @@ namespace road {
   /// We use this epsilon to shift the waypoints away from the edges of the lane
   /// sections to avoid floating point precision errors.
   static constexpr double EPSILON = 10.0 * std::numeric_limits<double>::epsilon();
+  static constexpr double TREE_PLACEMENT_EPSILON = 1.0e-4;
 
   // ===========================================================================
   // -- Static local methods ---------------------------------------------------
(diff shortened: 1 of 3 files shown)


im-hashim:
@copilot review

All four review comments have been addressed in the latest push (0e1b8dd):
- `"urban"` → `"Town"` to match existing RoadInfoSpeed descriptors
- Exact float equality replaced with double-precision squared-length epsilon guard
- `min_lane == 0` skip replaced with positive-lane fallback so one-way roads retain tree placement
- FilterRoadsByPosition fallback now picks the lane with smallest `abs(id)` instead of first in map order

im-hashim:
All Copilot review rounds complete — 24 threads resolved, latest pass generated no new comments.

**Summary of changes in the single squashed commit (`7024a5f`):**

- **Bug 1** — null dereference on roads without `maxspeed` OSM tag: null-check `RoadInfoSpeed`, fall back to `"Town"`
- **Bug 2** — trees spawning on driving surfaces: track outermost positive-ID lane as fallback, epsilon guard on degenerate widths, correct outer-corner selection for both lane sides after Unreal Y-flip
- **Bug 3** — one-way streets silently excluded from generation: `FilterRoadsByPosition` falls back to the driving lane with smallest `abs(id)` when lane ±1 is absent

Regression tests added in `LibCarla/source/test/client/test_road_regression_9565.cpp` covering all three paths. `CHANGELOG.md` updated.

Ready for final review and merge.

im-hashim:
@copilot merge conflicts resolved in 026fb725f872b228448bccb2481a443ed10acbc8. Please re-review

LuisPovedaCano (changes requested):
Thanks for the hard work on this, the fixes are solid and the test coverage is thorough!
Just one minor suggestion. In `test_road_regression_9565.cpp` at lines 205 and 240, when comparing floating-point positions it's safer to add a small epsilon tolerance in both directions, since rounding can push the result slightly above or below the analytical value.

Change it to something like:
```cpp
const float min_expected_dist = outer_road_edge + dist_from_edge;  // 9.0
  for (const auto& entry : result) {
    const float tree_y = std::abs(entry.first.location.y);
    EXPECT_GE(tree_y, min_expected_dist - 0.01f)
        << "Tree Y=" << entry.first.location.y
        << " is too close to or inside the driving surface for positive-only lanes"
        << " (|Y| must be >= " << min_expected_dist << ")";
    EXPECT_LE(tree_y, min_expected_dist + 0.01f)
        << "Tree Y=" << entry.first.location.y
        << " is too close to or inside the driving surface for positive-only lanes"
        << " (|Y| must be >= " << min_expected_dist << ")";
  }
```

The same applies to the equivalent block at line 240.
Thanks again for the effort put into this!

im-hashim:
@LuisPovedaCano 
Thank you very much for your feedback and the suggestion. The test assertions are now updated.

EXPECT_NEAR with ±0.01f epsilon tolerance in both directions.

LibCarla/source/test/client/test_road_regression_9565.cpp
```diff
@@ -0,0 +1,246 @@
...
+          </lane>
+        </left>
+        <center><lane id="0" type="none" level="false"/></center>
+        <right>
+          <lane id="-1" type="driving" level="false">
+            <width sOffset="0" a="3.5" b="0" c="0" d="0"/>
+          </lane>
+        </right>
+      </laneSection>
+    </lanes>";
```
LuisPovedaCano: Copilot is wrong about this.
This should end with `</lanes>)";`

There more places with this error in the same file

LibCarla/source/carla/road/Map.cpp
```diff
@@ -1280,27 +1282,69 @@ namespace road {
...
-                  geom::Vector3D treeposition = edges.first - director.MakeUnitVector() * distancefromdrivinglineborder;
-                  geom::Transform lanetransform = lane->ComputeTransform(s_current);
-                  geom::Transform treeTransform(treeposition, lanetransform.rotation);
-                  const carla::road::element::RoadInfoSpeed* roadinfo = lane->GetInfo<carla::road::element::RoadInfoSpeed>(s_current);
-                  transforms.push_back(std::make_pair(treeTransform,roadinfo->GetType()));
+                geom::Vector3D director = edges.second - edges.first;
+                // Use double precision for the length check to avoid false
+                // negatives from float cancellation on near-equal corners.
+                const double director_squared_length =
+                    static_cast<double>(director.x) * director.x +
```
LuisPovedaCano: This code is correct due to arithmetic promotion rules, but it looks accidental. Cast both operands explicitly or use `static_cast<double>(director.x) * static_cast<double>(director.x)`

LibCarla/source/carla/road/Map.cpp
```diff
@@ -1280,27 +1282,69 @@ namespace road {
...
+                // trees on or very close to the road surface.
+                if (director_squared_length <= (TREE_PLACEMENT_EPSILON * TREE_PLACEMENT_EPSILON)) {
+                  s_current += distancebetweentrees;
+                  continue;
+                }
+                // GetCornerPositions returns the lane corners in (t_offset + width, t_offset - width) order.
+                // The true outer edge therefore depends on the lane side: for positive lane IDs it is the second corner,
+                // and for negative lane IDs it is the first. Offset farther outward so trees are always placed away from the driving surface.
+                const bool is_positive_lane = (lane->GetId() > 0);
+                const geom::Vector3D first_corner = edges.first;
```
LuisPovedaCano: `first_corner` and `second_corner` are two local copies that are used only once each in the ternary expressions below them. These can eliminated.

LuisPovedaCano (changes requested):
I've able to test it and it works.
Please fix the changes and I will merge the PR

LibCarla/source/test/client/test_road_regression_9565.cpp
im-hashim: done.

LibCarla/source/carla/road/Map.cpp
im-hashim: done.

LibCarla/source/carla/road/Map.cpp
im-hashim: done. used edges.first and second directly.

LuisPovedaCano (approved):
Thanks for you contribution, merging. 👍

PR #9478, opened 2025-12-10
Support for Ubuntu 24.04

berndgassmann:
#### Description

Fixes #  

#### Where has this been tested?

  * **Platform(s):** ... 22.04  | 24.04
  * **Python version(s):** ... 3.10 | 3.12
  * **Unreal Engine version(s):** ... 4.26
  * ** ROS2 version(s):** humble | kilted

#### Possible Drawbacks

Need to wait until PR #9451 is merged since it's built upon that one.
Then will be able to rebase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9478)

PR #9689, opened 2026-04-23
Ubuntu 24 support for the UE4 build pipeline

LuisPovedaCano:
Description:

  ## Summary

  Ubuntu 24 (glibc 2.38+) breaks the UE4/clang-10 toolchain in several ways.
  This PR introduces a detection-based compatibility shim (`Ubuntu24Compat.sh`)
  that is a no-op on Ubuntu 20 and 22, and applies fixes only where needed.

  ### Problems fixed

  **`__isoc23_*` undefined symbols at link time**
  Server-side deps (xerces, proj, osm2odr) compiled against system glibc 2.38+
  headers reference C23 aliases (`__isoc23_sscanf`, `__isoc23_strtol`, …) that
  don't exist in UE4's centos7 sysroot. Fixed by adding `--sysroot` to
  `CMAKE_CXX_FLAGS` for all server builds, with `-idirafter /usr/include` as a
  fallback for headers absent from the centos7 SDK (e.g. `zlib.h`).

  **Bundled `ld` cannot read `.relr.dyn` ELF sections (glibc ≥ 2.36)**
  UE4's 2019 bundled linker crashes on objects from modern distros. Fixed by
  runtime-detecting the failure and wrapping `CC`/`CXX` to inject `-fuse-ld=lld`.

  **Boost b2 bootstrap fails with `--with-toolset=clang`**
  `bootstrap.sh` invokes the linker directly, bypassing the `CC`/`CXX` wrapper.
  Fixed by using `--with-toolset=gcc` on Ubuntu 24+; Boost libraries are still
  built with clang via b2 `toolset=`.

  **PEP 668 blocks system-wide `pip install` (Ubuntu 24+)**
  The wheel is now built and placed in `dist/` without being installed, with clear
  log messages showing how to install it manually (pip, conda, pyenv, uv, …).

  **CMake ≥ 4.0 rejects `cmake_minimum_required < 3.5`**
  Fixed by wrapping `cmake` to inject `-DCMAKE_POLICY_VERSION_MINIMUM=3.5`.

  ### Other changes

  - `BuildCarlaUE4.sh`: `USE_SIMREADY` defaults to `false`
  - `.gitignore`: added `.zed`, `compile_commands.json`, `Unreal/CarlaUE4/.ignore`

  ### Credits

  Approach and reference implementation by @berndgassmann in #9478.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9689)


Util/BuildTools/Ubuntu24Compat.sh
Blyron, line 1: What is this script for? is necesary?

LuisPovedaCano: yes, this script only executes if the Ubuntu version is 24 or newer, it sets the linker, because the bundled linker cannot read `.relr.dyn` sections present in `glibc ≥ 2.36` and also adds the compatibility mode for CMake

Util/BuildTools/BuildCarlaUE4.sh
```diff
@@ -14,7 +14,7 @@ BUILD_CARLAUE4=false
 LAUNCH_UE4_EDITOR=false
 USE_CARSIM=false
 USE_CHRONO=false
-USE_SIMREADY=true
+USE_SIMREADY=false
```
Blyron: we want this to be keep as true

Blyron (changes requested):

Util/BuildTools/BuildCarlaUE4.sh
LuisPovedaCano: reverted

Blyron (approved):

Issue #9706, opened 2026-05-04
Ubuntu 26.04 support tracking for ue5-dev


hdh7485:
## Is your feature request related to a problem?

The `ue5-dev` branch currently documents Ubuntu 22.04/24.04 and Windows 11 as supported environments for the Unreal Engine 5.5 version of CARLA. Ubuntu 26.04 is now available, and contributors/users on newer host systems may want to build and run CARLA UE5 without downgrading or relying only on a container workaround.

I would like to track what is needed for Ubuntu 26.04 compatibility and contribute fixes where possible, while preserving existing Ubuntu 22.04/24.04 behavior.

## Related work

This is related to the recent Ubuntu 24.04 support work:

- #9597
- #9596
- #9601
- #9617

The 24.04 work is a useful reference because 26.04 may expose similar or newer issues around glibc, compiler/toolchain versions, Python packaging behavior, apt package names, NVIDIA/Vulkan dependencies, and CI/container images.

## Proposed scope

Initial goal: identify and document the compatibility delta for Ubuntu 26.04 on `ue5-dev`.

Potential tasks:

- [ ] Test `CarlaSetup.sh --interactive` on Ubuntu 26.04
- [ ] Record toolchain versions: Python, GCC/Clang, CMake, glibc, NVIDIA driver, Vulkan
- [ ] Identify apt package changes compared with 22.04/24.04
- [ ] Check whether the existing 24.04 glibc compatibility shim is sufficient on 26.04
- [ ] Check Python/pip behavior, including PEP 668 and Python version compatibility
- [ ] Verify CMake configure and `cmake --build Build`
- [ ] Verify `cmake --build Build --target carla-python-api-install`
- [ ] Verify editor launch or package build if native build succeeds
- [ ] Add documentation notes for 26.04 if full support is not yet ready
- [ ] Consider CI/devcontainer follow-up if maintainers want 26.04 covered there

## Expected outcome

A small, incremental path toward Ubuntu 26.04 compatibility for CARLA UE5. This does not need to imply full official support immediately; even documenting current blockers and known workarounds would help contributors using newer Ubuntu hosts.

## Notes

I can provide build logs from Ubuntu 26.04 and work on focused follow-up PRs once the first failure points are identified.

hdh7485:
I started native Ubuntu 26.04 validation on the current `ue5-dev` head.

## Environment

- OS: Ubuntu 26.04 LTS (`resolute`)
[...]

PR #9709, opened 2026-05-05
Support pygame source builds on Ubuntu 26.04

hdh7485:
## Summary

This keeps the Ubuntu 26.04 setup path working when `requirements.txt` installs `pygame` under Python 3.14.

Ubuntu 26.04 defaults to Python 3.14, but `pygame==2.6.1` does not currently publish `cp314` wheels on PyPI. As a result, pip falls back to building pygame from source. That source build needs SDL2 and freetype development headers, which were not installed by `InstallPrerequisites.sh`.

This PR also skips pip self-upgrade on Ubuntu 26.04. The distro `python3-pip` package is Debian-managed and fails when pip tries to uninstall it during `pip install --upgrade pip`; the packaged pip is sufficient to build/install the current requirements.

Related to #9706.

## Changes

- Add `libsdl2-dev` and `libfreetype-dev` to Ubuntu prerequisites.
- Skip `python -m pip install --upgrade pip` on Ubuntu >= 26.04.
- Keep the existing `--break-system-packages` behavior for Ubuntu >= 24.04.

## Verification

Ran the prerequisite script in a fresh Ubuntu 26.04 container:

```bash
docker run --rm -v "$PWD:/carla" -w /carla ubuntu:26.04 bash -lc '
  set -euo pipefail
  export DEBIAN_FRONTEND=noninteractive
  apt-get update
  apt-get install -y sudo ca-certificates
  bash -x Util/SetupUtils/InstallPrerequisites.sh --python-path=python3
'
```

Key output:

```text
Skipping pip self-upgrade on Ubuntu 26.04.
Created wheel for pygame: filename=pygame-2.6.1-cp314-cp314-linux_x86_64.whl
Successfully built numpy pygame
Successfully installed ... pygame-2.6.1 ...
DOCKER_PREREQ_EXIT_STATUS=0
```

Also ran:

```bash
bash -n Util/SetupUtils/InstallPrerequisites.sh
```

## Not tested

- Full native `CarlaSetup.sh --interactive` after sudo authentication.
- CARLA content clone.
- Unreal Engine clone/build.
- CARLA CMake configure/build.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9709)

### DIFF
--- a/Util/SetupUtils/InstallPrerequisites.sh
+++ b/Util/SetupUtils/InstallPrerequisites.sh
@@ -66,7 +66,9 @@ sudo apt-get -y install \
     libxkbcommon-dev \
     libgbm-dev \
     libpango1.0-dev \
-    libasound2-dev
+    libasound2-dev \
+    libsdl2-dev \
+    libfreetype-dev
 
 if [ "$python_path" == "python3" ]; then
     sudo apt-get -y install \
@@ -84,7 +86,14 @@ PIP_EXTRA_ARGS=""
 if dpkg --compare-versions "$UBUNTU_VERSION_ID" ge "24.04"; then
     PIP_EXTRA_ARGS="--break-system-packages"
 fi
-$python_path -m pip install --upgrade pip $PIP_EXTRA_ARGS
+if dpkg --compare-versions "$UBUNTU_VERSION_ID" ge "26.04"; then
+    # Ubuntu 26.04's python3-pip package is Debian-managed and cannot be
+    # uninstalled by pip during a self-upgrade. Keep the distro pip and only
+    # install the project requirements.
+    echo "Skipping pip self-upgrade on Ubuntu $UBUNTU_VERSION_ID."
+else
+    $python_path -m pip install --upgrade pip $PIP_EXTRA_ARGS
+fi
 $python_path -m pip install -r requirements.txt $PIP_EXTRA_ARGS
 
 # -- INSTALL CMAKE --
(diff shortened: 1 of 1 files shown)


Util/SetupUtils/InstallPrerequisites.sh
```diff
@@ -84,7 +86,14 @@ PIP_EXTRA_ARGS=""
 if dpkg --compare-versions "$UBUNTU_VERSION_ID" ge "24.04"; then
     PIP_EXTRA_ARGS="--break-system-packages"
 fi
-$python_path -m pip install --upgrade pip $PIP_EXTRA_ARGS
+if dpkg --compare-versions "$UBUNTU_VERSION_ID" ge "26.04"; then
```
LuisPovedaCano: The condition uses `ge "26.04"`, so the pip self-upgrade skip will also apply to Ubuntu 28.04, 30.04, and all future releases. If a future Ubuntu ships a pip that can self-upgrade safely, this will silently skip it anyway.
Consider scoping it to the Ubuntu 26 version for now:
`if dpkg --compare-versions "$UBUNTU_VERSION_ID" eq "26.04"; then`
If Ubuntu 28.04 ships with the same Debian-managed pip constraint, the condition can be changed back to `ge "26.04"` at that point.

LuisPovedaCano (changes requested):

Util/SetupUtils/InstallPrerequisites.sh
hdh7485: Fixed in 53172db by scoping the pip self-upgrade skip to Ubuntu 26.04 only:

`dpkg --compare-versions "$UBUNTU_VERSION_ID" eq "26.04"`

I also checked the condition locally: 24.04 -> no skip, 26.04 -> skip, 28.04 -> no skip, and `bash -n Util/SetupUtils/InstallPrerequisites.sh` passes.

LuisPovedaCano (approved):

Issue #9733, opened 2026-05-19
Black screen in pygame window when running the automatic_control.py


peizhang2:
**Setup**
Describe the setup you are using to run CARLA along with its version:
  * CARLA version: latest code in branch ue5-dev
  * Platform: Ubuntu 24.04
  * Hardware: AMD Ryzen 7 9700X / 32GB RAM / 1TB SSD
  * GPU: NVIDIA GeForce RTX 5070 Ti 16GB VRAM
  * GPU Drivers: 595.58.03

**Describe the bug**
When I run the automatic_control.py
```
cd CarlaUE5/Build/Package/Carla-0.10.0-Linux-Shipping/PythonAPI/examples
python3 automatic_control.py
```
The pygame window shows black screen, and almost everything are rendered black in the pygame window, except the traffic lights (red or green). I can only see the traffic lights.
I also started a `manual_control.py` window. The `manual_control.py` works fine.
See:
<img width="4869" height="1427" alt="Image" src="https://github.com/user-attachments/assets/c15b79bc-d7c2-4951-b677-a2fa0481e112" />
There are 3 red dots in the `automatic_control.py` window, and they are traffic lights.

**Steps to reproduce**
1. Pull the last code . The `git log` shows:
```
commit e0531f34d2b5d96eef910bfaac25d8aa0bb9d2ac (HEAD -> ue5-dev, origin/ue5-dev, origin/HEAD)
Author: Jesus Armando Anaya <armando@robomous.ai>
Date:   Mon May 18 10:35:17 2026 -0700
```
2. Build the Carla package.
```
./CarlaSetup.sh
cmake --build Build --target package
```
3. Enter the package directory and run Carla.
```
cd /home/peizhang/github.com/CarlaUE5/Build/Package/Carla-0.10.0-Linux-Shipping/Linux/
./CarlaUnreal.sh
```
4. Open a new terminal.
And run the `automatic_control.py`.
```
cd /home/peizhang/github.com/CarlaUE5/Build/Package/Carla-0.10.0-Linux-Shipping/PythonAPI/examples/
python3 automatic_control.py
```

**Expected behavior**
The `automatic_control.py` should render the 3D graphics same as the `manual_control.py`.

**Logs**
The log file below is from:
```
/home/peizhang/github.com/CarlaUE5/Unreal/CarlaUnreal/Saved/Logs/CarlaUnreal.log
```
This file is generated when I run the:
```
cmake --build Build --target launch-only
```

[...]

PR #9736, opened 2026-05-20
fix(plugin): default camera post-process profile per active map

JArmandoAnaya:
#### Description

The RGB camera variation defaults `post_process_profile` to the literal `default`, and the server falls back to that same string when the client does not set the attribute. The matching JSON lives in carla-content as `Carla/Config/PostProcess/Default.json`, which Linux's case-sensitive filesystem does not match against `default`. The load fails silently, the `SceneCapture` keeps the ctor's flags-only zero overrides set by `SetCameraDefaultOverrides`, and Town10HD_Opt renders nearly black except for emissive surfaces (the red and green traffic-light LEDs the user reported on the issue).

Even with the case matched, `Default.json` is calibrated for a much brighter scene than the shipped maps (`AEM_Manual` with ISO 100 and shutter 1/320), so blindly applying it leaves Town10HD_Opt 15 stops underexposed. `manual_control.py` already works around this client side via `get_post_process_profile(map_name)`, but `automatic_control.py` and every third-party script that does not know about the profile system hit the broken default path.

When the client passes no profile (or the legacy lowercase `default` sentinel), the server now picks the JSON named after the active map and falls back to `Default.json` only when no per-map profile is shipped.

Issue: #9733

#### Changes

- `ActorBlueprintFunctionLibrary.cpp`: in the camera setup path under `SetCamera`, treat an empty or `default` (any case) profile as "no preference"; resolve it to `<active map name>.json` if that file exists under `Carla/Config/PostProcess/`, else `Default.json`. Map name is read via `World->GetMapName()` with the standard PIE-prefix strip used elsewhere in the plugin (e.g. `VehicleActorFactory.cpp`).
- `ActorBlueprintFunctionLibrary.cpp`: bump the variation's `RecommendedValues` advertised to clients from `default` to `Default` so the API hint matches the actual filename on case-sensitive filesystems.

  * **Platform(s):** Linux Ubuntu 24.04
  * **Python version(s):** 3.12
  * **Unreal Engine version(s):** 5.5

Built locally (not in the Docker dev container) against the host UE 5.5 and ran the packaged build on `Town10HD_Opt`. Before the patch, stock `automatic_control.py` reproduces the issue: pygame view renders black with only the traffic-light LEDs visible. After the patch, the same unmodified script renders the scene correctly. `manual_control.py`, which already passes an explicit profile, is unaffected.

#### Possible Drawbacks

- Clients that intentionally pass the literal string `default` and rely on the legacy `Default.json` profile being applied regardless of the active map will now get the per-map profile instead. Pass any other explicit profile name to keep the old behavior (e.g. `Default` with a capital `D`); the substring match is case-insensitive only for the legacy `default` sentinel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9736)

### DIFF
--- a/Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
+++ b/Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
@@ -1424,8 +1424,12 @@ void UActorBlueprintFunctionLibrary::SetCamera(
         MapName = World->GetMapName();
         MapName.RemoveFromStart(World->StreamingLevelsPrefix);
       }
-      const FString MapJsonPath =
-          FPaths::ProjectContentDir() / TEXT("Carla/Config/PostProcess/") + MapName + TEXT(".json");
+      else
+      {
+        UE_LOG(LogCarla, Warning,
+            TEXT("SetCamera: camera has no UWorld; falling back to Default post-process profile."));
+      }
+      const FString MapJsonPath = UPostProcessJsonUtils::GetPostProcessConfigPath(MapName);
       PostProcessProfileName = FPaths::FileExists(MapJsonPath) ? MapName : TEXT("Default");
     }
 
--- a/Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
+++ b/Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
@@ -411,7 +411,7 @@ void UActorBlueprintFunctionLibrary::MakeCameraDefinition(
     FActorVariation post_process_profile;
     post_process_profile.Id = TEXT("post_process_profile");
     post_process_profile.Type = EActorAttributeType::String;
-    post_process_profile.RecommendedValues = {TEXT("default")};
+    post_process_profile.RecommendedValues = {TEXT("Default")};
     post_process_profile.bRestrictToRecommended = false;
 
     Definition.Variations.Append({PostProccess, post_process_profile});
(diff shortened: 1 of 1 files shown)


Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
```diff
@@ -1407,11 +1407,31 @@ void UActorBlueprintFunctionLibrary::SetCamera(
...
+    if (PostProcessProfileName.IsEmpty() || PostProcessProfileName == TEXT("default"))
+    {
+      const UWorld *World = Camera->GetWorld();
+      FString MapName{};
+      if (World != nullptr)
+      {
+        MapName = World->GetMapName();
+        MapName.RemoveFromStart(World->StreamingLevelsPrefix);
+      }
+      const FString MapJsonPath =
```
LuisPovedaCano: There is a static method for retrieving this path.
Should be `const FString MapJsonPath = UPostProcessJsonUtils::GetPostProcessConfigPath(MapName);`

Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
```diff
@@ -1407,11 +1407,31 @@ void UActorBlueprintFunctionLibrary::SetCamera(
...
+        TEXT(""));
+
+    // Empty or the legacy lowercase "default" sentinel means "no preference":
+    // load the profile named after the active map. Case-sensitive so an
+    // explicit "Default" still force-loads Default.json.
+    if (PostProcessProfileName.IsEmpty() || PostProcessProfileName == TEXT("default"))
+    {
+      const UWorld *World = Camera->GetWorld();
+      FString MapName{};
+      if (World != nullptr)
```
LuisPovedaCano, line 1422: Add a log when `World == nullptr` to avoid silently falls through to "Default".

LuisPovedaCano (changes requested):

Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
JArmandoAnaya: Change done.

Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Actor/ActorBlueprintFunctionLibrary.cpp
JArmandoAnaya: Change done.

peizhang2:
I verified this PR in https://github.com/carla-simulator/carla/issues/9733 .

LuisPovedaCano (approved):

PR #9756, opened 2026-05-26
feat(ros2): [6/8] add ROS2TopicVisibility default-startup flag

JArmandoAnaya:
## Description

> Blocked on PR https://github.com/carla-simulator/carla/pull/9751

Adds a `ROS2TopicVisibility` startup setting that controls whether sensor topics are visible to ROS 2
from the moment their stream is created. When `true` (the default), every topic offered by the
implementation publishes from startup without an explicit `enable_for_ros` call; when `false`, only the
sensors explicitly enabled through `Sensor.enable_for_ros()` are visible (useful for disabling interfaces,
for example on the leaderboard).

On ue5-dev a sensor only produces and publishes data while `AreClientsListening()` is true, which holds
when a TCP client is subscribed, the stream is force-active, or the stream is marked enabled-for-ros.
That ROS flag defaults to off per stream and today flips on only through the per-sensor
`enable_sensor_for_ros` RPC, so a freshly spawned sensor stays silent on ROS 2 until a client enables it.
This setting lets the server flip the default for every newly created stream at startup.

Lands the topic-visibility-default slice of upstream `3b938d718` (#9431), adapted to ue5-dev's existing
stream-level visibility model. The multistream and per-actor scaffolding from that commit
(`RpcServerInterface`, the `Message.h` relocation, the `Session` abstraction, the Dispatcher shared
pointer migration, per-actor visibility) is **not included**: it is groundwork for the out-of-scope DDS
middleware series and the ROS-2-as-RPC-client work, and has no consumer on ue5-dev.

Closes: #9627

## Changes

- `Settings/CarlaSettings.h`: new `config` UPROPERTY `ROS2TopicVisibility` (default `true`), bound to
  `DefaultGame.ini` like the existing `ROS2` flag.
- `streaming/detail/Dispatcher.{h,cpp}`: a `_topic_visibility_default_enabled` flag with a
  `SetROS2TopicVisibilityDefaultEnabled(bool)` setter; when set, every stream created through
  `MakeStream()` and through the `GetToken()` create-on-miss path starts enabled for ROS 2.
- `streaming/Server.h` and `streaming/low_level/Server.h`: thin pass-throughs for the new setter. The
  existing stream-id based `EnableForROS` / `DisableForROS` / `IsEnabledForROS` surface is unchanged.
- `Game/CarlaEngine.cpp`: inside the existing `if (Settings.ROS2)` block of `NotifyInitGame`, apply the
  configured default to the streaming server before any sensor stream is created. Gating on
[...]

PR #9748, opened 2026-05-22
feat(ros2): [4/8] unify point-cloud + scalar publishers, retire legacy listeners

JArmandoAnaya:
## Description

> Blocked on PR https://github.com/carla-simulator/carla/pull/9746

Lands the remainder slice of upstream ROS2 Refactor `cd3d640f3` (#9097) on `ue5-dev`. PR-3 unified the camera publishers; this PR finishes the refactor on the point-cloud + scalar side. A new `CarlaPointCloudPublisher` base over PR-2's `BasePublisher` template collapses lidar, semantic-lidar and radar into ~25-LOC subclasses; the scalar publishers (clock, gnss, imu, transform, collision) migrate from the legacy `CarlaPublisher` base to `BasePublisher` + `PublisherImpl<Traits>` directly; DVS becomes a composite that owns an image-side `CarlaCameraPublisher` subclass (3-channel `bgr8`) and a point-cloud-side `CarlaPointCloudPublisher` subclass.

PR-1's IMU compass-yaw fix (`ImuMath.h`) and PR-1's transform degree-to-radian fix carry forward unchanged. The transform math moves into a new `TransformQuaternion.h` header-only seam so a GTest regression-pins the conversion without linking FastDDS.

`ROS2.{h,cpp}` retires the legacy `AddActorRosName` / `RemoveActorRosName` / `UpdateActorRosName` / `GetActorRosName` / `GetActorParentRosName` / `AddActorCallback` / `RemoveActorCallback` surface in favour of PR-2's `RegisterSensor` / `UnregisterSensor` / `RegisterVehicle` / `UnregisterVehicle`. `_actor_ros_name` becomes an `ActorRegistration` map keyed by `void*` that carries `ros_name + frame_id + publish_tf`. A new `GetOrCreateTransformPublisher` returns a per-sensor TF publisher lazy-constructed on first publish, gated by the actor's `publish_tf` flag. Seven legacy files are deleted (`CarlaPublisher.h`, `CarlaMapSensorPublisher.{h,cpp}`, `CarlaSpeedometerSensor.{h,cpp}`, `listeners/CarlaListener.{h,cpp}`).

Closes: #9627

## Changes

[...]

PR #9743, opened 2026-05-22
fix(ros2): [1/8] correct IMU compass yaw and reorder actor disposal

JArmandoAnaya:
## Description

> It is necessary to merge the open PR https://github.com/carla-simulator/carla/pull/9736 before this one to fix the camera issue present in the ROS2 camera topic.

`CarlaIMUPublisher::SetData` shipped `(float(M_PI_2) / 2.0f) - compass` for the yaw used to build the orientation quaternion. That evaluates to `pi/4 - compass`, off by 45 degrees from the intended REP-103 conversion `pi/2 - compass`. Every ROS 2 IMU consumer on `ue5-dev` has been receiving an orientation rotated by 45 degrees. The math is extracted into a small `ImuMath.h` helper so it can be unit-tested without a live FastDDS publisher.

`UActorDispatcher::OnActorDestroyed` now runs the ROS 2 `RemoveActorRosName` callback before `Registry.Deregister`. Today on `ue5-dev` this is observably a no-op (the legacy API only consumes the raw `AActor*`), but it prepares cleanly for the upcoming `UnregisterSensor` / `UnregisterVehicle` work, which reads `FCarlaActor::GetActorInfo()->Description` and therefore needs the actor still alive at unregister time. Adapted from `ue4-dev` `201d375d3`.

Closes: #9627

## Changes

- New `LibCarla/source/carla/ros2/publishers/ImuMath.h` with `OrientationFromCompass(float)` that returns the yaw-only quaternion as `std::array<float, 4>` (`w, x, y, z`).
- `CarlaIMUPublisher::SetData` calls the helper; the inline Euler-to-quaternion block is removed.
- `UActorDispatcher::OnActorDestroyed` hoists the `WITH_ROS2` block ahead of `Registry.Deregister(...)` and collapses the dead nested `if` into a single predicate.
- New GTest `LibCarla/source/test/common/test_imu_compass.cpp` covering North, East, South, West, plus a regression guard against the previous `pi/4 - compass` spelling.

## Where has this been tested?

- Platform: Ubuntu 24.04.
- Python: 3.10.
- UE: 5.5.

`libcarla_test_server` (73/73) and `libcarla_test_client` (87/87) pass with the new `ImuMath.*` cases green in both suites; `cmake --build Build --target package` completes for the Release build.

## Possible Drawbacks

[...]

Issue #9627, opened 2026-04-02
Port ue4-dev ROS2 enhancements to ue5-dev (Category 2: ROS2)


youtalk:
**Is your feature request related to a problem? Please describe.**

The `ue4-dev` branch contains significant ROS2 architectural improvements and new functionality that have not been ported to `ue5-dev`. This includes a major ROS2 refactor (BasePublisher/PublisherImpl template pattern), new geometry data types, multistream preparation, V2X sensor support, and fine-grained server synchronization. Part of the broader effort discussed in https://github.com/carla-simulator/carla/discussions/9581.

**Describe the solution you'd like**

Port the following ROS2 enhancements as new implementations targeting `ue5-dev`:

- [ ] **PR #0e** feat(ros2): transformation fix and refactor — transform calculation fix (`a46b2690b`), ROS2 architecture refactor with BasePublisher/PublisherImpl pattern (`cd3d640f3`), multi-GPU actor disposal fix (`201d375d3`), dependency compilation fix (`a6f6ed03d`)
- [ ] **PR #0f** feat(ros2): geometry data types and multistream support — new geometry types: AngularVelocity, Acceleration, Quaternion, Velocity, RightHandedVector3D, pitch/roll rotation fix (`590d54f02`), multistream preparation with RpcServerInterface and dynamic ROS visibility (`3b938d718`)
- [ ] **PR #0g** feat(ros2): V2X sensor and serializer — V2X sensor channel_id, binary data, multi-message, V2I support (`8f33324d9`), serializer/RPC extensions with SensorHeader relative transform, ArrayConst, control timestamps (`1146277a7`). Note: V2X base implementation (`c7cfdad9a`) also needs UE5 porting.
- [ ] **PR #0h** feat(ros2): fine-grained server synchronization — per-client independent synchronization with time window constraints (`aa9c92b21`)

**Describe alternatives you've considered**

Cherry-picking from ue4-dev was considered but rejected due to UE4→UE5 architectural differences.

**Additional context**

This is **Category 2** of the ue4-dev→ue5-dev porting effort. Dependency chain: PR #0e (refactor) → #0f (geometry+multistream) → #0g (V2X+serializer) / #0h (sync). The ROS2 refactor in `cd3d640f3` (95 files) is the foundation — all subsequent PRs build on the refactored architecture.

youtalk:
Closed because the refactor porting PR cannot be merged. https://github.com/carla-simulator/carla/pull/9638

[...]

PR #9757, opened 2026-05-27
feat(sensor): [7/8] add V2X sensor family (CAM service, path-loss, CustomV2X, V2I)

JArmandoAnaya:
## Description

Ports the V2X sensor family to ue5-dev: the CAM sensor (`sensor.other.v2x`) with ETSI CAM generation, the custom-payload sensor (`sensor.other.v2x_custom`) carrying a fixed binary blob, the `PathLossModel` propagation engine, and owner-less V2I (infrastructure) sensors.

V2X is a client-streamed sensor: data reaches the client through the `listen()` callback like any other sensor, not through a DDS topic. No ROS 2 publisher is added here, matching upstream. The "ROS 2 support" framing is inherited from the upstream commit title (#9435); the LibITS message model and serializers landed here are the foundation a future V2X ROS 2 publisher would consume.

Part of the ue4-dev to ue5-dev ROS 2 port (#9627). Adapted from ue4-dev `c7cfdad9a` and `8f33324d9`.

## Changes

- New LibCarla message model: `LibITS.h` (ETSI ITS types), `V2XData.h` / `V2XEvent.h` containers, `V2XSerializer`, and the `CustomV2XBytes` RPC type (100-byte blob).
- New plugin sensors `AV2XSensor` / `ACustomV2XSensor`, with `CaService` (CAM generation) and `PathLossModel` (FSPL / two-ray / ETSI fading), owned via `std::unique_ptr`; static actor maps cleared in `EndPlay` to avoid dangling `AActor*` keys.
- Registers both sensors in `SensorRegistry` (appended) and adds the client `send()` path: `ServerSideSensor` to `Simulator` to `Client` to server `BIND_SYNC(send)`.
- Actor-definition builders in `ActorBlueprintFunctionLibrary`, Python bindings (`V2XData.cpp`), and reference docs.
- Tests: GTest serializer and blob round-trip; smoke test for CAM, custom send/receive, and V2I.

## Series

| PR | Theme | Status |
|----|-------|--------|
| 1 | IMU compass yaw + actor disposal order | #9743 |
| 2 | Template infra + Ackermann subscriber | #9745 |
| 3 | Camera publisher unification | #9746 |
| 4 | Point-cloud + scalar publishers | #9748 |
| 5 | geom types + pitch/roll fix | #9751 |
| 6 | ROS2TopicVisibility default flag | #9756 |
| 7 | V2X sensor family (this PR) | here |
| 8 | WalkerManager null-guard + series CHANGELOG | #9758 |

## Where has this been tested?

- Platform/OS: Ubuntu 24.04, local cmake Development package build, 0 errors and no warnings on touched files.
- Python version: 3.10. `smoke.test_v2x` 3/3 pass; V2X serializer GTests pass on both `libcarla_test_client` and `libcarla_test_server`.
- Unreal Engine version: 5.5. Server stayed alive through the smoke run (no crash).

## Possible Drawbacks

- The CAM message is serialized as a raw POD blob over `long` fields, so a cross-platform client/server pair (Linux 8-byte vs Windows 4-byte `long`) would misparse CAM. Same-platform is unaffected. Inherited from upstream; a future ROS 2 publisher would serialize field-by-field and sidestep it.
- `PathLossModel` frequency parameters are static, so all V2X sensors share the last-set frequency. Inherited from upstream.
- CHANGELOG entry is deferred to the final PR of the series.

## Related

#9627, #9435, #9743, #9745, #9746, #9748, #9751, #9756

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9757)


youtalk (commented):
Reviewed locally on a fresh worktree (LibCarla built with `ENABLE_ROS2=ON`; `CustomV2X*` GTests and the full suite pass; `carla-unreal-editor` compiles+links clean). The `CustomV2XBytes` MsgPack round-trip is sound and bounds-checked on both `set_bytes` / `set_string`. A few of the items below are inherited from the upstream V2X transliteration, but since they ship here I'm flagging them anyway.

### Should fix before merge

**1. `mGenCamLowDynamicsLimit` is never initialized.** Declared at `CaService.h:58` and read at `CaService.cpp:181` (`if (++mGenCamLowDynamicsCounter >= mGenCamLowDynamicsLimit)`) — `mGenCamLowDynamicsCounter` is initialized (line 175) but the limit is not, so this is an uninitialized read (UB); the CAM low-dynamics fallback rate is nondeterministic per process run. Upstream CARLA initializes it to `3`. Fix: `unsigned int mGenCamLowDynamicsLimit{3};`.

**2. `PrevLocation` / `PrevDeltaTime` are never seeded.** Declared at `CaService.h:93,96`, used at `CaService.cpp:610-625` (`ComputeAccelerometer`) with no first-call guard. UE's `FVector` default ctor leaves components uninitialized, so the first acceleration computation reads garbage; and if `PrevDeltaTime` (H2) happens to be 0, the `Y1/(H1*H2)` and `Y2/(H2*(H1+H2))` terms divide by zero → inf/NaN propagated into the CAM longitudinal/lateral/vertical acceleration fields. Fix: in the vehicle-init branch of `SetActor`, seed `PrevLocation[0] = PrevLocation[1] = CurrentLocation` and `PrevDeltaTime` to a small positive value (mirroring `AInertialMeasurementUnit`).

### Please confirm

**3. `PathLossModel` `Frequency_GHz` / `Frequency` / `lambda` are `static`** (`PathLossModel.h:86-88`, mutated in `SetParams`). Two V2X / CustomV2X sensors configured with different `frequency_ghz` clobber each other — the last `SetParams` wins globally and corrupts the FSPL/two-ray/knife-edge math for all sensors (and `CalculateFSPL_d0()` reads the shared static `Frequency` while `m_fspl_d0` is per-instance). It's also a latent data race if sensor setup ever overlaps a tick. Recommend making frequency/lambda non-static per-instance fields.

**4. `log10(Distance3d)` has no zero-guard.** In `ComputeLoss` (NLOSv branch) and `CalculatePathLoss_WINNER` / `CalculateTwoRayPathLoss` (`PathLossModel.cpp:330-335,384,403`), two actors sharing a location give `Distance3d == 0` (which passes the `< filter_distance` filter), so `20·log10(Distance3d)` / `log10(Distance3d/ref)` yield `-inf`/NaN that then flow into the `ReceivedPower` comparison. Clamp `Distance3d` to a small epsilon (e.g. `max(Distance3d, reference_distance_fspl)`) before any `log10`.

**5. CAM serialization is a raw byte copy of structs that contain a raw pointer.** `PathPoint_t::pathDeltaTime` is a `PathDeltaTime_t*`, and `V2XSerializer.h` (`output.copy_from(data.MessageList)`) memcpys the whole `std::vector<CAMData>` — including that pointer's server-side value — onto the wire, which the client reinterprets back via `Array<CAMData>`. It's safe today only because `NumberOfPathPoint` is always 0 (so the array is never walked) and client/server share arch/endianness. Any code that populates path history, or any cross-arch client, would dereference a wild pointer. Recommend dropping the raw pointer from the wire type (store the value inline / an availability flag) or defining an explicit field-by-field serializer.

### Improvements

- `CustomV2XSensor.cpp:172`: `pair.second.front().Power` without an empty-list guard — relies on the invariant that the map never holds an empty list; add `if (pair.second.empty()) continue;`.
- `CustomV2XSensor.cpp:189` / `V2XSensor.cpp:211`: `std::map::at(...)` on server-side code can throw `std::out_of_range`; prefer `find` + skip-on-miss (or wrap under `#ifndef LIBCARLA_NO_EXCEPTIONS`).
- `PythonAPI/carla/src/V2XData.cpp` `SetCustomV2XBytes`: the copy length is taken from `boost::python::len(object)` (element count) rather than `buffer.len` (bytes); for a buffer with `itemsize > 1` (e.g. a NumPy int32 array) this mis-sizes the copy. Use `buffer.len` clamped to `max_size()`.

### Nits

- `PythonAPI/carla/src/V2XData.cpp:25`: typo `"Uunknown"` in `GetStationTypeString` (the `default` case correctly returns `"Unknown"`) — user-visible string.
- `CaService.cpp:211`: comment typo `"differance"` → `"difference"`.
- The new UE-plugin V2X files (`CaService.*`, `PathLossModel.*`, `V2XSensor.*`, `CustomV2XSensor.*`, the `ActorBlueprintFunctionLibrary` additions, `LibITS.h`) use Allman braces with some trailing whitespace; a uncrustify pass per the UE coding standard would normalize them. Mostly inherited from upstream — low priority.

JArmandoAnaya:
@youtalk Thanks for all the points added to the review. I already refactored the code and added new commits with the suggested changes.

youtalk (approved):
LGTM!

JArmandoAnaya:
This PR is ready for code review now.

Aperiss (approved):

Aperiss:
Tested the changes, LGTM. When workflow passes this will be merged.

PR #9780, opened 2026-06-08
fix(windows): fix Windows build after UE 5.5.4 raised the MSVC toolset minimum

### LINKED ISSUE #9781: Windows (ue5-dev): build fails to compile.
**Setup**
  * CARLA version: 0.10.0 (`ue5-dev` branch), CARLA's Unreal Engine fork upgraded to 5.5.4
  * Platform: Windows 11 Pro (x64), Visual Studio 2022 with a current MSVC toolset (newer than 14.38, e.g. 14.4x)
  * Python version: 3.10
  * GPU: N/A (this is a build-time / compilation failure, independent of the GPU)
  * GPU Drivers: N/A

**Describe the bug**
After CARLA's Unreal Engine fork was upgraded from **5.5 to 5.5.4** (an official Epic patch), the Windows build of `ue5-dev` no longer compiles out of the box. The 5.5.4 patch raises the minimum MSVC toolset the engine builds against to **14.38.33130** (recorded in the engine's `Windows_SDK.json`). Because the required compiler version changed, the existing Windows build setup no longer matches it:

- The setup pinned an older toolset (14.36), which is now below the new engine minimum, so the build fails (e.g. the RigLogic build).
- The default "x64 Native Tools Command Prompt" instead selects the *newest* installed toolset. Building with a newer-than-target toolset surfaces stricter warnings and tighter conformance/linker behavior that break compilation:

  1. **Warnings flood promoted to fatal errors.** CARLA applies its extra compiler warnings globally as `/Wall`, emitting tens of thousands of off-by-default warnings from system and third-party headers (C4820, C4710/C4711, C4514, C5045, C4625/C4626/C5026/C5027, ...). Because the bundled GoogleTest dependency compiles with "warnings as errors" (`/W4 /WX`), these are promoted to fatal `C2220` errors and the build aborts.
  2. **Unresolved external symbol (link error).** `carla::geom::Location` is forward-declared as a `class` in the navigation agents but defined as a `struct`. MSVC name-mangles `class` and `struct` differently, so the navigation-agent demo fails to link with `LNK2019`. Linux (Itanium ABI) ignores the class/struct key, so this only breaks on MSVC.
  3. **Non-conforming type rejected by newer MSVC.** A LibCarla test uses `std::independent_bits_engine` with an `unsigned char` result type; the standard requires an unsigned type wider than `char`, and newer MSVC rejects it with `C2338`.
  4. **Latent uninitialized read.** A `TrafficLightState` is read uninitialized on the walker path (`C4701`), surfaced by the stricter toolset.

The net effect is that, after the 5.5.4 upgrade, following `Docs/build_windows_ue5.md` on a fresh `ue5-dev` checkout does not produce a working build.

**Steps to reproduce**
1. On Windows 11 with Visual Studio 2022 installed with a current MSVC toolset (newer than 14.38).
2. Clone CARLA and check out `ue5-dev` (with the UE fork on 5.5.4).
3. Follow `Docs/build_windows_ue5.md`: open the "x64 Native Tools Command Prompt for VS 2022" (which selects the newest toolset) and run the setup/build (`CarlaSetup.bat`, then the CMake/Ninja build).
4. The build fails during compilation/linking with the errors below.

**Expected behavior**
A fresh checkout of `ue5-dev` compiles and builds successfully on Windows by following the documented build steps, using the toolset the upgraded engine (5.5.4) requires, regardless of which (supported) Visual Studio 2022 edition or MSVC toolset version the contributor has installed.

**Logs**
Server/client runtime logs are not applicable (the failure happens at compile/link time). The characteristic failing messages are:

```
# (1) GoogleTest warnings-as-errors flood
cl : Command line warning D9025 : overriding '/W3' with '/Wall'
xxxxxx.h(NN): error C2220: the following warning is treated as an error
xxxxxx.h(NN): warning C4711: function '...' selected for automatic inline expansion
xxxxxx.h(NN): warning C4710: function '...' not inlined
xxxxxx.h(NN): warning C4820: 'N' bytes padding added after data member '...'

# (2) class/struct mismatch link error (navigation agents)
LNK2019: unresolved external symbol "... ComputeDistance ..." referenced in function ...
carla-agent-demo.exe : fatal error LNK1120: 1 unresolved externals

# (3) independent_bits_engine non-conforming type
error C2338: static_assert failed: 'note: char not allowed as a result_type'  (independent_bits_engine)
```

**Scripts**
N/A — this is reproduced by the standard documented build process, not a runtime script.

**Screenshots**
N/A.

**Additional context**
The trigger for all of this is the engine fork upgrade 5.5 -> 5.5.4, which bumped the required MSVC toolset to 14.38. The fix pins and activates 14.38 specifically (rather than the older pinned 14.36 or the newest default toolset), and additionally cleans up the warning configuration and repairs the source/build issues newer toolsets expose so the build is robust across toolset versions. A fix is proposed in PR #9780.

JArmandoAnaya:
#### Description

Fixes #9781.

UE 5.5.4 raised the engine's minimum MSVC toolset to **14.38.33130**, but `Util/SetupUtils/InstallPrerequisites.bat` still pinned 14.36, so a fresh Windows setup could not build. This PR makes a clean Windows build work end to end:

- **Pin MSVC 14.38 in `InstallPrerequisites.bat`** (one line). UnrealBuildTool selects the engine-preferred toolset automatically once it is installed.
- **Fix three pre-existing MSVC compile/link breaks in LibCarla** (identical with 14.38 and 14.44, so independent of the toolset version):
  - Use `/W4` instead of `/Wall` on MSVC — `/Wall` floods system-header warnings that googletest's `/WX` turns into fatal errors (C2220). GCC/Clang behavior unchanged.
  - `LibCarla/source/test/Buffer.cpp`: `std::independent_bits_engine<unsigned char>` is forbidden by the standard and rejected by the MSVC STL (C2338).
  - `LibCarla/source/carla/agents/navigation/Misc.h`: `class`/`struct` forward-declaration mismatch for `geom::Location` breaks MSVC linking (LNK2019).
- **Fix packaged builds freezing at startup**: `g.TimeoutForBlockOnRenderFence` (set in `DefaultEngine.ini` by #9744) is in milliseconds, so the value `30` armed a 30 ms hang watchdog that fired during normal first-frame GPU allocation. Now `30000` for the intended 30 s.

  * **Platform(s):** Windows 11
  * **Unreal Engine version(s):** CARLA UE5 fork 5.5.4 (`ue5-dev`)

Clean-configure builds with MSVC 14.38 and 14.44: 378/378 targets, LibCarla tests 96/98 (the two failures are localhost-TCP tests that fail in this environment without these changes), CarlaUnreal compiles with 978 UBT actions. A Development package boots headless, serves the Python API on port 2000, and advances frames with no startup freeze.

#### Possible Drawbacks

MSVC builds now report CARLA's own `/W4` warnings (previously buried in the `/Wall` flood). They are not fatal: `ENABLE_WARNINGS_TO_ERRORS` remains OFF.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9780)


JArmandoAnaya:
Ready for code review. This PR fixes the current Windows-side compilation issue with the minimum required changes.

JArmandoAnaya:
Moved to draft while I prepare a new fix to be added here. I'm validating that the fix is complete.

JArmandoAnaya:
Done! Pushed and validated the patch to prevent a compilation error reported by the user "Josh" in the general chat in Discord related to the Unreal Engine Carla's fork. The reported bug was real, and I was able to reproduce it locally. The last commit is to prevent that issue.

Blyron (approved):

PR #9782, opened 2026-06-10
fix(ros2): correct lidar publisher dispatch and realign sensor enum

LuisPovedaCano:
ProcessDataFromLidar dispatched GetOrCreateSensor on DataStream::GetSensorType(), which returns the per-stream id (a sequential counter), not an ESensors value. That routed lidar data to a wrong-typed (or missing) publisher, so the following dynamic_pointer_cast<CarlaLidarPublisher> yielded null and SetData segfaulted on a null this. Pin the type to RayCastLidar like the other ProcessDataFrom* helpers and guard the cast result. Ray-cast and HSS lidars share CarlaLidarPublisher, so fold the hss_lidar__ placeholder name into the ray-cast case to keep HSS topics named correctly.

Also realign the ESensors enum with the SensorRegistry tuple: the fisheye / wide-angle-lens cameras were added to SensorRegistry but were missing here, shifting every following value out of sync with the registry index.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9782)

### DIFF
--- a/LibCarla/source/carla/ros2/ROS2.cpp
+++ b/LibCarla/source/carla/ros2/ROS2.cpp
@@ -70,6 +70,14 @@ enum ESensors {
   SemanticSegmentationCamera,
   InstanceSegmentationCamera,
   WorldObserver,
+  // Keep these in lock-step with the SensorRegistry tuple order: the fisheye /
+  // wide-angle-lens cameras (ported in #9741) were added to SensorRegistry but
+  // were missing here, which shifted every following value out of sync with the
+  // registry index.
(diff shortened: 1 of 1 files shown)


JArmandoAnaya:
Hello @LuisPovedaCano, nice. I wonder if this change is necessary in the UE4 version as well.

LuisPovedaCano:
I just tried it in `ue4-dev` and is working there. Thanks for the suggestion.

Blyron (approved):

PR #9788, opened 2026-06-12
feat(LibCarla/ros2): [3/3] publish traffic light info and status topics

### LINKED ISSUE #9784: Extend native ROS2 interface with vehicle state, TF tree, traffic lights and map topics (parity with ros-bridge data outputs)

The native ROS2 interface (`LibCarla/source/carla/ros2`, enabled with `--ros2`) currently behaves as a sensor-out / control-in pipe: it publishes raw sensor streams (cameras, lidar, semantic lidar, radar, IMU, GNSS, collision, lane invasion), `/clock` and per-sensor transforms, and subscribes to `vehicle_control_cmd` (see `ROS2::GetOrCreateSensor` dispatch and the `ros2 topic list` output in #9408).

This is not enough to run standard autonomous driving stacks against CARLA without a sidecar bridge. Compared with the legacy `carla-simulator/ros-bridge` (pinned to 0.9.13 and effectively unmaintained for newer CARLA), the native interface is missing the data outputs a real drive-by-wire vehicle would produce:

- Ego vehicle status (`CarlaEgoVehicleStatus`) and static vehicle info (`CarlaEgoVehicleInfo`)
- `nav_msgs/Odometry`
- A REP-105 compliant TF tree (`map → odom → base_link`); today only sensor transforms are published
- Traffic light status/info
- Object-level ground truth
- A latched OpenDRIVE map topic (`std_msgs/String`, `TRANSIENT_LOCAL`)

Concretely: Autoware cannot run without odometry, vehicle status and a proper TF tree; Nav2 cannot run without `odom → base_link` and `nav_msgs/Odometry`. Users are forced back onto the deprecated external bridge or custom shims (#9592 asks exactly what the native/bridge differences are).

Port the missing *data* features from ros-bridge into the native ROS2 layer on `ue4-dev` (where the most mature implementation and active rework lives, cf. #9450/#9451), as a chain of small, independently mergeable PRs:

1. **Latched OpenDRIVE map publisher** — `std_msgs/String` on `/carla/map` with `TRANSIENT_LOCAL` durability on the DataWriter, mirroring ros-bridge's `opendrive_sensor` semantics. Trivial, no new types beyond String.
2. **Ego vehicle status + vehicle info publishers** — type definitions wire-compatible with `ros-carla-msgs`, so the ROS side deserializes with the existing package.
3. **Odometry + TF tree** — `nav_msgs/Odometry` publisher and `map → odom → base_link` transforms, composing geometry sub-types already present in `ros2/types/`. Acceptance criterion: Nav2 brings up its costmap against CARLA with no shim node.
4. **Traffic light status/info publishers** — exposed as perceived data, opt-in like sensors via `enable_for_ros()`-style registration.
5. **Object ground truth** — proposed as a flat custom type in `ros-carla-msgs` rather than `derived_object_msgs/ObjectArray` (see alternatives below).

All of this follows the existing implementation pattern: hand-written CDR-serializable structs in `LibCarla/source/carla/ros2/types/` with correct DDS type names (`pkg::msg::dds_::Type_`) and `rt/` topic mangling — no rclcpp, no rosidl, no ROS build environment inside CARLA. The ROS-side contract is only that users install `ros-carla-msgs` for the CARLA-specific types.

Suggested validation: a CI smoke test using a stock ROS2 container that subscribes to each new topic and asserts successful deserialization, since wire-format mistakes fail silently rather than at build time.

**Describe alternatives you've considered**

- **Updating the external ros-bridge for current CARLA**: it is pinned to 0.9.13, the Python-API surface has drifted, and maintaining a second full interface duplicates effort. The maintainer direction (ros-bridge PR #762) already points toward the built-in interface.
- **Bridging via a client-side ROS node**: works, but reintroduces the latency and tick-coupling the native implementation was built to avoid, and every user re-implements the same shim.
- **Using `derived_object_msgs` for objects**: rejected for the first iteration; hand-rolling its nested CDR layout (shape_msgs, polygons) without rosidl is the most error-prone option. A flat type in `ros-carla-msgs` achieves the same with far less risk.
- **Also porting simulator-control features (spawn/destroy services, weather, sync tick control, debug markers)**: explicitly out of scope by design. The native ROS interface should expose only data a real vehicle would produce and accept only control commands; simulation control remains the job of the Python API.

**Additional context**

- Scope boundary proposed: *native ROS2 = real-vehicle-like data out + control commands in*. Everything else stays in the Python API.
- Map conversion for downstream stacks (Lanelet2 `.osm` for Autoware, `.pcd` point clouds, OccupancyGrid for Nav2/RViz) is intentionally **not** part of this request — it belongs in offline/client-side tooling, with the published OpenDRIVE string as the authoritative source.
- Related: #9408 (control cmd issues, topic list evidence), #9592 (native vs bridge feature comparison request), #9289/#9294 (configurable DDS backend), #9551 (Fast-DDS version vs Humble/Jazzy), #9450/#9451 (ROS2 handling rework), #9735 (Zenoh middleware proposal).
- Happy to split this into a tracking issue + per-phase PRs, and to coordinate with the #9451 rework so publishers land on the refactored handling rather than the old paths.

---

**Planned PR series (design review concluded 2026-06-11)**

Two scope amendments after reviewing the implementation against the consumers:

- **Object-level ground truth is dropped from this series.** Nav2 has no object-list input, Autoware planning consumes its own `autoware_perception_msgs` types so a conversion node is needed regardless of what CARLA publishes, RViz cannot display a custom list natively, and benchmarking users already get time-aligned poses, velocities and bounding boxes from `world.get_actors()` in the Python API. It was also the only feature here requiring a brand-new message definition. It can be revisited later as its own proposal if a concrete consumer appears.
- **Traffic light topics will be always-on when `--ros2` is enabled** instead of opt-in: traffic lights are world actors users never spawn, so there is no natural per-actor opt-in hook, and the traffic cost is negligible (info latched once per map, status only on state change).

With those amendments, no new message definitions are needed anywhere: every topic uses either standard ROS 2 types (`std_msgs/String`, `nav_msgs/Odometry`, `tf2_msgs/TFMessage`) or types already published in [ros-carla-msgs](https://github.com/carla-simulator/ros-carla-msgs) from the bridge era, the same contract the existing `vehicle_control_cmd` subscriber uses. No new CARLA flags, settings or actor attributes are introduced; activation stays `--ros2` / `--rmw=`.

| PR | Branch | Scope |
|----|--------|-------|
| 1 | `feature/ros2-qos-durability-and-map` | TRANSIENT_LOCAL durability support in the middleware layer (FastDDS, CycloneDDS, Zenoh) + latched `std_msgs/String` OpenDRIVE map on `/carla/map`, published once per episode |
| 2 | `feature/ros2-ego-vehicle-topics` | `nav_msgs/Odometry` on `/carla/<ros_name>/odometry`, REP-105 TF tree (static `map -> odom` identity on `/tf_static`, dynamic `odom -> <vehicle>` on `/tf`), `CarlaEgoVehicleStatus` (per frame) and `CarlaEgoVehicleInfo` (latched) |
| 3 | `feature/ros2-traffic-lights` | `CarlaTrafficLightInfoList` on `/carla/traffic_lights/info` (latched, once per map) + `CarlaTrafficLightStatusList` on `/carla/traffic_lights/status` (published only on state change), plus docs and CHANGELOG for the whole series |

All branches off `ue4-dev`. Each PR compiles standalone with all LibCarla tests green. Acceptance criterion for PR 2 stays as stated above: Nav2 brings up its costmap against CARLA with no shim node.

JArmandoAnaya:
#### Description

With the map, the ego vehicle and the transform tree already covered by the previous pull requests of this series, one piece of the world remained invisible to the native ROS 2 interface: traffic lights. The simulator knows every light and its state at all times, but none of that reaches the ROS graph, so a planner that needs to stop at a red light has to fetch it through the Python API or fall back on the deprecated ros-bridge.

This pull request makes the simulator describe its traffic lights whenever ROS 2 is enabled, with no opt-in step, on two latched topics:

* `/carla/traffic_lights/info`, published once per episode, carries the static picture of every light: its id, its pose in the world and its trigger volume, the same data the Python API exposes as `actor.trigger_volume`.
* `/carla/traffic_lights/status` carries the current state of every light and is republished only when a state changes. Because the last sample is latched, a node that starts between changes still receives the current state immediately instead of waiting for the next transition.

A late joiner can check both at any time:

```bash
ros2 topic echo --qos-durability transient_local --once /carla/traffic_lights/info
ros2 topic echo --qos-durability transient_local --once /carla/traffic_lights/status
```

Both topics reuse the message definitions already published in `ros-carla-msgs`, the same contract the rest of the interface follows, and the state encoding matches what the bridge published. Nothing new is asked from the user: no flags, no settings, no attributes, and no ROS environment takes part in the build.

As the closing piece of the series, this pull request also adds `Docs/ros2/topics.md`, a reference of every topic the native interface publishes and subscribes to, and the consolidated changelog entry covering the three pull requests. It builds on top of #9787, whose commits it includes until that one merges; only the last commit is new here.

Closes #9784

  * **Platform(s):** Linux (Ubuntu 20.04 build container, Ubuntu host)
  * **Python version(s):** 3.12
  * **Unreal Engine version(s):** 4.26 (CARLA fork)

Validation: `make LibCarla ARGS="--ros2"` and `make check.LibCarla ARGS="--ros2"` with 167 server and 66 client tests passing in debug and release, including new tests for the message types, the wire serialization and the change detection; `make CarlaUE4Editor` with no new warnings; a full `make package` build; the ROS 2 smoke tests against the packaged simulator (8 of 8 passing, including a new traffic light lifecycle test that forces state changes and reloads the world); and a check from a ROS 2 Humble container confirming both topics appear with the expected types and transient local durability.

#### Possible Drawbacks

* The status list is keyed by actor id, which changes between episodes; consumers should pair it with the info topic of the same episode, exactly as they did with the bridge.
* A frozen light publishes nothing after its last change, by design; the latched sample keeps its current state available to late joiners.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9788)


youtalk (commented):
Good closing piece: `info` latched once and `status` republished only on change is the right cadence, and `Docs/ros2/topics.md` is an excellent, much-needed contract reference. Same item as #9787 for the new `CarlaTrafficLight*` types (golden-hash pinning), and the same request to include Jazzy in validation. One more suggestion: add a smoke test in `PythonAPI/test/smoke/test_ros2.py` that subscribes to these topics with the real `ros-carla-msgs` package installed and asserts a successful decode — the deserialization check #9784 proposed; it catches wire drift a hash test alone won't.

LibCarla/source/test/server/test_type_hash.cpp
```diff
@@ -38,12 +38,21 @@ static void check_hash(const char* hash, const std::string& type_label) {
...
+  CHECK_HASH(CarlaBoundingBox);
   CHECK_HASH(CarlaCollisionEvent);
   CHECK_HASH(CarlaEgoVehicleControl);
+  CHECK_HASH(CarlaEgoVehicleInfo);
+  CHECK_HASH(CarlaEgoVehicleInfoWheel);
+  CHECK_HASH(CarlaEgoVehicleStatus);
   CHECK_HASH(CarlaLineInvasion);
+  CHECK_HASH(CarlaTrafficLightInfo);
+  CHECK_HASH(CarlaTrafficLightInfoList);
+  CHECK_HASH(CarlaTrafficLightStatus);
```
youtalk, line 56: **[important]** Extend the golden-hash pinning (see #9787) to the new `CarlaTrafficLight*` types, per distro, so the traffic-light contract is verified against the real `ros-carla-msgs` package rather than only self-consistent.

JArmandoAnaya: Applied in 5102ace9d, same flow as #9787: built ros-carla-msgs master in the osrf/ros:jazzy-desktop image and pinned the five new hashes (CarlaBoundingBox and the four CarlaTrafficLight* types) in the golden test, so all 40 specializations are now checked against the canonical packages. The computed values matched the ones already pinned in CdrTopicInfo.h, so there is no drift today. On the per-distro point: the RIHS01 hash is a function of the message definition only, and Humble predates REP-2011 (it ships no hash metadata and never validates it), so the Jazzy values are the golden set for every distro that checks them.

PR #9790, opened 2026-06-17
Fix NumPy alias usage in PythonAPI examples

yinli-systems:
## Summary
- replace removed `np.bool` aliases with `np.bool_` in DVS event dtypes
- replace removed `np.int` aliases with `int` in lidar-to-camera coordinate/color conversions

These aliases raise `AttributeError` with NumPy 2.x, which breaks the affected PythonAPI examples before they can run.

## Tests
- `rg -n "np\.(bool|int)\b" PythonAPI -S` (no matches)
- `python3 -m py_compile PythonAPI/examples/manual_control.py PythonAPI/examples/manual_control_carsim.py PythonAPI/examples/manual_control_chrono.py PythonAPI/examples/lidar_to_camera.py`
- `python3` NumPy dtype/conversion smoke check for `np.bool_` and `astype(int)`
- `git diff --check`

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9790)

### DIFF
--- a/PythonAPI/examples/lidar_to_camera.py
+++ b/PythonAPI/examples/lidar_to_camera.py
@@ -231,8 +231,8 @@ def tutorial(args):
             intensity = intensity[points_in_canvas_mask]
 
             # Extract the screen coords (uv) as integers.
-            u_coord = points_2d[:, 0].astype(np.int)
-            v_coord = points_2d[:, 1].astype(np.int)
+            u_coord = points_2d[:, 0].astype(int)
+            v_coord = points_2d[:, 1].astype(int)
 
             # Since at the time of the creation of this script, the intensity function
             # is returning high values, these are adjusted to be nicely visualized.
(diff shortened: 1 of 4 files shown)


JArmandoAnaya:
Hello! I think the `master` branch is no longer in use. You need to compile from the current default, `ue5-dev`, or from the UE4 version in the branch `ue4-dev`. Your suggestions are mostly already implemented on those branches.

Blyron:
Master branch is not used anymore, closing this PR

yinli-systems:
Thanks for the clarification. I checked both active branches (`ue5-dev` and `ue4-dev`) and confirmed the NumPy alias issue from this PR is already covered there, including the existing `PythonAPI/test/unit/test_numpy_compat.py` regression guard. Closing this master-based PR makes sense.

I’ve opened future CARLA work against `ue5-dev` instead: https://github.com/carla-simulator/carla/pull/9791

PR #9807, opened 2026-07-07
Add ROS 2 message types and CDR serialization foundation

JArmandoAnaya:
> **Retargeting status for `ue58-dev`:** Most of the changes described below are already present in the targeted branch. This retargeted PR contains only one remaining line in `LibCarla/CMakeLists.txt`, which excludes `test_ros2_serialization.cpp` from the non-ROS2 test glob. The remaining PRs in this series are no longer needed because their changes are already present in `ue58-dev`, so they are being closed.
>
> **Series references:** #9808, #9809, #9810, #9811, #9813, #9814, #9815, #9816.

#### Description

First PR of the ROS 2 middleware decoupling series for ue5-dev (tracking issue [#9762](https://github.com/carla-simulator/carla/issues/9762)): the middleware-neutral type layer, ported from the merged ue4-dev series (#9612, #9643, #9681). It adds 31 plain C++ message structs under `carla/ros2/types/msg/`, unified Fast-CDR serialization (`CdrSerialization.h`), per-type topic metadata with REP-2011 RIHS01 type hashes (`CdrTopicInfo.h`), and the REP-2016 USER_DATA payload builder (`UserDataFormat.h`). The first commit also raises the `carla-ros2-native` shared library to C++20: it is configured as a separate CMake process, so it never inherited the project-wide standard and silently compiled at the Clang default (C++17).

The new headers are not referenced by any runtime build target yet; live publishers continue to use the generated FastDDS types until the middleware abstraction lands in the next PR in the series. No behavior change.

UE5 adaptation: ue5-dev pins FastDDS 2.11.2 (Fast-CDR 1.x), so `CdrSerialization.h` keeps the Fast-CDR 1.x API spellings instead of the 2.x forms ue4-dev carries since its Fast-DDS 2.14.6 upgrade (#9789). Five lines differ; the golden bytes pin the wire format, which is identical (classic CDR, little-endian).

#### What this PR changes

- `Ros2Native/LibCarlaRos2Native/CMakeLists.txt` and `Ros2Native/CMakeLists.txt`: `carla-ros2-native` builds at C++20.
- 34 new header-only files under `LibCarla/source/carla/ros2/types/`.
[...]

Issue #9762, opened 2026-05-27
Port the ROS 2 DDS-middleware decoupling (FastDDS + CycloneDDS + Zenoh, --rmw=) from ue4-dev to ue5-dev


JArmandoAnaya:
**CARLA version:** 0.10.x (ue5-dev)
**Platform/OS:** Ubuntu 22.04 / 24.04 (Linux), Windows 11
**Target branch:** ue5-dev

## Summary

The ROS 2 DDS-middleware decoupling that shipped on `ue4-dev` (PRs #9608, #9612, #9619, #9620, #9643, #9644, #9645, plus the follow-ups #9665, #9666, #9669, #9672, #9681, #9692, #9711) is not present on `ue5-dev`. This issue tracks porting it so the UE5 simulator can run its native ROS 2 connector on FastDDS, CycloneDDS, or Zenoh, selectable at runtime via `--rmw=`.

Two later `ue4-dev` PRs build on top of the decoupling and are included in this port as separate trailing PRs, so the core porting strategy is unchanged: #9735 (merged) adds Zenoh as a third middleware (`--rmw=zenoh`), and #9778 (merged) adds a configurable ROS 2 domain id via `--ros-domain-id`. Both are now merged on `ue4-dev`, so no PR in the series is blocked on upstream.

On `ue4-dev` the work replaced the hard FastDDS dependency with a strategy-pattern abstraction (`Middleware` enum, `MiddlewareFactory`, `IPublisherMiddleware` / `ISubscriberMiddleware`), introduced plain-old-data message structs serialized through a single unified CDR path (`CdrSerialization` + a generic `GenericCdrPubSubType` that replaces all the per-type generated classes), added a CycloneDDS implementation over raw `dds_writecdr()`, and removed the fastddsgen-generated type files.

Porting this to `ue5-dev` is not a cherry-pick, because the two branches build ROS 2 in fundamentally different ways. This issue records the analysis and the consolidated plan.

## The architecture gap (why this is an adaptation)

| Concern | ue4-dev | ue5-dev |
|---|---|---|
| Where ROS 2 code compiles | statically into `carla-server` | into a standalone shared library `libcarla-ros2-native.so` (the `Ros2Native/` ExternalProject) |
| DDS vendor libraries | static archives linked into `carla-server` | shared libraries built by ExternalProject, copied to the plugin `Binaries`, declared as runtime dependencies and loaded with `AddDynamicLibrary` |
[...]

PR #9681, opened 2026-04-15
fix(LibCarla/ros2): add REP-2011 type hashes and shared FastDDS participant for Jazzy compatibility

JesusAnaya:
#### Description

Fixes two ROS 2 Jazzy interoperability problems in the native DDS integration.

##### 1. Missing REP-2011 type hash on every topic (#9674)

ROS 2 Iron and later RMWs (`rmw_fastrtps_cpp`, `rmw_cyclonedds_cpp`) parse the DDS `USER_DATA` QoS field during SEDP/PDP endpoint discovery and expect a REP-2016 KV payload of the form `typehash=RIHS01_<64-hex>;`. When the payload is absent, the RMW logs for every discovered endpoint:

```
[WARN] [rmw_cyclonedds_cpp]: Failed to parse type hash for topic 'rt/...'
```

CARLA did not set `USER_DATA` on any of its 31 message types, so running a Jazzy subscriber (rviz2, any `ros2 topic echo`, any autonomy stack) against a CARLA server produced a warning per topic per endpoint, and prevented future strict type-hash-based endpoint matching.

Fix: hardcode a REP-2011 RIHS01 hash per message type in `CdrTopicInfo<T>::type_hash()` and attach it via a new `UserDataFormat.h::build_user_data_for<T>()` helper on every `DataWriter` and `DataReader` QoS in both FastDDS and CycloneDDS middlewares. Hashes are hardcoded because CARLA has no IDL code-generation step at build time and recomputing RIHS01 correctly requires `rosidl_generator_type_description` inside a ROS 2 build environment. The values are stable per distro for every standard message (`std_msgs`, `geometry_msgs`, `sensor_msgs`, `nav_msgs`, `builtin_interfaces`, `tf2_msgs`, `rosgraph_msgs`, `ackermann_msgs`) and match on Humble and Jazzy.

A helper script `Util/ros2/compute_type_hash.sh` computes the RIHS01 hash for any message type using only Docker, so adding a new type needs no local ROS 2 install.

##### 2. FastDDS `rt/tf` duplicate-topic crash (#9675)

Since the shared `DomainParticipant` was introduced, multiple sensor publishers on the same actor (or multiple hero actors) all call `create_topic("rt/tf", ...)` on the same participant. FastDDS rejects duplicate topic creation and aborts the server:

```
Topic with name : rt/tf already exists -> Function create_topic
ERROR: FastDDSPublisherMiddleware: Failed to create Topic
[...]

PR #9643, opened 2026-04-05
feat(LibCarla/ros2): [5/7] add unified CDR serialization and GenericCdrPubSubType

JesusAnaya:
### Description

Relates to #9294 - DDS middleware decoupling PR series. Blocked on https://github.com/carla-simulator/carla/pull/9620.

#### What this PR does

Introduces a vendor-agnostic CDR serialization layer that bridges the 31
`carla::ros2::msg::*` POD structs to the DDS wire format used by ROS2.

**New files (4):**

| File | Purpose |
|------|---------|
| `LibCarla/source/carla/ros2/types/CdrSerialization.h` | `serialize_to_cdr<T>()` / `deserialize_from_cdr<T>()` for all 31 msg types. Uses Fast-CDR XCDR1 LE with DDS encapsulation header. Handles primitives, strings, arrays, vectors, and nested structs. |
| `LibCarla/source/carla/ros2/types/CdrTopicInfo.h` | `CdrTopicInfo<T>` template specializations: `type_name()` (ROS2-compatible DDS type name) and `max_serialized_size()` (conservative CDR payload bound without encapsulation). |
| `LibCarla/source/test/server/test_dds_middleware.cpp` | +14 CDR tests: `cdr_topic_info` (2) + `cdr_serialization` (12 round-trips). |
| `LibCarla/cmake/test/CMakeLists.txt` | Add FastCDR include path, `libfastcdr.a` link, and `-fexceptions` to the server test binary (Fast-CDR templates use `try/catch`). |

#### Serialization architecture

```
carla::ros2::msg::X  (vendor-agnostic POD struct)
         |
         v
  CdrSerialization.h — serialize_to_cdr() / deserialize_from_cdr()
         |
    CDR byte buffer (XCDR1 LE, wire-compatible with all ROS2 distros)
         |
    +----+----+
    |         |
    v         v
FastDDS    CycloneDDS
write()    dds_writecdr()
(Pending PR)      (Pending PR)
```

#### Tests

- Server: 97 to 111 tests (all pass)
- Client: 56/56 (unchanged)

Run CDR tests only:
```sh
./Build/libcarla-server-build.debug/test/libcarla_test_server_debug \
  --gtest_filter='*cdr_*'
```

### PR series

[...]

PR #9808, opened 2026-07-07
[2/9] Add ROS 2 middleware abstraction and FastDDS-on-CDR middleware

JArmandoAnaya:
#### Description

Second PR of the ROS 2 middleware decoupling series for ue5-dev (tracking issue [#9762](https://github.com/carla-simulator/carla/issues/9762)): the vendor-neutral middleware strategy layer and its FastDDS implementation over the unified CDR serializer from the previous PR. Ported from the merged ue4-dev series (#9608, #9643, #9681). Everything compiles into `libcarla-ros2-native.so`; no DDS header reaches `carla-server`.

The layer adds the `Middleware` enum and factory, the publisher/subscriber strategy interfaces, `GenericCdrPubSubType` (one FastDDS `TopicDataType` that serializes any `carla::ros2::msg::*` struct through `CdrSerialization.h`, replacing the generated per-type PubSubType classes), the FastDDS publisher/subscriber middlewares, and a refcounted shared `DomainParticipant`. It also adds the DDS-free `SetActiveMiddleware` bridge so `ROS2.cpp` can select a middleware without pulling DDS symbols across the shared-library boundary.

The existing publishers are untouched and keep using the generated FastDDS types. The cutover, including the `PublisherImpl`/`SubscriberImpl` rewrite and its 14 tests, lands in the next PR of the series. No behavior change.

Depends on [#9807](https://github.com/carla-simulator/carla/pull/9807); review after it merges. The diff below includes that PR's commits until then.

#### What this PR changes

- 10 new files under `LibCarla/source/carla/ros2/middleware/`: `Middleware.h`, the two interfaces, `MiddlewareFactory.h`, `ActiveMiddleware.{h,cpp}`, and `fastdds/{GenericCdrPubSubType,FastDDSPublisherMiddleware,FastDDSSubscriberMiddleware,FastDDSSharedParticipant}`.
- `Ros2Native/LibCarlaRos2Native/CMakeLists.txt`: globs the new `middleware/` sources into `carla-ros2-native` and defines `CARLA_ROS2_MIDDLEWARE_FASTDDS` on that target only.
- Renames `test_ros2_serialization.cpp` to `test_ros2_middleware.cpp` and adds 32 cases (middleware enum/factory helpers, `GenericCdrPubSubType` round-trips through a real `SerializedPayload_t`, and large-payload sizing). No DDS participant is created.
- `LibCarla/CMakeLists.txt` and `Ros2Native/CMakeLists.txt`: when `ENABLE_ROS2` is ON, `libcarla_test_server` links `libfastrtps.so` (added as `INSTALL_BYPRODUCTS` on the `fastdds` ExternalProject) for the `TopicDataType` base and `SerializedPayload_t`, with `-rpath-link`/`-rpath`/`--disable-new-dtags` so its private foonathan_memory dependency resolves; when OFF the suite builds exactly as before.

#### PR series

| PR | Scope | Status |
|----|-------|--------|
| POD message types + unified CDR serialization foundation | `types/msg/*` POD structs, `CdrSerialization`/`CdrTopicInfo`/`UserDataFormat`, CDR round-trip GTests | #9807 |
| Middleware abstraction + FastDDS-on-CDR middleware | strategy interfaces + factory, FastDDS middleware over `GenericCdrPubSubType`, shared participant, `ActiveMiddleware` bridge | this PR |
| Cutover: route publishers/subscribers through the abstraction | rewrite `PublisherImpl`/`SubscriberImpl` to delegate through the factory, migrate publisher traits to the POD types, `ROS2.cpp` calls `SetActiveMiddleware` | not started |
| Remove dead generated FastDDS type files | delete the legacy generated `types/*` set | not started |
| CycloneDDS middleware + vendor shared-library build | CycloneDDS sertype + middleware, CycloneDDS ExternalProject, plugin runtime deps | not started |
| Runtime `--rmw=` selection, plugin integration, examples, docs | `CarlaSettings`/`CarlaEngine` wiring, launch flag, examples, docs, consolidated CHANGELOG | not started |
| Zenoh middleware (`--rmw=zenoh`) | `middleware/zenoh/*`, factory branch, zenoh-c build in `Ros2Native`, rviz example, tests | not started |
| Configurable ROS 2 domain id (`--ros-domain-id`) | `MiddlewareConfig`, resolution `--ros-domain-id` > `ROS_DOMAIN_ID` > 0, honored by all middlewares, docs + example + tests | not started |

Related: #9762, #9807

  * **Platform(s):** Linux (Ubuntu 24.04 host). `libcarla_test_server` 196/196 (including the 32 new middleware cases); `libcarla_test_client` 193/193, unaffected; `ENABLE_ROS2=OFF` builds the suite with no ROS 2 cases; `package-development` build green.
  * **Python version(s):** n/a (no Python API change).
  * **Unreal Engine version(s):** 5.5 (CARLA fork).

#### Possible Drawbacks

- None for runtime: the layer is compiled into the shared library but not yet on any publish path, so existing FastDDS publishing is unchanged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9808)


youtalk (commented):
Thank you for the quick PR work! One correctness note inline below.

LibCarla/source/carla/ros2/middleware/fastdds/FastDDSSubscriberMiddleware.h
```diff
@@ -0,0 +1,180 @@
...
+      efd::DataReader* reader,
+      const efd::SubscriptionMatchedStatus& info) override {
+    _alive.store(info.current_count > 0, std::memory_order_relaxed);
+  }
+
+  void on_data_available(efd::DataReader* reader) override {
+    efd::SampleInfo info;
+    erc rcode = reader->take_next_sample(_message_ptr, &info);
+    if (rcode == erc::ReturnCodeValue::RETCODE_OK) {
+      *_new_message_ptr = true;
```
youtalk, line 67: **Guard on `info.valid_data` before flagging a new message.** `take_next_sample` returns `RETCODE_OK` for dispose/unregister notifications (`valid_data == false`); as written, a publisher going away delivers a default-constructed message (e.g. a zeroed `CarlaEgoVehicleControl`) as a real command. The pre-refactor `SubscriberImpl` and the CycloneDDS/Zenoh middlewares all check `valid_data` — only this path dropped it.

```suggestion
      if (info.valid_data) *_new_message_ptr = true;
```

germanros1987:
First of all, thanks for this series — we cherry-picked all nine PRs ([1/9]…[9/9], head of #9816) onto our UE 5.8 integration branch (`ue58-dev`, PR #9826 lineage) and everything built and passed its tests on the first try (261/261 server tests, 5/5 `test_ros2.py` smoke, all three middlewares verified against ROS 2 Humble containers).

While validating we hit one trap that we think is worth fixing in the series itself, and we'd like to propose a small patch.

## Symptom

With `--rmw=fastdds`, a subscriber in a Docker container (`--net=host`) sees every CARLA topic in `ros2 topic list`, but `ros2 topic echo` never receives a single sample — `/clock` included — and **nothing is logged on either side**. The `test_ros2.py` smoke suite cannot catch this by design, since it validates the server-side publish path without an external subscriber. The rviz example works only because `run_rviz.sh` mounts `fastrtps-profile.xml`, which forces the *subscriber* onto UDP.

## Root cause — three independent silent failure modes

The shared `DomainParticipant` uses `PARTICIPANT_QOS_DEFAULT`, so Fast-DDS instantiates UDPv4 + SHM and **prefers SHM for any peer that looks same-host**. Discovery is UDP multicast, so it always works — which is exactly what makes the failure invisible. We ran a controlled matrix (live simulator, Fast-DDS 2.14.6, continuous ticking, `ros2 topic echo /clock --once` from containers):

| Subscriber | Result |
|---|---|
| Humble (Fast-DDS 2.6), `--net=host` | ✗ no data |
| Humble, `--net=host --ipc=host` | ✗ no data |
| Humble, `--net=host --ipc=host --user <host uid>` | ✗ no data |
| Jazzy (Fast-DDS **2.14.6**, same as CARLA's), `--net=host` | ✗ no data |
| Jazzy, `--net=host --ipc=host` (root) | ✗ no data |
| Jazzy, `--net=host --ipc=host --user <host uid>` | ✓ **data (SHM)** |
| Humble with the shipped UDP profile (control) | ✓ data (UDP) |

So there are three *independent* ways the SHM path black-holes, and any one of them is enough:

1. **IPC namespace** — default Docker gives the container its own `/dev/shm`; the segments simply aren't shared.
2. **uid mismatch** — Fast-DDS SHM segment files are created `0644`, and the writer must write into the *reader's* port segment. A container running as root against a simulator running as a regular user (or any two different-uid processes on the same host) fails this silently.
3. **SHM generation** — Humble's Fast-DDS 2.6 cannot exchange shared memory with 2.14 even with IPC and uid fully aligned.

Crucially, Fast-DDS does **not** fall back to the reader's UDP locator when the SHM path it selected is unreachable, and RELIABLE QoS doesn't recover it because ACKs/NACKs route over the same dead path.

## Proposed fix

Default the shared participant to UDPv4-only, with the standard `FASTDDS_BUILTIN_TRANSPORTS` environment variable as the opt-out (when it's set, skip the override entirely so Fast-DDS honors it natively), and log the active transport next to the middleware line:

```diff
--- a/LibCarla/source/carla/ros2/middleware/fastdds/FastDDSSharedParticipant.cpp
+++ b/LibCarla/source/carla/ros2/middleware/fastdds/FastDDSSharedParticipant.cpp
@@ -10,6 +10,9 @@
 
 #include <fastdds/dds/domain/DomainParticipantFactory.hpp>
 #include <fastdds/dds/domain/qos/DomainParticipantQos.hpp>
+#include <fastdds/rtps/attributes/BuiltinTransports.hpp>
+
+#include <cstdlib>
 
@@ -26,6 +29,18 @@ efd::DomainParticipant* FastDDSSharedParticipant::acquire() {
   std::lock_guard<std::mutex> lock(_mutex);
   if (_refcount == 0u) {
     efd::DomainParticipantQos pqos = efd::PARTICIPANT_QOS_DEFAULT;
+    // Fast-DDS's builtin default is UDPv4 + shared memory, and SHM is
+    // preferred for any peer that looks same-host. SHM delivery silently
+    // drops every sample across a container, uid, or Fast-DDS-version
+    // boundary (discovery is UDP multicast, so topics stay visible while no
+    // data arrives), and Fast-DDS never falls back to the reader's UDP
+    // locator when the SHM path is unreachable. Default to UDPv4-only so any
+    // subscriber works out of the box; setting the standard
+    // FASTDDS_BUILTIN_TRANSPORTS environment variable (e.g. to DEFAULT)
+    // skips this override and restores SHM for same-host, same-uid setups.
+    if (std::getenv("FASTDDS_BUILTIN_TRANSPORTS") == nullptr) {
+      pqos.setup_transports(eprosima::fastdds::rtps::BuiltinTransports::UDPv4);
+    }
     // Use the effective domain id resolved by the abstraction layer
```

plus a corresponding `UE_LOG` pair in `FCarlaEngine::NotifyInitGame` (`"ROS2: Fast-DDS transport: UDPv4 only (default; set FASTDDS_BUILTIN_TRANSPORTS=DEFAULT to re-enable shared memory)."` / `"…set from FASTDDS_BUILTIN_TRANSPORTS='%s'."`) and a docs section with the compatibility table.

Verified on our branch (`97f453bbd`, `5332be6a1`):

* 261/261 server tests unchanged;
* **bare, unconfigured** `ros:humble` and `ros:jazzy` containers now receive `/clock` and full camera images with zero container-side configuration;
* `FASTDDS_BUILTIN_TRANSPORTS=DEFAULT` restores stock behavior exactly — the bare-container black-hole reappears and the aligned same-uid `--ipc=host` Jazzy container delivers over genuine SHM, so the performance path is intact for setups that can actually use it.

The trade-off is losing SHM zero-copy for same-host, same-uid native subscribers *by default* — but anyone who measurably needs it is exactly the user who can set one standard eProsima env var, whereas the current default fails silently for every containerized consumer. CycloneDDS is unaffected (no SHM support compiled in — it already behaves like this proposal), which also means the two middlewares currently differ in out-of-the-box behavior; this patch makes them consistent.

Happy to open a PR against the series branch if that's easier — the full commits are on `ue58-dev` (`97f453bbd` code, `5332be6a1` docs matrix).

Issue #9635, opened 2026-04-02
Port ue4-dev features and fixes to ue5-dev (tracking issue)


youtalk:
**Is your feature request related to a problem? Please describe.**

Many useful bug fixes, features, and improvements have accumulated in `ue4-dev` but have not been ported to `ue5-dev`. This is the parent tracking issue for porting them as new implementations (not cherry-picks). Each category is tracked in its own sub-issue. See https://github.com/carla-simulator/carla/discussions/9581 for the original discussion.

**Describe the solution you'd like**

### Progress by Category

- [x] #9583
- [x] #9627
- [ ] #9628
- [ ] #9629
- [ ] #9630
- [ ] #9631
- [ ] #9632
- [ ] #9633
- [ ] #9634

### Submission Schedule

| Phase | PRs | Dependencies |
|-------|-----|--------------|
| Phase 1 | #0a, #0b, #0c, #0d, #0s | Independent |
| Phase 2 | #0k, #0l, #0t, #0i, #0j, #0m, #0q | Independent |
| Phase 3 | #0e → #0f → #0g / #0h | Sequential (ROS2 refactor chain) |
| Phase 4 | #0u, #0p | Independent |
| Phase 5 | #0n, #0o, #0r | Large — phased submission |

**Describe alternatives you've considered**

Cherry-picking from ue4-dev was considered but rejected due to UE4→UE5 differences in build system, Unreal plugin structure, and rendering pipeline.

**Additional context**

Reference commits are from `ue4-dev` and serve as the basis for new `ue5-dev` implementations. All PRs within each phase are independent and can be reviewed in parallel unless noted otherwise.

youtalk:
@Blyron cc @JesusAnaya Based on the feedback https://github.com/carla-simulator/carla/pull/9638#issuecomment-4200609988, I have limited the following PRs to only fixing existing known bugs from `ue4-dev`, excluding any changes that would require major refactoring:

- https://github.com/carla-simulator/carla/pull/9649
- https://github.com/carla-simulator/carla/pull/9651
- https://github.com/carla-simulator/carla/pull/9652
- https://github.com/carla-simulator/carla/pull/9653
- https://github.com/carla-simulator/carla/pull/9654
- https://github.com/carla-simulator/carla/pull/9655
- https://github.com/carla-simulator/carla/pull/9656
- https://github.com/carla-simulator/carla/pull/9657
- https://github.com/carla-simulator/carla/pull/9658

[...]

PR #9826, opened 2026-07-21
Port CarlaUnreal project and plugins to UE 5.8 APIs

germanros1987:
Build system:
- Toolchain.cmake: UE 5.8 no longer bundles libc++ under ThirdParty/Unix/LibCxx; fall back to the clang sysroot's include/ and lib64/ when the legacy layout is absent.
- CarlaUnreal.Build.cs: 5.8 UBT stops adding flat-layout module roots to the include path; add ModuleDirectory explicitly so the generated PCH resolves CarlaUnreal.h.
- Carla.Build.cs: keep -Wshadow at warning level (5.8 escalates it to error; V2X/PathLossModel shadows fields extensively).

API replacements:
- TMap iterators no longer expose Value(): use It->Value (CarlaServer, CarlaRecorder, CarlaReplayerHelper, FrameData).
- UWorld line-batcher fields replaced by GetLineBatcher(ELineBatcherType) (CustomTerrainPhysicsComponent, DebugShapeDrawer).
- TArray SetNum/Pop bool argument replaced by EAllowShrinking (StreetMapComponent, MeshToSplineActor).
- UPackage::SavePackage flags overload removed: use FSavePackageArgs (CarlaToolsFunctionLibrary).
- ALandscapeProxy::Import takes an InImportLayers view (MapGeneratorWidget).
- FCoreUObjectDelegates::OnObjectSaved removed: broadcast OnObjectPreSave with the save context (CityMapGenerator).
- FlushPendingDeleteRHIResources_GameThread and ALandscape::bUseCompressedHeightmapStorage removed outright.
- FTexture2DRHIRef alias removed: use FTextureRHIRef (CameraModelUtil).
- JSON field-name literals must be TCHAR now that FStringView is UTF-16 (GeometryImporter).
- FString::Printf validates format strings at compile time; the OnScreenAssert helper forwards its format via FCString::Snprintf instead (ActorBlueprintFunctionLibrary). GEngine requires Engine/Engine.h under 5.8 IWYU.
- CityMapMeshTag.h never included its .generated.h, so ECityMapMeshTag was unreflected and 5.8's deleted StaticEnum primary rejected it.
- Extend the WideAngleLens camera_model lookup with the brown_conrady and lut models introduced by the ray-traced lens camera.
- Silence nodiscard ContainsByPredicate misuse (CarlaSettingsDelegate).

Placeholder Content/Carla git repo satisfies the configure-time content gate; real content clone deferred to the editor/package stage.

#### Description

Fixes #  

  * **Platform(s):** ...
  * **Python version(s):** ...

#### Possible Drawbacks

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9826)


JArmandoAnaya:
Great work! This is an important upgrade for the project.

Blyron:
@JArmandoAnaya @youtalk 
Would be very nice if you could able to test this PRs

https://github.com/carla-simulator/carla/pull/9826
https://github.com/CarlaUnreal/UnrealEngine/tree/ue58-dev-carla

We know lately we have been slow with PRs, but is due to on going internal projects + preparing new tools and projects for CARLA. Sorry for that.

We would really appreciate your help with this PRs

JArmandoAnaya:
I understand, no problem. And of course, I will try to take some time this week to test these changes. I will let you know my feedback then. This is excellent work. Congrats!

youtalk:
@Blyron I'd be happy to try it out in this week too.

JArmandoAnaya:
@Blyron Tested this branch end to end. Five issues block the build or the package. After fixing them locally, everything builds, tests pass, and a traffic run is clean. Details below.

Environment: Ubuntu 24.04, engine `ue58-dev-carla` (UE 5.8.0), content repo `ue58-dev-carla`, Python 3.10, ROS 2 on, LibCarla tests on.

### 1. Toolchain picks the wrong clang sysroot

`CMake/Toolchain.cmake` globs `v*_clang-*` and takes the first match, which is alphabetical.

An engine upgraded in place keeps `v23_clang-18.1.0-rockylinux8` next to `v26_clang-20.1.8-rockylinux8`. v23 wins the glob, and only v26 ships `libc++.a`. With the new sysroot fallback the configure fails:

```
Check for working CXX compiler: .../v26_clang-20.1.8/bin/clang++ - broken
clang++: error: no such file or directory: '.../v23_clang-18.1.0/.../lib64/libc++.a'
```

`-DUE_SYSROOT=` does not work around this. `try_compile` re-runs the toolchain file in a fresh sub-cache without the override, so the glob re-picks v23 there.

Fix: sort candidates descending, or require `libc++.a` in the candidate before accepting it.

### 2. CarlaEngine.cpp missing Engine/Engine.h

`Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Game/CarlaEngine.cpp`, lines 353 to 366, 7 errors:

```
error: use of undeclared identifier 'GEngine'
```

Same IWYU issue the PR fixes in `ActorBlueprintFunctionLibrary`. The usage comes from the `-RenderOffScreen` commit, not the port commit. Errors point at `Module.Carla.3.cpp`, so it compiles or not depending on unity build grouping.

Fix: add `Engine/Engine.h` and `Engine/GameViewportClient.h`, the latter because the code dereferences the viewport client.

### 3. StreetMapComponent.cpp still passes bool for EAllowShrinking

`Plugins/StreetMap/Source/StreetMapRuntime/Private/StreetMapComponent.cpp`, lines 341, 359, 374, 712:

```
error: cannot initialize a parameter of type 'EAllowShrinking' with an rvalue of type 'bool'
```

The PR description lists `StreetMapComponent` with `MeshToSplineActor`, but the diff only changes `MeshToSplineActor`.

Not caught by the game target: `StreetMapRuntime` is `"Type": "Editor"` in the uplugin, so it only compiles in the editor target that packaging needs.

Fix: `false` to `EAllowShrinking::No` at the four sites.

### 4. DefaultScalability.ini uses CVars deprecated in 5.7

`r.TranslucencyLightingVolumeDim` and `r.TranslucencyVolumeBlur`, 8 sites across `[EffectsQuality@0..3]`.

Both are `FAutoConsoleVariableDeprecated` in 5.8 (`TranslucentLighting.cpp:3365-3366`). Reading them raises a handled ensure in `Scalability::InitScalabilitySystem()`. Each ensure emits about 25 error lines, which the cook counts:

```
Failure - 50 error(s), 10 warning(s)
AutomationTool exiting with ExitCode=25 (Error_UnknownCookFailure)
```

Fix: rename to `r.TranslucencyLightingVolume.Dim` and `r.TranslucencyLightingVolume.Blur`. I checked the other 35 deprecated CVar renames in 5.8 against the config files, these two are the only hits.

### 5. Packaging stages an empty package and exits 0

The UAT command from `Unreal/CMakeLists.txt` has no `-pak` and no `-iostore`:

```
BuildCookRun ... -cook -stage -archive -package -iterate -clientconfig=Development ...
```

In 5.5 the cook wrote loose files and staging copied them. In 5.8 the cook writes to the Zen store, so staging finds nothing loose and ships binaries only.

Result: 2.3 GB package, no `Content` directory, no maps, no paks. The cook reports `Packages Cooked: 5371`, the target exits 0, no warning or error anywhere.

Fix: add `-pak -iostore`. Package goes to 15 GB with `pakchunk0-Linux.pak`, `pakchunk0-Linux.ucas`, `global.utoc`, and the server boots.

Related: the archive step copies without deleting, so a `ue.projectstore` from an earlier Zen stage survives into a later `-pak` package and the binary refuses to start with `Network data streaming failed to connect`. Deleting that file fixes it. A clean `-pak` stage never writes it.

### Results after the five fixes

- `carla-client`, `carla-server`, `carla-python-api-install`: 0 errors
- `carla-unreal`: 0 errors, from a full wipe of `Build/`, `Intermediate/` and `Binaries/`
- LibCarla tests: 149/149 server, 202/202 client
- `package-development`: 15 GB, 11 maps
- Traffic run, `Town10HD_Opt`, 10 minutes, `generate_traffic.py -n 40 -w 0`:
  - 34 vehicles on autopilot, count stable for the whole run
  - 18 to 26 moving per sample, rest held at junctions
  - mean 11 to 15 km/h, max 21 km/h
  - about 147 FPS offscreen, frames 300654 to 341277, no tick stalls
  - 0 fatal errors, assertions or segfaults in 2056 log lines
  - clean actor teardown at shutdown

### Two issues outside this PR

Client file cache accepts a path that escapes its own directory. The server sends:

```
LogCarla: Warning: Additional file not found: ../../../CarlaUnreal/Content/Carla/Config/Town10HD_Opt/Vehicles.
```

`FileTransfer::WriteFile` (`LibCarla/source/carla/client/FileTransfer.cpp:50-52`) joins that onto the cache root with no sanitisation. The `../../../` walks three levels up from `~/carlaCache/<version>/`, landing outside the user home, so every `get_world()` throws `Permission denied` on `create_directories`. Two problems: the server emitting a binary relative path, and `WriteFile` allowing `../` at all, which lets a server write outside the client cache. Reproduces on `ue5-dev`, so not caused by this PR, but it blocks any client against a packaged 5.8 server. Also appears to cause `Nav: failed loading binary`, so I ran vehicles only. Can open a separate issue.

Nanite material warnings, about 280 per run:

```
Invalid material [MID_MI_ChevroletImpala_Emissive_...] used on Nanite static mesh [SM_ChevroletImpala_Emissive].
Only opaque or masked blend modes are currently supported, [BLEND_Translucent] blend mode was specified.
```

Vehicle emissive meshes, from the Nanite rollout in the content repo. Content side fix.

Can send the five fixes as a PR against `ue58-dev` if useful.

Unreal/CarlaUnreal/Plugins/Carla/Source/Carla/Game/CarlaEngine.cpp
```diff
@@ -29,6 +29,7 @@
 #include <util/ue-header-guard-begin.h>
 #include "Misc/App.h"
 #include "PhysicsEngine/PhysicsSettings.h"
+#include "SceneInterface.h"
```
youtalk, line 38: Please also include the headers for `GEngine` and the viewport client. Without them this file only compiles when the unity blob happens to pull them in; on a clean UE 5.8 build it fails with 7 x `use of undeclared identifier 'GEngine'` from `Module.Carla.3.cpp`, for the uses added at L353, L365 and L423.

```suggestion
#include "SceneInterface.h"
#include "Engine/Engine.h"
#include "Engine/GameViewportClient.h"
```

youtalk (commented):
I reproduced @JArmandoAnaya's five issues on a clean-room build and all five confirm. Setup: fresh `ubicloud-standard-60`, Ubuntu 24.04, engine `ue58-dev-carla` (UE 5.8.0), content `ue58-dev-carla`, clang 20.1.8, PR head `9ccf1891f`, cook limited to `Town10HD_Opt`. For each issue I ran the step unpatched, captured the failure, applied one fix, and re-ran it as a hard gate.

| # | Unpatched | After fix |
|---|---|---|
| 1 Toolchain sysroot | picks `v23_clang-18.1.0` while `v26_clang-20.1.8` is present | picks `v26_clang-20.1.8` |
| 2 `CarlaEngine.cpp` | 7 x `use of undeclared identifier 'GEngine'` | `carla-unreal` succeeds |
| 3 `StreetMapComponent.cpp` | 4 x `cannot initialize a parameter of type 'EAllowShrinking'` | `carla-unreal-editor` succeeds |
| 4 `DefaultScalability.ini` | `AutomationTool exiting with ExitCode=25 (Error_UnknownCookFailure)` | 0 ensures, `ExitCode=0` |
| 5 `BuildCookRun` | no cooked content staged | `pakchunk0-Linux.pak` + `.ucas` (5.78GB) + `global.utoc` |

Runtime after all five fixes: packaged build on an RTX 5090, `generate_traffic.py -n 40 -w 0`, 10 minutes. 32 vehicles stable, 14-30 moving, max 21.0 km/h, 112 FPS offscreen, 0 fatals or assertions in 2142 log lines. This matches the original report.

I left inline notes on issues 1 and 2 and on `CarlaSetup.sh`. The remaining points are on files this PR does not touch, which is exactly why they are still broken:

**Issue 3 cannot be fixed here.** `StreetMapComponent.cpp` lives in `carla-simulator/StreetMap`, pinned by `CARLA_STREETMAP_TAG` = `ue5-dev-carla`, and that repo has no 5.8 branch. `EAllowShrinking` already exists in UE 5.5 with both `SetNum` overloads, so `EAllowShrinking::No` compiles on 5.5 and 5.8 alike. The fix can go straight onto the existing `ue5-dev-carla` branch, with no new branch and no version bump here.

**Issues 4 and 5** need `Unreal/CarlaUnreal/Config/DefaultScalability.ini` (rename to `r.TranslucencyLightingVolume.Dim` / `.Blur`, 8 sites) and `Unreal/CMakeLists.txt` (add `-pak -iostore` to `BuildCookRun`). One note on ordering: issue 5 is only reachable once issue 4 is fixed, since the cook aborts before staging otherwise. Its fix is verified.

**The client cache path escape is pre-existing but blocks every client against a packaged 5.8 server**, so I do not think it can be deferred. `get_world()` throws `create_directories: Permission denied ["<home>/carlaCache/0.10.0/../../../CarlaUnreal"]`. I confirmed the write side too: with the cache root placed deep enough that the escape lands somewhere writable, `FileTransfer::WriteFile` returns `true` and writes two levels above the cache root. A fix must normalise and contain rather than reject separators, since paths like `Config/Town10HD_Opt/Vehicles.xml` are legitimate.

Logs, patches and the workflow that produced all of this: https://github.com/youtalk/carla/actions/runs/31641880084

CMake/Toolchain.cmake
```diff
@@ -66,15 +66,29 @@ set (
...
-
-set (
-	UE_LIBS
-	${UE_THIRD_PARTY}/Unix/LibCxx/lib/Unix/${TARGET_TRIPLE} CACHE PATH ""
-)
+# UE 5.8 no longer bundles libc++ under ThirdParty/Unix/LibCxx; the headers and
+# static libs ship inside the clang sysroot instead. Prefer the legacy LibCxx
+# layout when it actually contains the libraries (UE <= 5.5), otherwise fall
+# back to the sysroot.
+if (EXISTS ${UE_THIRD_PARTY}/Unix/LibCxx/lib/Unix/${TARGET_TRIPLE}/libc++.a)
```
youtalk, line 91: This fallback is right, but it only holds if `UE_SYSROOT` already points at the new SDK, and the selection above (L38-L52) takes the first `file (GLOB)` match, which is alphabetical. An engine upgraded in place keeps `v23_clang-18.1.0` next to `v26_clang-20.1.8`; v23 wins, only v26 ships `lib64/libc++.a`, and configure then fails with `no such file or directory: '.../v23.../lib64/libc++.a'`. `-DUE_SYSROOT=` does not work around it because `try_compile` re-runs this file in a fresh sub-cache.

One caution on the fix: requiring `libc++.a` on its own would break UE <= 5.5, whose sysroots ship none (libc++ is under `ThirdParty/Unix/LibCxx`). What I verified does both passes - walk newest-first preferring a candidate that has `libc++.a`, then fall back to a plain newest-first directory pass. Tested both layouts: two SDKs with only v26 carrying `libc++.a` picks v26, and a lone v23 without it still picks v23.

CarlaSetup.sh
```diff
@@ -93,6 +93,16 @@ else
...
+# Fetches github.com/NVIDIA/DLSS into ~/SDKs/DLSS (unless DLSS_SDK already
+# points at a valid SDK) so the engine's DLSS-RR denoiser / DLSS upscaler
+# plugin builds with NGX support. Failure is non-fatal: without the SDK the
+# plugin compiles to a no-op and CARLA falls back to NFOR / spatial upscale.
+bash Util/SetupUtils/SetupDLSS.sh
+if [ -z "$DLSS_SDK" ] && [ -f "$HOME/SDKs/DLSS/lib/Linux_x86_64/libnvsdk_ngx.a" ]; then
+    export DLSS_SDK=$HOME/SDKs/DLSS
+fi
+
 # -- DOWNLOAD + BUILD UNREAL ENGINE --
```
youtalk, line 107: The engine clone just below still uses `-b ue5-dev-carla`, and the content clone above still uses `-b ue5-dev`. A fresh `./CarlaSetup.sh` on this branch therefore downloads UE 5.5 and cannot build the port at all. Both should become `ue58-dev-carla`.

glopezdiest:
I've also done some testing. For now, a couple of issues I've found:

1) As far as I saw, DLSS is optional when building the engine, which shouldn't be the case
2) The RT lens cameras have some artifacts when the traffic lights change colors

Image

Image

Aperiss:
# `VK_ERROR_DEVICE_LOST` in the UE 5.8 editor

The editor built from this branch loses the Vulkan device and terminates.

**Environment:** RTX 3090, driver **595.84**, Ubuntu 24.04, engine branch `ue58-dev-carla`.

## The error

```console
LogVulkanRHI: Error: Result failed, VkResult=-4
 at ./Runtime/VulkanRHI/Private/VulkanSynchronization.cpp:136
 with error VK_ERROR_DEVICE_LOST

LogVulkanRHI: Error: Shader diagnostic messages and asserts:
        Device: 0, Queue Graphics:      No shader diagnostics found for this queue.
        Device: 0, Queue AsyncCompute:  No shader diagnostics found for this queue.

LogVulkanRHI: Error:
DEVICE FAULT REPORT:
* Description:
* Address Info:
* Vendor Info:
* Vendor Binary Size: 0

LogCore: FUnixPlatformMisc::RequestExit(1, FVulkanDynamicRHI.TerminateOnGPUCrash)
```

`VulkanSynchronization.cpp:136` is the generic error arm of `FVulkanFenceManager::WaitForFence`, where every fence wait in the RHI funnels:

```cpp
VkResult Result = VulkanRHI::vkWaitForFences(Device.GetHandle(), 1, &Fence->Handle, true, TimeInNanoseconds);
switch (Result) {
case VK_SUCCESS:  ...
case VK_TIMEOUT:  break;
default:          VERIFYVULKANRESULT(Result);   // line 136
}
```

So the CPU was blocking on already-submitted work and `vkWaitForFences` returned `VK_ERROR_DEVICE_LOST`. GPU faults are asynchronous and surface at the next synchronisation point. The engine then tears the process down via `TerminateOnGPUCrash`.

Three things about the report matter:

- **`VK_EXT_device_fault` returns an empty structure** every single time — no description, no address, no vendor info, `Vendor Binary Size: 0`.
- **No shader diagnostics** on either queue, which rules out a shader page fault (those populate the address block).
- **The kernel logs zero Xid**, on every boot (`journalctl -k | grep -ic xid` → `0`). A genuine GPU page fault or channel timeout would normally log one.

The GPU breadcrumbs put the stop at the tail of the frame: the whole scene reports `Finished` and the GPU is left in the present / Slate path.

```console
LogRHI: Error: Active GPU breadcrumbs:

	Device 0, Pipeline Graphics: (In: 0x80291ca2, Out: 0x80291ca1)
		(ID: 0x80291b6d) [     Active]	Frame 8785
		(ID: 0x80291b70) [   Finished]		SceneRender - ViewFamilies    (…entire scene Finished…)
		(ID: 0x80291c9f) [   Finished]		RenderGraphExecute - Slate
		(ID: 0x80291c9d) [   Finished]			SlateUI Title = CarlaUnreal - Unreal Editor
		(ID: 0x80291ca1) [     Active]		EndDrawingViewport
		(ID: 0x80291ca2) [     Active]		EndDrawingViewport
```

## Conditions

**It only happens in the editor.** The same build launched with `-game` has never lost the device, headless (`-RenderOffScreen`) has never lost it either. Both of those skip the editor viewport / Slate presentation path, and headless has no swapchain at all.

Within the editor the trigger conditions are broad and non-deterministic:

- **No simulation is required.** It happens with no play session, no RPC server and no sensors — the editor simply drawing its own viewport is enough.
- **Time to failure varies widely** between runs of the same configuration, with no correlation to frame rate, frame count, map or scene content.
- **No path tracing needed.** Both DLSS-active losses happened with no path-traced camera in existence, so the DLSS denoiser itself never ran; the plugin's only effect on those sessions was at device creation.
- **No loss with Vulkan validation layers enabled** (`-vulkanvalidation=2`), including a long idle run in the configuration that died quickly.

gsmario:
This is what I can see in Windows using the RT lens (pinhole). 
image

germanros1987:
@JArmandoAnaya @youtalk @glopezdiest thank you for the thorough testing — every issue reproduced and all of them are now addressed. Here's the rundown:

**On `ue58-dev` (this PR):**

1. **Toolchain sysroot selection** — fixed in `b209e0b18` following @youtalk's two-pass approach: walk candidates newest-first preferring one that ships `lib64/libc++.a`, fall back to plain newest-first so UE ≤ 5.5 layouts (libc++ under `ThirdParty/Unix/LibCxx`) keep working. Verified with a decoy `v23_clang-18.1.0` directory planted next to `v26`: configure picks v26 and survives `try_compile`'s fresh sub-cache.
2. **`CarlaEngine.cpp` missing includes** — fixed in `c2c339c84`, applying @youtalk's suggestion (`Engine/Engine.h` + `Engine/GameViewportClient.h`).
3. **StreetMap `EAllowShrinking`** — landed directly on `carla-simulator/StreetMap@ue5-dev-carla` (`71497b9`) as @youtalk suggested; no tag bump needed. Verified: `CarlaUnrealEditor` compiles `StreetMapRuntime` cleanly.
4. **Deprecated scalability CVars** — fixed in `56e97d667` (8 sites → `r.TranslucencyLightingVolume.Dim`/`.Blur`). Cross-checked all 36 `FAutoConsoleVariableDeprecated` old names in the 5.8 source against our config files: no other hits.
5. **Empty package** — fixed in `e41d739b1`: `-pak -iostore` added to `BuildCookRun`. The stale `ue.projectstore` is purged in the archive-cleanup step — which it turns out had never run at all: the script read `CARLA_TARGET_PACKAGE_PATH` while the caller passes `CARLA_PACKAGE_ARCHIVE_PATH`, so the manifest cleanup was silently dead too. Both fixed.
6. **Client cache path escape** — agreed this couldn't be deferred; fixed on both sides of the wire in `686dadf35`:
   - Server: `get_required_files` now normalizes each found file to an absolute path before stripping the content dir, so packaged builds (where the file manager returns pak-relative `../../../CarlaUnreal/Content/...` paths) send content-relative names again.
   - Client: `FileTransfer` normalizes and *contains* every server-supplied path inside the cache root (dropping root/drive/leading-`..` components) — legitimate nested paths like `Config/Town10HD_Opt/Vehicles.xml` resolve unchanged, and escaping paths are remapped rather than rejected so the client also behaves sanely against unfixed servers.
   - Bonus: `request_file` on the server appended a client-controlled name to the content dir unchecked (arbitrary file **read** on the server host via `..`); it now refuses names that resolve outside the content folder.
7. **Setup script branches** (@youtalk's inline note) — fixed in `c3ba2ac70`: both `CarlaSetup.sh` and `CarlaSetup.bat` now clone `ue58-dev-carla` for engine *and* content (the content branch of that name exists on Bitbucket).

**On `CarlaUnreal/UnrealEngine@ue58-dev-carla`:**

8. **DLSS must not be optional** (@glopezdiest) — `0871d2df7160`: the licensing constraint stands (the NVIDIA SDK cannot be vendored in an MIT repo), so instead of silently compiling the DLSSRRDenoiser plugin as a no-op, a missing/invalid `DLSS_SDK` is now a hard build error with fetch instructions. Building without DLSS is an explicit opt-in via `DLSS_SDK=disabled`. `CarlaSetup.sh` fetches the SDK automatically and now also fails loudly instead of warning.
9. **RT lens artifacts on traffic light changes** (@glopezdiest) — root-caused and fixed in `0d1491d51de6`. The lamps toggle via dynamic-material parameter changes; those restart path-tracer accumulation, but our DLSS-RR integration deliberately preserves the denoiser's temporal history across restarts (so camera motion doesn't degrade it to per-frame spatial denoising). A material change has no motion signal, so the stale history ghosts the old lamp into the restarted accumulation — and on a static camera the ghost then *freezes into the output permanently*, because the converged path tracer stops re-rendering until the next invalidation. The fix tracks material-parameter invalidations in their own counter and clears the denoiser history for exactly those (like a camera cut), while transform-driven invalidations (moving traffic, every frame) keep the history. Verified with a static rt_lens camera on a frozen light switched red→green: before, the old lamp froze at a lit ghost indefinitely; after, it goes dark on the first frame past the switch.

**Verification on this end** (RTX workstation, engine `ue58-dev-carla` at `0d1491d51de6`, Town10HD_Opt-only cook): `carla-server`/`carla-client`/wheel build clean; `CarlaUnreal` and `CarlaUnrealEditor` targets succeed; `package-development` produces an 8.1 GB package with `pakchunk0-Linux.pak/.ucas/.utoc` + `global.utoc` and no `ue.projectstore`; the packaged server boots from paks (map load 8.6 s) and a client with a fresh cache runs `get_world()` successfully, with `Carla/Maps/OpenDrive/Town10HD_Opt.xodr`, `Carla/Maps/Nav/Town10HD_Opt.bin` and `Carla/Maps/TM/Town10HD_Opt.bin` all landing inside the cache root (the nav-loading failure is gone as well). An rt_lens camera on an autopilot vehicle among 20 TM vehicles stays temporally stable with the denoiser-history change.

Also addressed in the second pass:

- **Nanite translucent-material warnings** — fixed on the content side in two passes on `carla-content@ue58-dev-carla`:
  - `738013e08`: the 308 vehicle/street glass, emissive, blinker, and siren meshes (some translucent only at runtime via blueprint-assigned MIDs, which is why the warning survives a static material scan). Verified by spawning every vehicle blueprint on Town10HD_Opt: warning count 260 → 0.
  - `8730f69f0`: the remaining 4,853 meshes on the large maps — what the audit first labeled "Terrain tiles" turned out to be 4,716 tiny generated road-sign meshes (median 4 triangles, and orphaned: the asset registry shows zero referencers) plus 137 building-glass/streetlight/bus-stop meshes that do stream in. These weren't cosmetic: Nanite draws invalid sections with the default material, and the all-glass towers weren't rendered at all — entire buildings were missing from Town13's downtown. Verified: editor audit now reports zero Nanite meshes with non-opaque materials under `/Game/Carla`, full-map streaming sweeps of Town12/Town13/Town15 log zero `Invalid material` warnings (Town13 before: 67), and A/B downtown screenshots show the previously missing glass geometry rendering again.
- **Town12 crashed the server on `load_world`** (found while validating the above): Town12 is World Partition in the 5.8 content, but `ALargeMapManager` dereferenced the null `UWorldComposition` in `BeginPlay` and segfaulted 100% of the time. Initially stopped with a guard (`77080cbb3`); now properly resolved by the branch integration below, which brings the full World Partition migration.

**`ue58-dev` is now the integration branch for the whole UE 5.8 migration.** The follow-up migration branches (`ue58-dev-carla`, `ue58-mapgen-features`, `ue58-migration-02/-03` — 91 commits) have been merged in, together with their paired engine and content branch updates. Highlights:

- **Large maps on World Partition**: legacy tiled towns migrated to native WP streaming with double-precision Large World Coordinates (`CarlaLargeMapConvert` commandlet converts a tiled town in place); hero vehicles get streaming sources so the ground under them always loads; WP streaming crashes fixed in traffic lights, the actor registry, the LiDAR, and semantic tagging of streamed cells.
- **Weather & night lighting**: the weather API reconnected to the UE5 sky rig — this also fixes the converted large towns rendering near-black at noon — plus a full night-lighting rework (street lamps emit real light under 5.8, ambient moon/skylight floor, vehicle headlight beams, night sky stars) and the street light API in `manual_control`.
- **Traffic Manager & vehicle control**: linear Chaos steering with a reimplemented `GetWheelSteerAngle`, retuned lateral PID, stuck-vehicle K-turn recovery, parked-vehicle gating, speed seeding, route-end hold, and a deterministic TM A/B benchmark.
- **Generated OpenDRIVE maps**: terrain heightfield following road elevation, crosswalk zebras, lane markings through splice junctions, `lateralProfile <shape>` support, median gap filling, PCG street furniture with working night lighting, and a `spawn_custom_mesh` RPC.
- **Rendering**: Lumen on hardware ray tracing with ray-traced shadows, DLSS-RR denoising re-enabled for the path tracer, rt_lens auto-exposure pinned to the daylight window, Vulkan bindless descriptor heap grown to 2M entries, Nanite cluster streaming pool tripled for the large towns.
- **NuRec**: the NVIDIA NuRec example suite ported to CARLA UE5 / NRE 26.04.
- One new fix found while validating the integration (`9dc1e3756`): a `get_lights` RPC racing a World Partition stream-out (or a map transition) segfaulted the server — `UCarlaLight::GetLocation` now guards owner/game-mode, `GetLights` skips dead lights, and the light registry is GC-visible.
- **Daytime shadows rendered 100% black on every map** (found driving Town12: whole shaded streets and facades at zero — impossible under any GI). Three stacked causes, each verified in isolation: (1) BP_Carla_Sky ships every component with `bAutoActivate` false and only the night branch of the weather push re-activated the skylight, so by day the scene had no ambient source at all against the physical 100k lux sun; (2) the skylight's authored source is a null cubemap — black at any intensity — and the `bRealTimeCapture` override that would render the rig's SkyAtmosphere into it had nothing to see (the visible sky is the legacy non-IsSky sphere mesh, and nothing fed the capture); (3) the engine only lets the **primary viewport** update the sky env map, and on a `-RenderOffScreen` server the primary viewport renders exactly once at startup — so the map stayed black forever regardless. Fixed on both sides: `carla@c15954d99` forces the skylight active + real-time capturing and activates the rig's SkyAtmosphere (main-view sky look unchanged — the legacy sphere still covers it; ambient now lands in the same photometric units as the sun and tracks sun altitude/weather for free), and `UnrealEngine@7362923d3a94` lets a scene-capture sensor drive the sky env map capture as a fallback, gated to only kick in when no primary view has rendered within 8 frames (behavior with a live viewport unchanged; the per-frame GPU guard keeps it to one update per frame). Validated on Town12: the five worst poses drop from 30–73% pure-black pixels to 0–7%, shaded facades and streets fully readable; night look (moon floor + stars + street lamps) unchanged.
- **Traffic Manager destabilized by low server frame rates** (found testing rt_lens on Town12): the TM PID was discretized at the sync-mode design period (0.05 s) with fixed per-tick slew and derivative clamps, so any render load stretching async frames (a path-traced sensor, or just two 1440p RGB cameras) doubled the loop gain/delay — vehicles diverged into weaving, then the stuck-recovery cycled alternating full-lock ("jerking left and right"). Fixed in `17068ab73`: the controller now measures the real simulation-time period, discretizes I/D with it, scales the lateral gains down by min(1, DT/dt), and converts the steering slew/derivative clamps to per-second rates. Sync mode at 0.05 s is bit-for-bit unchanged. Validated: at 11.9 fps (previously full-lock oscillation within seconds) the hero holds lane with steer_std 0.0046, max |steer| 0.016.
- **Signs floating in the path-traced view** (same session): 5.8's dynamic Nanite ray tracing builds BLASes from a cluster cut that refines only as pages stream in, and a coarse cut destroys thin geometry first — driving past signs, poles rendered as bent stubs or vanished while the flat plates survived, leaving regulatory signs hovering mid-air (raster always draws the fine cut, so `sensor.camera.rgb` looked correct). Fixed on the content side (`carla-content@10b267098`): Nanite disabled on the 180 sign-plate/pole meshes (36–1,526 triangles — far below where Nanite pays for itself), giving them exact-geometry static BLASes from the first frame; same reasoning as the existing `r.Nanite.AllowSkinnedMeshes=0`. Verified with a KB-fisheye rt_lens drive-past on Town12: all sign poles grounded in every frame.
- **Server crash after ~30 min of continuous path-traced captures** (Vulkan bindless descriptor heap exhaustion — the same signature that motivated the earlier 768k→2M heap bump, which turned out to treat a symptom): the ray tracing SBT pins per-frame uniform buffers (ray-traced decal parameters, PT cloud/preshading) and the only release site was the primary-viewport cleanup path — which a `-RenderOffScreen` server never runs after startup. Every path-traced sensor frame leaked one set (measured: 1.09M live `FDecalParametersRayTracing` buffers in under 10 min on Town12's decal-heavy roads, plus matching host-memory growth over hours). Fixed in `UnrealEngine@83970eeef6db`: both scene-capture paths now enqueue the same SBT transient-buffer release the viewport performs (redundant but harmless when a viewport exists). Companion diagnostics in `a6d6a71ce20a`: per-type bindless descriptor pressure telemetry, a proper Fatal-with-breakdown at the exhaustion site instead of a delayed cryptic type-check crash, and a per-layout histogram that names the leaking uniform buffers. Validated with a teleport-churn soak under a persistent PT sensor: decal buffers now oscillate at one frame's worth (13–324) instead of monotone growth.
- **Client segfault when switching sensors** (found cycling cameras in `manual_control`): `streaming::detail::tcp::Client::Stop()` closed the socket and flipped `_done` from the game thread while all completion handlers run on the asio strand — a data race on a non-thread-safe socket — and the data handler delivered its payload without re-checking `_done`, calling into a subscriber already being torn down (use-after-free, crash symbol `tcp::Client::ReadData`). Fixed in `6f1f800e6`: `Stop()` posts the shutdown onto the strand (kept alive via `shared_from_this`), and the read handler bails out once `_done` is set like every other handler. Validated with a 300-cycle sensor-switch stress (5 camera types incl. rt_lens, autopilot + 4 auxiliary sensor streams + snapshot stream, varied dwell): PASS, zero client segfaults.
- **Bonus from the same engine fix: the path tracer's black sky in `-RenderOffScreen` is gone** — the PT samples the same never-captured sky env map. Verified: an rt_lens capture on the fixed server renders a bright blue sky (top-third mean ~190/255, previously 0).

- **Pedestrians on large maps** (`carla@bae19c1a2` + `carla-content@2b49fc775`): large maps had no walker support at all (the legacy pipeline needs the whole town as a single FBX through the external RecastBuilder, and the client-side Detour crowd is streaming-blind). Phase 1 of the replacement: the engine's own world-partitioned navmesh (built offline per WP cell, double-precision, streams with the map) plus a server-side crowd — `AWalkerController` is now an `AAIController` on a `UCrowdFollowingComponent`, with six new RPCs behind the *unchanged* Python API (`get_random_location_from_navigation`, `controller.ai.walker` start/go_to/speed/stop). On maps without an engine navmesh the legacy `.bin` path runs untouched (regression-tested). Validated on Town12: walkers spawn at random nav locations, path-follow across streaming cells (median 56 m walks), stop on command, zero falls — and keep walking with no client ticking, which the legacy crowd cannot do. Phase 2 (`carla@9cddf35c7` + content follow-up): road/crosswalk nav areas (roads 10x cost via per-mesh NavCollision AreaClass, crosswalk NavModifier boxes generated from the OpenDRIVE by a new CarlaTools commandlet), `set_pedestrians_cross_factor`/`set_pedestrians_seed` served server-side with legacy semantics (per-walker crosser draw, deterministic under a seed) — validated A/B on Town12: factor 0.0 → 0/20 walkers ever touch a driving lane, 1.0 → crossing enabled. This surfaced a content bug worth flagging: the mapgen exporter wrote every crosswalk as `type="-1"` in the .xodr, so `carla.Map.get_crosswalks()` returned nothing on Town12/13/15 — fixed in the same content push. Remaining: vehicle avoidance and virtual-walker LOD.

- **Walkers took 0.5–6 s to start moving** (worse the more loaded the server — initially misread as a spawn-penetration issue): `AWalkerController::Tick` encoded the commanded speed as the movement-input *magnitude* (`Speed/4096`), and `UCharacterMovementComponent::ScaleInputAcceleration` scales acceleration by that same magnitude — so a 1.5 m/s walker accelerated at ~4% of its Blueprint `MaxAcceleration`. Fixed in `5362fa5ef`: manual control drives `MaxWalkSpeed` from the commanded speed and feeds a direction-only input at full magnitude (navigation mode resets `MaxAcceleration` so manual state can't leak in). Causality-proven with per-frame instrumentation, then validated: time-to-first-movement after spawn goes from median 0.65 s (multi-second under load) to a deterministic 0.22 s, with the walk speed still capped at exactly the commanded value.
- **Vehicles launched and spun on pedestrian contact** (`de34a80ac`): at 45 km/h a pedestrian hit yawed the car 20–49° off heading at up to 84°/s — and the walker survived. Root cause was not the contact impulse: the walker capsule is QueryOnly, so there *is* no physical contact — the Chaos wheeled-vehicle suspension raycasts (queries) rode the blocking capsule like a rigid, invisible bollard. The ue4-dev "walker dies" behavior had also been silently lost in the UE5 content migration (BP_Walker's vehicle death-trigger event binding disappeared). Both are now handled natively in `AWalkerBase`: a vehicle contact above ~7.6 km/h kills the walker — navigation stops, further controls get the existing `WalkerDead` rejection, capsule and mesh stop registering on the Vehicle channel (lidar/cameras/crowd still see the body), and the walker either plays the ue4-style directional death animation (`deterministic_ragdolls`, recorder-replayable) or physics-ragdolls with the impact velocity, auto-destroying after 10 s. The kill also re-broadcasts the hit on the vehicle so `sensor.other.collision` (scenario_runner/leaderboard infraction scoring) keeps reporting the pedestrian collision. Parking-speed brushes and walkers bumping into stationary vehicles never kill. Validated 5/5 in both death modes at 45 km/h: max roll 1.5°, pitch 2.6°, no airborne phase, heading deviation < 1° (baseline: roll 8.5°, yaw rate 68–84°/s); 5/5 low-speed brushes leave the walker alive; Town10/Town12 walker acceptance unaffected. A follow-up (`d0b1d98b5`, found staging a live 107 km/h hit): the proximity poll fires at most once per frame, and a fast vehicle crosses the entire contact shell between polls — the kill detection now also tests each candidate vehicle's predicted path over the next poll interval (direction-aware, with an adjacent-lane guard), validated 5/5 head-on at ~93 km/h and 5/5 non-killing adjacent-lane passes at the same speed.
- **Pedestrians never visibly died — and could still launch a car** (found drive-testing the above; `b2d94c3d8` + `carla-content@3042817dc1` + `UnrealEngine@3f57cff9a439`): three more root causes surfaced under live manual driving. (1) The four `Anim_Death_*` sequences "played" (position advanced) but the walker stayed standing: `force_root_lock` pinned the root bone, and the entire fall lives in the root track — editor pose evaluation ignores the lock, which is why the assets looked healthy under inspection; fixed content-side (root lock off, recompressed off the ue4-era ACL codec). (2) Sub-threshold contacts interpenetrated: any moving-vehicle contact now kills (50 cm/s threshold, ue4 parity); only standing vehicles are non-lethal. (3) The launch/spin had *three* stacked physics paths, hunted down with per-trace engine instrumentation: suspension raycasts riding the blocking capsule (engine: wheel traces now ignore Pawn objects), migrated walker BPs with physics-enabled primitives acting as infinite-mass walls to the chassis (clamped to query-only at spawn), and — the sneakiest — walker body meshes migrated with a `WorldDynamic` object type, so a sphere-cast wheel climbed the pedestrian's mesh like an invisible ramp (normalized to Pawn at spawn). End state: plowing through 14 mixed-model pedestrians at 50 km/h leaves the car with 1.5° roll, no lift, 0.9° total heading drift (baseline: 163° roll flips, 705°/s spins); walkers crumple to the pavement via the directional death animation (or physics ragdoll with `deterministic_ragdolls=False`), corpses auto-clear after 10 s, and `sensor.other.collision` still reports every pedestrian hit for infraction scoring.
- **Traffic signs floated over missing ground, then vanished entirely** (user-reported on Town12; `170c7dccc` + `7864e8853` + `d75d3b0f7` + a Town12 content cleanup): four stacked defects in how converted World Partition maps host the ~3,700 runtime OpenDRIVE-spawned signals. (1) *Placement is correct but unmanaged*: signals are persistent while terrain streams per cell, so signs rendered over not-yet-resident ground — `ATrafficLightManager` now sweeps them (WP maps only, 150/tick) and hides any signal with no ground beneath (short downward trace), mirroring the legacy `ALargeMapManager` dormancy the conversion lost. (2) *The conversion left 7,523 Town12 external actors with dangling `ColorVariation` MID references* to the deleted legacy tile worlds — a `CreateExport` error storm on every cell load (15k+ per session) that hammered and occasionally wedged the async loader; all 7,523 resaved with the stale imports stripped (verified: zero load errors across a full Town12 streaming session, was 15,305). (3) *The wipe*: signals spawned via `OverrideLevel=nullptr` fall back to their OWNER's level, and the owner — the baked `TrafficLightManager`, carried over by the conversion as a spatially-loaded cell actor — put all ~3,700 signals inside one streamed cell, so they were silently torn down with the level (no `AActor::Destroy`, no registry callbacks) minutes into a session, registry 3,670 → 0. `GetULevelFromName` now falls back to the persistent level, and a baked in-cell manager is migrated to a persistent-level replacement (model configuration adopted, late-streaming duplicate culled). (4) *Even over resident ground, stop and yield signs hovered ~2.3 m in the air* (609/609 stop signs, median +2.34 m; traffic lights +0.48 m; speed limits unaffected — which singled out the mechanism): the mapgen exporter writes each signal's plate mounting height into the OpenDRIVE `zOffset`, LibCarla folds it into the signal transform per the spec (the transform names the plate reference point), and the manager spawned a *complete* signal blueprint — pole included, pivot at the pole base — at that plate-height point, raising the whole sign by exactly `zOffset`. Fixed in `d75d3b0f7` by spawning full-signal blueprints at road level; maps whose signals carry `zOffset=0` are bit-for-bit unchanged. Verified on Town12: signals hold 3,670/3,670 through 30 min of idle streaming (previously wiped at t=70 s), and sign bases sit on the ground everywhere measured (stop signs median dz +0.02 m, yield +0.00, traffic lights +0.02, n=120 across classes).
- **`CarlaLargeMapConvert` left the legacy `ALargeMapManager` in converted maps** (`25cd64649`): its saved `MapTiles` references to the deleted streaming tiles produced a failed-import load error per tile on every subsequent load (144 on Town12) and forced any commandlet run on the map into exit code 1. The converter now drops the actor — World Partition streams natively.
- **Town13 and Town15 brought fully online as World Partition maps** (content up to `carla-content@7ba5c75256`; carla `e5b5b9158` + `7e92ec707`): both towns converted with `CarlaLargeMapConvert`, crosswalk nav modifiers generated, and their dangling-MID import storms stripped like Town12's (19,157 actors on Town13, 18,751 on Town15). Getting walkers to pass acceptance surfaced three more root causes worth knowing about. (1) Town13: the navmesh builder never re-serialized the `RecastNavMesh` actor itself, so the runtime came up with a 3-tile pool ("Failed to add tile - 3 tile limit reached") and nav queries collapsed; a degenerate zero-scale nav-bounds volume was removed and the tile size bumped (1000→1500), forcing a rebuild that persists the actor. (2) Town15 exposed a genuine `CarlaCrosswalkNavBuilder` bug: it only closed a crosswalk outline when a point returned to the run's *first* corner, but Town15's OpenDRIVE outlines close onto an interior corner — all 34 outlines overflowed the corner cap and merged into 34 town-spanning walker-excluded stamps that blanketed the street navmesh (85% of `get_random_location_from_navigation` calls returned zeros, and the only sampleable polys left were rooftops Recast had rasterized from collision-less roof planes — walkers spawned in the sky and fell). Fixed in the builder (`e5b5b9158`: close on any repeated corner → 63 correct boxes) and in content (buildings/vegetation tagged nav-irrelevant; note for posterity: the engine's ISM nav export ignores `bUseSurfaceArea`, so instanced meshes need the dynamic-obstacle path). (3) A client-side LibCarla bug this uncovered: `Episode` never dropped its cached `WalkerNavigation` on `load_world` — the reset lived in `OnEpisodeStarted()`, dead code since the stream listener switched to `OnEpisodeChanged()` — so nav queries kept answering from the *previous* map's navmesh indefinitely; fixed in `7e92ec707` (reset on episode-id change, before the new id unblocks `load_world`). Walker acceptance now passes on all three large towns: Town12 median 56 m, Town13 63 m, Town15 61 m — 20/20 walkers moving, zero falls, all halting on `stop()`.

Paired branch heads for this state: engine `CarlaUnreal/UnrealEngine@ue58-dev-carla` = `3f57cff9a439`, content `carla-content@ue58-dev-carla` = `7ba5c75256` (both what `CarlaSetup.sh` clones).

**Integration validation** (RTX workstation, `ue58-dev` at `9dc1e3756`): LibCarla client suite 202/202 and server suite 149/149 (including the streaming benchmark this time); Town12 loads and runs a full-map streaming sweep with per-step RPC round-trips plus a 150 s idle soak — zero crashes, zero invalid-material warnings; Town10 regression sweep + soak equally clean.
- **CHANGELOG** updated (`8c9ad95de`) per the checklist/bot.
- **Documentation** updated (`15d979385`): full `sensor.camera.rt_lens` reference (lens models, denoising, attribute table), the new camera attributes (`use_ray_tracing`, `enable_dlss`, `dlss_screen_percentage`), actual ray-tracing/DLSS defaults in the rendering options page, TM tick-rate robustness, World Partition large maps + `CarlaLargeMapConvert`, weather/night lighting, generated-map features, and the restored `LightManager` Python API reference. A follow-up (`04973839e`) adds the NVIDIA ecosystem page and a full NuRec guide for the UE5/NRE-26.04 integration, mirroring the ue4-dev docs structure.
- **LibCarla tests**: client 202/202; server 148/149 — the one failure is `benchmark_streaming.image_1920x1080_mt`, a throughput benchmark that consistently lands at 1337/1600 messages on this workstation (threshold 1440). The only LibCarla change vs. the head you tested 149/149 on is client-side `FileTransfer.cpp`, which isn't linked into the server suite, so this is machine throughput, not a regression.

(The path-tracer black-sky note from the earlier draft is resolved — root cause was the never-captured sky env map described above, fixed by `UnrealEngine@7362923d3a94`. The near-black-at-noon rendering of the converted large towns was the weather rig disconnect and is fixed by the integrated weather work.)

JArmandoAnaya:
@germanros1987 Thank you for the status. I'm excited about this upgrade. I will run new tests once you finish the current refactoring.

germanros1987:
Two Traffic Manager fixes landed on `ue58-dev` after the summary above, from a developer report worth documenting: on Town10, TM vehicles were running wide through junction turns — the trajectory connecting into the target lane invaded adjacent lanes, as if the turn radius were too large.

**First fix — the pursuit chord cut (`9f8c3395f`).** Instrumenting the TM per tick ruled out the two obvious suspects: the Localization-stage buffer through junction connecting roads lies exactly on the OpenDRIVE lane centerlines (the plan is clean), and steering authority is ample (at junction speed the allowed steer envelope achieves a 5.9 m radius against the ~11 m needed, and vehicles peaked at 0.28 of an allowed 0.44 — no saturation). The defect is the pursuit *reference*: the controller chases a point `d = speed × 1 s` ahead along the buffer, and through an R≈11 m connecting road at ~8 m/s that ~8 m chord cuts the curve (geometric sagitta `d²/8R ≈ 0.8 m`, ~1.4 m with vehicle dynamics) — the vehicle turns in before the lane does, pokes up to 0.69 m into the adjacent lane at the apex, and exits ~0.55 m wide. An earlier commit had capped `d` at a fixed 10 m for highway-radius curves, which does nothing at junction radii. Fix: cap the pursuit distance by *local path curvature* — `d ≤ √(8·R_local·0.2 m)` with `R_local` the three-point circumradius of the buffer ahead — and scale the lateral P/D by `d_ref/d` (pure-pursuit equivalent gain ∝ 1/d, tuned at the 10 m cruise anchor) so the shorter target keeps the tuned bandwidth. At cruise the cap is inactive and the scale is exactly 1, so highway/straight behavior is bit-unchanged. Seeded A/B (Town10HD_Opt, 20 cars, sync 0.05 s, 3000 ticks): lane-invasion events 16 → 3, max excursion past the lane edge 0.69 → 0.12 m, apex radius p50 12.1 → 10.1 m against a planned ~11 m, straight-road steer_std unchanged, zero contacts in both runs.

**Second fix — vehicle-geometry-aware lateral control (`d6baf33a2`).** Live observation after the first fix: large vehicles still visibly ran wide. Per-class traces made the mechanism unambiguous: trucks (8–8.6 m — firetruck, carlacola, fusorosa) invaded on *every* junction turn, with peak error vs their own planned buffer at p90 4.29 m and **45% of junction ticks pinned at the steering envelope**, while cars and vans in the same runs had zero invasions and zero saturation, with a clean length gradient between them. Every blueprint has the same 70° wheel lock, so the deficit is purely wheelbase: normalized steer maps to curvature as `tan(steer·lock)/wheelbase`, ~1.75× less curvature per command for a fire truck than for the car the gains — and the envelope — were anchored on. The trucks physically could not follow their (correct) plan inside a car-anchored steer cap. Two alternatives were prototyped, measured, and rejected on data: a rear-axle pursuit reference halved the error but added enough loop delay to weave on tight turns (and a bounding-box-center reference is actually optimal for lane containment of a rigid body), and there was no oscillation signature from the first fix's gain scaling. Fix: compute the lateral command through each vehicle's *own* steering geometry — command the pure-pursuit curvature `κ = margin·2·sin(α)/d` and invert the vehicle's steering map `steer = atan(wheelbase·κ)/lock`, with the physical wheel lock read once per vehicle from its physics control and the steering envelope rescaled as a physical *curvature* guard rather than a normalized-command cap. The margin is chosen so the law reduces exactly to the validated linear P=2 gain for the anchor car at cruise — cars are unchanged by construction. Per-class A/B (same seed): trucks' wide error p90 4.29 → 0.22 m, envelope saturation 45% → 0%, apex radius 13.0 → 10.9 m (plan ≈ 11 m); vans max excursion 0.03 m; the 20-car parity harness improved to 0 invasions in 44 turns (was 3 marginal) with straight-road steer_std bit-identical (0.0452 vs 0.0453).

One honest limit worth recording: at the tightest R≈11 m junctions a 2.9 m-wide, 5.15 m-wheelbase truck still sweeps its body corner up to ~0.76 m over the lane line at the apex. That is the rigid-body geometric floor — the chord `wb²/8R` alone consumes the vehicle's entire 0.30 m lane margin — not a control error; a human driver would need a swing-out maneuver the TM doesn't attempt.

TM is client-side, so both fixes ship in the PythonAPI wheel — no server rebuild needed.

hakuturu583:
@germanros1987 I found this PR today.
I’m currently working on [porting the Chrono integration to UE5.](https://github.com/carla-simulator/carla/pull/9861)
Since the original code is written in UE4, I assume it’s best to continue working on the UE5-dev branch, but if it would be better to merge into this PR, I’ll close my PR and merge into this one instead.

youtalk (commented):
I'm really looking forward to the update to Unreal Engine 5.8, as well as the various feature expansions and additions, and I can't wait for this branch to be merged into the `ue5-dev` branch.

I was surprised to see the commit history this time and found that native Autoware support has been integrated. This was actually the real reason I was working on the `ue4-dev` to `ue5-dev` porting https://github.com/carla-simulator/carla/issues/9635, and I'm very glad to see it was accomplished by you. As a member of the @autowarefoundation, I'm thrilled to be here.

I reviewed the integration with ROS 2 and Autoware in particular, and created two PRs to address the following items. 
@germanros1987 @Blyron @JArmandoAnaya Please take a look and merge them into this branch if everything looks good.

- https://github.com/carla-simulator/carla/pull/9873
- https://github.com/carla-simulator/carla/pull/9874

JArmandoAnaya:
So, to clarify, will this branch, for now, be the new target for adding refactoring to the Carla UE5? @germanros1987 @Blyron

Blyron:
> So, to clarify, will this branch, for now, be the new target for adding refactoring to the Carla UE5? @germanros1987 @Blyron

Until it is merged into ue5-dev, yes. We are expecting to merge it asap. Expect this week or next one. We are gonna make 5.8 our standard version

JArmandoAnaya:
@Blyron, Ok, got it. Thank you.

PR #9836, opened 2026-08-10
fix(nav): defer retries for non-traversable walker routes

### LINKED ISSUE #9103: WalkerManager::SetWalkerRout  infinite recursion
**Setup**
  * CARLA version: 0.9.14
  * Platform: ...
  * Python version: 3.8
  * GPU: ...
  * GPU Drivers: ...

**Describe the bug**
I run walker in town05, sometime may cause client stackoverflow ,so I add log in my program ,it stop at walker_ai_control.go_to_location; so i check the Libcarla and add some log in cpp in  WalkerManager::SetWalkerRoute,that call same id many times ,WalkerManager::SetWalkerRoute call WalkerManager::SetWalkerNextPoint, when info.currentIndex > info.route.size() WalkerManager::SetWalkerNextPoint call  WalkerManager::SetWalkerRoute  that may cause   infinite recursion

**Steps to reproduce**
Explain the steps needed to reproduce the issue.

**Expected behavior**

**Logs**
Paste the server and client logs. Server ones can be found at:
  * Unreal Engine 4: _Unreal/CarlaUE4/Saved/Logs_
  * Unreal Engine 5: _Unreal/CarlaUnreal/Saved/Logs_

**Scripts**
Add any scripts relevant for the reproducibility of the error.

**Screenshots**

**Additional context**

tianrking:
#### Description

`WalkerManager::SetWalkerNextPoint` immediately requested a new random route when a route was exhausted. `SetWalkerRoute` then called back into `SetWalkerNextPoint` even when `GetAgentRoute` failed or produced only the walker's current point. If route generation kept returning a failed, empty, or single-point path, the two methods recursively called each other until the client stack overflowed.

This change:

- checks failures from `GetRandomLocation` and `GetAgentRoute`;
- treats fewer than two filtered route points as non-traversable, since the first point is the walker's current position;
- keeps the walker paused and retries route generation from the next navigation update instead of the same call stack;
- preserves the immediate path for valid routes; and
- extends the walker navigation smoke coverage with a current-position destination before continuing to a valid destination.

Fixes #9103

- **Platform(s):** Windows 11 host, WSL2 Ubuntu
- **Python version(s):** 3.x syntax validation and Ruff linting
- **Unreal Engine version(s):** Full UE5 server smoke test not run locally because the packaged server/content is not available; the regression is added to the existing packaged-server smoke suite for CI.

Validation performed:

- Compiled the actual modified `WalkerManager.cpp` with GCC 15 using `-std=c++17 -Wall -Wextra -Werror` against a controlled `Navigation` boundary.
- Reproduction guard observed 65 same-stack route-generation calls on the previous code for a single-point path.
- After the change, failed and single-point paths each make one attempt per update and remain paused; a two-point route still unpauses and dispatches its direct target immediately.
- `python -m py_compile PythonAPI/test/smoke/test_walker_navigation.py`
- `python -m ruff check PythonAPI/test/smoke/test_walker_navigation.py`
- `git diff --check`

#### Possible Drawbacks

A completed or non-traversable route now receives its random replacement on the next navigation update rather than recursively in the same call. With the smoke suite's 0.05-second fixed step this is at most one simulation tick, and it bounds both stack growth and retry work per tick.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9836)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## LATEST Changes
 
+* Fixed walker navigation repeatedly rebuilding failed, empty, or single-point routes on the same call stack, which could overflow the client stack when route generation remained unable to find a traversable path.
 * Added the `CARLA_MAPS_TO_COOK` CMake configure option to select which maps the `package` targets cook. Set it to a `+`-separated list of package paths (e.g. `/Game/Carla/Maps/Town10HD_Opt`) to produce a smaller package with only those maps; leave it empty (the default) to cook the full map list from `DefaultGame.ini`.
 * Fixed `from carla.command import ...` on the Python API by registering the `command` submodule under `carla.command` in `sys.modules` (the ue5 wheel ships `carla` as a bare extension module, so the previous `libcarla.command` namespace was not importable).
 * Modernized the Python navigation agents with type hints and code cleanup ported from `ue4-dev`, added typed obstacle/traffic-light detection-result tuples in `agents/tools/hints.py`, reworked `BasicAgent.set_destination` to accept an optional start location and queue-clean flag, and fixed obstacle detection treating an empty `vehicle_list` as "all vehicles".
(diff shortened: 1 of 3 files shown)


Blyron:
Please, do against ue58-dev

PR #9849, opened 2026-08-24
Fix CarlaSetup.sh to use prerequisite-installed CMake

### LINKED ISSUE #9843: CarlaSetup.sh uses system CMake after installing required CMake 3.28.3
**Setup**

- CARLA version: `ue58-dev` at `d75d3b0f7f638247c915beb6615ee583a35909fd`
- Platform: Ubuntu 22.04.5 LTS
- Python version: 3.10.12
- Preinstalled CMake: 3.22.1 (`/usr/bin/cmake`)
- GPU / driver: not relevant; failure occurs during CMake configure

**Describe the bug**

On a host whose system CMake is older than the required version, `CarlaSetup.sh` successfully installs CMake 3.28.3 under `/opt/cmake-3.28.3-linux-x86_64`, but later invokes the old system `cmake` and fails configuration.

`CarlaSetup.sh` runs the prerequisite installer as a child process:

```bash
bash -x Util/SetupUtils/InstallPrerequisites.sh --python-path=$python_path
```

`InstallPrerequisites.sh` adds the new CMake directory to `PATH` using `export`, but that environment change only exists in the child shell. When control returns to `CarlaSetup.sh`, its later bare `cmake` invocation still resolves to `/usr/bin/cmake` 3.22.1.

**Steps to reproduce**

1. Use Ubuntu 22.04 with the distribution CMake 3.22.1 and no CMake under `/opt/cmake-3.28.3-linux-x86_64`.
2. Check out the `ue58-dev` branch.
3. Provide a built engine through `CARLA_UNREAL_ENGINE_PATH`.
4. Run:

```bash
CARLA_UNREAL_ENGINE_PATH=/path/to/UnrealEngine ./CarlaSetup.sh --interactive
```

5. Observe that the prerequisite installer downloads and installs CMake 3.28.3, but the configure step uses CMake 3.22.1.

**Expected behavior**

The same `CarlaSetup.sh` invocation should immediately use the CMake version it installed and continue configuring CARLA without requiring a new shell or manually changing `PATH`.

**Logs**

```text
Could not find CMake >=3.28.0.
Installing CMake 3.28.3...
Installed CMake 3.28.3.
Found CARLA content.
DLSS SDK ready at /home/user/SDKs/DLSS
Found CARLA Unreal Engine at /path/to/UnrealEngine
Configuring the CARLA CMake project...
CMake Error at CMakeLists.txt:16 (cmake_minimum_required):
  CMake 3.27.2 or higher is required. You are running version 3.22.1
```

**Workaround**

Invoke the installed executable explicitly for configure and build:

```bash
/opt/cmake-3.28.3-linux-x86_64/bin/cmake -G Ninja -S . -B Build ...
/opt/cmake-3.28.3-linux-x86_64/bin/cmake --build Build
```

**Suggested fix**

After the prerequisite child process returns, either prepend `/opt/cmake-3.28.3-linux-x86_64/bin` to `PATH` in `CarlaSetup.sh`, or resolve and retain a `CMAKE_EXECUTABLE` path and use it for the configure/build/install commands. This also handles the case where the required `/opt` installation already exists but is not in the parent shell's `PATH`.

tandede:
#### Description

`CarlaSetup.sh` runs `InstallPrerequisites.sh` as a child process. Although that script installs CMake 3.28.3 and updates its own `PATH`, the environment change does not propagate back to the parent setup process. On Ubuntu 22.04, subsequent configuration steps can therefore continue using the system CMake 3.22.1 and fail the project's 3.28.0 minimum-version requirement.

This change checks the CMake visible to the parent process after the prerequisite step completes. If CMake is missing or older than 3.28.0, and the managed CMake 3.28.3 executable exists, its directory is prepended to the current setup process's `PATH`.

A qualifying parent version is preserved, so an existing newer CMake such as 3.30.1 is not replaced by 3.28.3. The `--skip-prerequisites` path is unchanged.

Fixes #9843.

* **Platform(s):** macOS arm64 host with an isolated Bash path-resolution regression harness for the affected Ubuntu setup logic
* **Python version(s):** N/A
* **Unreal Engine version(s):** N/A

Validation performed:

- `bash -n CarlaSetup.sh`
- `git diff --check`
- verified that a parent process exposing CMake 3.22.1 switches to the managed CMake
- verified that a parent process without CMake switches to the managed CMake
- verified that a parent process exposing CMake 3.30.1 keeps its existing `PATH`
- verified that the selected CMake reaches the configure, build, and Python API installation entry points

The full CARLA/Unreal build was not run because this machine does not provide a Linux container or an Unreal Engine 5 build environment.

#### Possible Drawbacks

The handoff mirrors the managed CMake path and minimum version defined by `InstallPrerequisites.sh`. If those constants change in the installer, this parent-process handoff will need to be updated as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9849)


Blyron:
Please, target ue58-dev

tandede:
> Contributor

Thanks for the heads-up. I’ve retargeted the PR to `ue58-dev` and rebased the change onto the latest branch head. It is now cleanly mergeable.

Issue #9160, opened 2025-08-15
Chrono support for CARLA with Unreal Engine 5.5


nallen94:
When will the support for Chrono be added to CARLA that runs with Unreal Engine 5.5 ?

Blyron:
Please fulfill the proper ticket template

PR #9861, opened 2026-09-02
Port the Project Chrono integration from ue4-dev

hakuturu583:
## What

Ports the Project Chrono vehicle-dynamics integration from `ue4-dev` to `ue5-dev`. Asked for in #9160 (closed for a missing template rather than on the merits).

`ue5-dev` already carries `UChronoMovementComponent`, the `enable_chrono_physics` RPC and its Python binding, but none of it is reachable today:

- `Carla.Build.cs` throws `NotImplementedException` on the Chrono path
- nothing builds or links the Chrono library
- `WITH_CHRONO` is never defined, so the component body is compiled out

This wires the feature up and restores the pieces that were lost on the way over from `ue4-dev`.

## Build integration

- New `Chrono/` subproject builds Project Chrono 6.0.0 with `ExternalProject_Add`, mirroring `Ros2Native/`. It is pinned to C++14 — Chrono 6 predates C++17 and does not compile at CARLA's C++20 level — and uses the Unreal clang/libc++ toolchain, so its headers still include cleanly from the C++20 Carla module.
- Its shared objects are staged into the Carla plugin's `Binaries/Linux` and registered by `Carla.Build.cs`. `Unreal/CMakeLists.txt` propagates the `CHRONO` entry through `Options.def` and puts the Chrono headers on `Includes.def`.
- Enable with `-DENABLE_CHRONO=ON`. `-DCARLA_CHRONO_SOURCE_PATH=<PATH>` builds a local checkout instead of fetching (replacing the old `--chrono-path` flag), and `-DCARLA_CHRONO_TAG` selects the upstream revision.

```sh
cmake --preset Release -DENABLE_CHRONO=ON
cmake --build Build/Release --target launch
```

## restore_physx_physics

`enable_chrono_physics` survived the port to `ue5-dev` but its counterpart did not, leaving no way back to the default physics once Chrono is enabled. The `UBaseCarlaMovementComponent::DisableSpecialPhysics` hook it relies on was dropped too. Restored across the stack:

| Layer | Added |
|---|---|
| Movement component | `DisableSpecialPhysics()` virtual + Chrono override |
| `FVehicleActor` | `RestorePhysXPhysics()` |
| RPC server | `BIND_SYNC(restore_physx_physics)` |
| LibCarla client | `Vehicle` / `Client` / `Simulator` |
| Python API | `restore_physx_physics` binding, `actor.yml`, `python_api.md` |

## Collision warning moved back to the collision path

`ue5-dev` had moved the "Chrono physics does not support collisions yet, reverting to default PhysX
physics" warning out of `OnVehicleHit` / `OnVehicleOverlap` and into `DisableChronoPhysics()`. That was
harmless while the collision handlers were the only callers, but `restore_physx_physics` calls the same
function deliberately, so switching back to the default physics on purpose reported a collision that
never happened. The warning is back on the two collision handlers, where `ue4-dev` has it, leaving
`DisableChronoPhysics()` neutral for both callers. Behaviour is otherwise unchanged.

## Sedan templates: JSON tracked, meshes staged

`ue4-dev` vendors Chrono's whole `data/vehicle/sedan` directory under `Co-Simulation/`. Comparing it against what Chrono itself installs, every file is byte-identical except `Sedan_Chassis.json`:

```
"Centroidal Frame": { "Location": [0, 0, 0.7] }   ->   [0, 0, 0.52]
```

One tuned number, carrying ~15 MB of meshes, POV-Ray exports and Blender sources with it — and `-DENABLE_CHRONO=ON` already installs an identical copy of all of it to `Build/Release/Chrono/install/share/chrono/data/vehicle/sedan/`.

So this PR tracks only the JSON templates (which hold CARLA's tuning) and has the `Chrono/` subproject copy the three referenced meshes and their `.mtl` files out of the Chrono install tree, gitignored. The documented `Co-Simulation/Chrono/Vehicles/sedan/` path and the tuned centroid both still work; the repository just stops carrying a second copy of upstream data. Verified from a clean tree with the meshes deleted.

Happy to vendor the full directory instead if you would rather the path work without configuring Chrono first.

## Also ported

- `Docs/tuto_G_chrono.md`, with the build instructions rewritten for the CMake flow, plus an `mkdocs.yml` nav entry.
- `PythonAPI/examples/manual_control_chrono.py`. `ue4-dev`'s copy is a fork of an older `manual_control.py` that had drifted 458 lines from `ue5-dev`'s, so this is the current `ue5-dev` example plus the CTRL+O Chrono toggle (+23/-1 lines over 5 hunks) instead of the stale file.

## Verification

https://github.com/user-attachments/assets/e0d11bec-945b-400d-8cc7-2053cfae112f

Ubuntu 22.04, `CarlaUnreal/UnrealEngine` @ `ue5-dev-carla` (UE 5.5), bundled clang 18.1.0 + libc++.

With `-DENABLE_CHRONO=ON`:

- `carla-chrono`, `carla-unreal-editor` and `carla-python-api-install` all build clean
- UBT reports `Chrono support is enabled.`
- `libUnrealEditor-Carla.so` lists `libChronoEngine.so` and `libChronoEngine_vehicle.so` as `NEEDED`, exports 32 `UChronoMovementComponent` symbols and references 19 `chrono::` symbols
- `FVehicleActor::RestorePhysXPhysics`, `UChronoMovementComponent::DisableSpecialPhysics` and the `restore_physx_physics` RPC name are all present in the built plugin
- `carla.Vehicle` exposes both `enable_chrono_physics` and `restore_physx_physics`
- the sedan meshes are staged correctly starting from a tree with them deleted

Runtime behaviour is covered in the verification comments below. Driven from a client script, the vehicle covers 17.29 m under Chrono without tripping the collision fallback, then keeps driving after `restore_physx_physics()`. `PythonAPI/examples/manual_control_chrono.py` was then run unmodified and driven with real key events, toggling Chrono on and off with CTRL+O.

A default build (`ENABLE_CHRONO` off, its default) was also verified to be unaffected.

## Not covered

- Only the Linux build is wired up, matching the current state of the UE5 branch. The `Carla.Build.cs` block stages from `Binaries/Linux`, as the ROS2 block does.
- Driven end to end against a headless server, both from a client script and interactively through `manual_control_chrono.py` — see the verification comments below. Not tested on a packaged (non-editor) build.
- Collision support is unchanged from `ue4-dev`: a collision still reverts the vehicle to default physics. Only the placement of the warning changed, as described above.
- The integration stays effectively sedan-only. `UChronoMovementComponent` applies a fixed `(0, 0, -0.25)` position offset and a 2.5 degree pitch correction that are tuned for the one shipped model, so Chrono's other vehicle templates (`citybus`, `longhaul`, `mtv`, `hmmwv`, …) render misaligned. Filed separately as #9862, together with two adjacent hardcodings in the same file — terrain friction pinned to 1.0, and the collision fallback being invisible to the client. All three predate this PR; none are introduced by it, and none are needed to make the existing integration buildable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013bSfUeZdAm96gauHK5HJct

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9861)


hakuturu583:
## Runtime verification

The PR description previously said the change was only verified at the build/link/API level. It has now been exercised against a running server.

**Setup** — Ubuntu 22.04, RTX 3090, `CarlaUnreal/UnrealEngine` @ `ue5-dev-carla` (UE 5.5), server built with `-DENABLE_CHRONO=ON` and launched headless:

```sh
UnrealEditor CarlaUnreal.uproject -game -vulkan -RenderOffScreen -nosound -prefernvidia
```

Client: synchronous mode, `fixed_delta_seconds=0.02`, `vehicle.lincoln.mkz` on `Town10HD_Opt`, full throttle held for 150 ticks (3 s) in each phase. The vehicle is allowed to settle on the ground for 60 ticks before physics is switched, so the collision fallback is not tripped by the spawn overlap.

### Under Chrono physics

Using the templates this PR ports, at `Co-Simulation/Chrono/Vehicles/`:

```
enable_chrono_physics(5000, 0.002, Sedan_Vehicle / SimpleMapPowertrain / TMeasyTire)
    t  0  speed   0.19 km/h   loc (  -24.35,   -57.78,  -0.09)
    t 25  speed   7.25 km/h   loc (  -23.88,   -57.78,  -0.08)
    t 50  speed  15.47 km/h   loc (  -22.25,   -57.76,  -0.09)
    t 75  speed  22.33 km/h   loc (  -19.65,   -57.74,  -0.10)
    t100  speed  28.19 km/h   loc (  -16.09,   -57.70,  -0.08)
    t125  speed  33.11 km/h   loc (  -11.84,   -57.67,  -0.06)
    t149  speed  38.55 km/h   loc (   -7.07,   -57.64,  -0.06)
  travelled 17.29 m, final speed 38.55 km/h
  collision fallback triggered: False
```

Server log for that phase, confirming the ported templates are what got loaded:

```
LogCarla: Loading Chrono files: Vehicle: .../Co-Simulation/Chrono/Vehicles/sedan/vehicle/Sedan_Vehicle.json,
          PowerTrain: .../sedan/powertrain/Sedan_SimpleMapPowertrain.json,
          Tire: .../sedan/tire/Sedan_TMeasyTire.json
Loaded JSON: .../Co-Simulation/Chrono/Vehicles/sedan/vehicle/Sedan_Vehicle.json
```

Z stays flat at about -0.06 to -0.10 m throughout, so the vehicle is tracking the road surface rather than sinking or being teleported, and no `reverting to default PhysX physics` warning is emitted.

### After `restore_physx_physics()`

```
    t  0  speed  41.89 km/h   loc (   -4.64,   -57.62,  -0.05)
    t 25  speed  55.85 km/h   loc (    2.14,   -57.57,  -0.05)
    t 50  speed  66.41 km/h   loc (   10.83,   -57.50,  -0.07)
    t 75  speed  66.87 km/h   loc (   20.06,   -57.43,  -0.05)
    t100  speed  69.42 km/h   loc (   29.53,   -57.36,  -0.00)
    t125  speed  71.94 km/h   loc (   39.32,   -57.28,  -0.06)
    t149  speed  75.14 km/h   loc (   49.13,   -57.21,  -0.07)
  travelled 54.01 m, final speed 75.14 km/h
```

The vehicle keeps driving on the road under the default Chaos physics, with a clearly different acceleration profile (75 km/h vs 38 km/h over the same 3 s at full throttle) — the two solvers are visibly doing the work in their respective phases.

### Camera frame under Chrono

vehicle driving under Chrono physics in Town10HD_Opt

The dark speckling on the road is a shadow artefact of headless `-RenderOffScreen` rendering on this box and is unrelated to this change.

### Reproducing

The log, the frame and the script are on a side branch:
<https://github.com/hakuturu583/carla/tree/chrono-runtime-evidence>

```sh
cmake --preset Release -DENABLE_CHRONO=ON
cmake --build Build/Release --target launch
python verify_chrono.py --templates <repo>/Co-Simulation/Chrono/Vehicles/ --server-log <server log>
```

### One note for reviewers

Enabling Chrono on a vehicle that has just been spawned trips the existing collision fallback immediately — the spawn overlap counts as a collision, the component reverts to default physics, and the caller gets no error, only a server-side warning. That behaviour is inherited from `ue4-dev` and is not changed here, but it makes `enable_chrono_physics` look like a no-op if you call it right after `spawn_actor`. Letting the vehicle settle for a few ticks first avoids it. Worth a follow-up if you would like the fallback to be visible to the client.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013bSfUeZdAm96gauHK5HJct

hakuturu583:
## Interactive verification: `manual_control_chrono.py`

The description previously listed the example as not yet exercised interactively. It has now been run
unmodified and driven with real key events, and the description has been updated.

**Setup** — Ubuntu 22.04, RTX 3090, `CarlaUnreal/UnrealEngine` @ `ue5-dev-carla` (UE 5.5), server built
with `-DENABLE_CHRONO=ON` and launched headless. The example itself runs untouched on an `Xvfb :99`
display at 1280x720, with keystrokes injected by `xdotool`, on `Town10HD_Opt` with `vehicle.lincoln.mkz`.

### File provenance

The committed example is the current `ue5-dev` `manual_control.py` plus the Chrono delta, reproduced
from scratch:

```
manual_control.py (ue5-dev)   sha256 669e1156ea05f5c22b35b4147524d1d0c365ca29832685a50ef660507868e4ff
  + Chrono delta (+23/-1 lines, 5 hunks)
  = rebuilt                    sha256 2acc301bd7a4bb8809454b49d80baa53280a69f6ad4b180dfa5eeaf459e2abfb
    committed file             sha256 2acc301bd7a4bb8809454b49d80baa53280a69f6ad4b180dfa5eeaf459e2abfb  -> IDENTICAL
```

### Driven session

```
  CHRONO HELD - recording this session
  >>> throttle 6s under Chrono
  still under Chrono after the drive
  >>> CTRL+O : restore default physics
  >>> throttle 6s under default
```

CTRL+O reaches the server and loads the templates this PR ports:

```
LogCarla: Loading Chrono files: Vehicle: /mnt/sata_ssd/carla/Co-Simulation/Chrono/Vehicles/sedan/vehicle/Sedan_Vehicle.json,
          PowerTrain: .../sedan/powertrain/Sedan_SimpleMapPowertrain.json,
          Tire: .../sedan/tire/Sedan_TMeasyTire.json
Loaded JSON: /mnt/sata_ssd/carla/Co-Simulation/Chrono/Vehicles/sedan/vehicle/Sedan_Vehicle.json
```

The car drives under Chrono for the whole 6 s without the collision fallback firing, then a second
CTRL+O calls `restore_physx_physics()` and it keeps driving under the default physics.

### A bug this found

Toggling Chrono **off** cleanly still logged `Chrono physics does not support collisions yet, reverting
to default PhysX physics.` every time. `ue5-dev` had moved that warning from `OnVehicleHit` /
`OnVehicleOverlap` into `DisableChronoPhysics()`, which was harmless until `restore_physx_physics` began
calling the same function on purpose. Fixed in 4e128ee3f by putting the warning back on the two
collision handlers, as `ue4-dev` has it; the run above emits it zero times after the final enable.

### Notes on the harness, for anyone reproducing this

Three things bite when synthesising input for this example, none of them a problem with the example itself:

1. `xdotool search --name` finds no window — SDL2 sets `_NET_WM_NAME` but not the legacy `WM_NAME`
   that `xdotool` matches on. Enumerate windows and match `getwindowname` instead.
2. `xdotool key ctrl+o` does not toggle — the example handles keys on `KEYUP` and reads the modifier
   with `pygame.key.get_mods()`, which reports the state at poll time, not at event time. A synthetic
   chord releases CTRL before the next poll, so CTRL has to be held across the tap.
3. Enabling Chrono right after spawn trips the pre-existing collision fallback at many spawn points
   (#9862). The driver retries until it gets a spawn where Chrono holds — 4 attempts here.

### Reproducing

The recording, the log and the driver script are on the evidence branch:
<https://github.com/hakuturu583/carla/tree/chrono-runtime-evidence>

- [`manual_control_chrono.mp4`](https://github.com/hakuturu583/carla/raw/chrono-runtime-evidence/manual_control_chrono.mp4) — 20 s of the session above
- [`manual_control_chrono_log.txt`](https://github.com/hakuturu583/carla/blob/chrono-runtime-evidence/manual_control_chrono_log.txt) — checksums plus the client and server logs
- [`drive_manual_control.sh`](https://github.com/hakuturu583/carla/blob/chrono-runtime-evidence/drive_manual_control.sh) — the harness

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HaWof6oGAVQLZMjsZU27cx

PR #9885, opened 2026-09-15
Fix stale PythonAPI unit test assertions

JArmandoAnaya:
#### Description

`PythonAPI/test/unit` had two stale assertions failing independent of platform or Python version.

`test_client.TestClient.test_client_version` compared the client's fixed `CARLA_VERSION` against `git describe --tags --dirty --always`, which includes a commit-count/hash suffix on any commit past the last tag, so it only ever passed sitting exactly on a release tag. Switched to `--abbrev=0` to compare against the nearest tag only, matching what `carla::version()` actually returns.

`test_vehicle.TestVehiclePhysicsControl.test_named_args` referenced `VehiclePhysicsControl`/`WheelPhysicsControl` fields from the old PhysX-era vehicle model (`moi`, `use_gear_autobox`, `gear_switch_time`, `tire_friction`, `radius`, `wheels[i].position`) that no longer exist on the current Chaos-based struct. Updated the test to construct and assert against the current field names (`rev_up_moi`, `use_automatic_gears`, `gear_change_time`, `cornering_stiffness`, `wheel_radius`, `location`), keeping the original intent: named-arg construction plus a property round-trip check.

#### Changes

* `test_client.py`: use `git describe --tags --abbrev=0` instead of `--dirty --always`.
* `test_vehicle.py`: update `TestVehiclePhysicsControl.test_named_args` to the current `VehiclePhysicsControl`/`WheelPhysicsControl` field names.

  * **Platform(s):** Linux (Ubuntu 26.04)
  * **Python version(s):** 3.10
  * **Unreal Engine version(s):** 5.8

`PythonAPI/test/unit` now passes 51/51 (previously 49/51).

#### Possible Drawbacks

None; test-only change, no production code touched.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9885)


JArmandoAnaya:
Hello folks. This PR fixes some broken tests that I identified after compiling the last changes on my computer.

Blyron (approved):

Issue #9890, opened 2026-09-21
Autoware integration: traffic lights are not detected correctly, stop signs are ignored


JArmandoAnaya:
**Setup**
  * CARLA version: 0.10.0 (UE5.8, ue58-dev branch)
  * Platform: Linux
  * Python version: 3.12
  * GPU: not relevant to this issue
  * GPU Drivers: not relevant to this issue

**Describe the bug**
When running the CARLA and Autoware integration (the classical stack demo under PythonAPI/examples/av_stacks/autoware), the ego vehicle does not react to traffic lights the way it should.

The Lanelet2 map generated for Autoware (converted from CARLA's OpenDRIVE data) does not carry reliable traffic light information. In practice the vehicle ends up treating a light as always red, always green, or not reacting to it at all, instead of following what the light is actually doing in the simulation at that moment.

On top of that, the vehicle also drives straight through every stop sign, since stop signs are not represented in the generated map at all.

**Steps to reproduce**
  1. Run the CARLA-Autoware classical mode demo (run_carla_autoware.sh) in Town10HD.
  2. Approach any signalized intersection or a stop sign.
  3. The car stops or goes with no clear relation to the actual light state, and it never stops at a stop sign.

**Expected behavior**
The vehicle should follow the real state of each traffic light as it changes in the simulation, and stop at every stop sign, the same way it is expected to behave at a real traffic light or sign.

**Additional context**
Fixed in PR https://github.com/carla-simulator/carla/pull/9891, which closes this issue. The work ended up covering two separate scopes.

*Scope 1: the reported traffic light and stop sign bug*

  * The OpenDRIVE to Lanelet2 conversion now associates each traffic light with the correct lanes, including junctions where a lane splits after the signal, so a light no longer ends up attached to the wrong lane or to none at all.
  * Map generation is now deterministic, so the same town always produces the same map instead of shifting signal associations between runs.
  * Stop signs are now emitted as stop sign regulatory elements, covering lane splits and stop signs placed near crosswalks.
  * A new bridge publishes CARLA's own ground-truth traffic light state straight to Autoware. Previously the stack depended on Autoware's camera-based recognition of CARLA's rendered lights, which is the root cause of the light being perceived as always red, always green, or not there at all.
  * Town10HD's Lanelet2 map was regenerated with the fixed conversion logic.

*Scope 2: ghost moving cars found while validating the fix*

[...]

PR #9891, opened 2026-09-21
Fix traffic light detection in the Autoware integration, add stop sign support

JArmandoAnaya:
#### Description

This PR covers two separate pieces of work: traffic light and stop sign support in the CARLA-Autoware integration, and two native CARLA sensor bugs found while validating that integration live.

**1. Traffic light and stop sign support in the CARLA-Autoware integration**

The OpenDRIVE-to-Lanelet2 conversion now associates traffic lights with the appropriate lanes more reliably, including junctions where a lane splits after the signal, and map generation is deterministic across repeated runs because stale intermediate state is cleared between conversions. The conversion also emits stop sign regulatory elements into the generated Lanelet2 map, covering lane splits and stop signs near crosswalks. To support both, the stop-waypoint and OpenDRIVE-ID accessors moved from the traffic light implementation up to the common traffic sign base class. A new bridge publishes CARLA's ground-truth traffic light state to Autoware directly, instead of relying on camera-based recognition of the rendered lights. Town10HD has been regenerated with the updated conversion logic.

**2. Native sensor timing and lidar configuration fixes**

Live testing on a long autonomous route caught the ego halting near objects that never actually moved, with Autoware's perception briefly classifying a "ghost" car, a track with no corresponding CARLA actor, as moving. Two native CARLA bugs compounded to cause it.

First, `ASensor::PostPhysTickInternal` fed sensors the current physics substep's delta time instead of the accumulated tick-interval delta UE5's own scheduler already computes, so any sensor whose `sensor_tick` spans more than one physics substep (lidar, radar, IMU, all cameras) simulated as if only the last substep had elapsed. For this integration's lidar (`sensor_tick=0.1` in a `fixed_delta_seconds=0.05` world) that halved the swept angle per publish, so consecutive scans alternated between two disjoint half-sweeps instead of one full rotation.

Second, `autoware_demo.py`'s lidar blueprint never set `rotation_frequency` or the dropout attributes, so CARLA's defaults applied: 60 Hz rotation, with 45% of points randomly dropped every tick. The lidar is now explicitly set to 10 Hz with dropout disabled. **The 10 Hz rotation frequency was chosen specifically to stop Autoware from perceiving ghost cars moving, and it is the scan rate and cadence that Autoware's `carla_sensor_kit` lidar pipeline is actually calibrated to expect. It is not an arbitrary tuning value.** Before the fix, the combined jitter on completely static geometry was enough for Autoware's tracker to read as motion.

Closes #9890

#### Changes

Traffic light and stop sign integration:

* Traffic light lanelet association, stop-line orientation, deterministic map generation, stale-state cleanup (`generate_lanelet2_map.py`)
* Stop sign regulatory elements in the generated Lanelet2 map, with new unit tests
* Hoisted `GetStopWaypoints()`/`GetAffectedLaneWaypoints()` from `TrafficLight` to `TrafficSign` (`LibCarla`, `PythonAPI` bindings)
* CARLA-to-Autoware ground-truth traffic light state bridge (`run_carla_autoware.sh`, two new run scripts)
* Regenerated Town10HD with the updated conversion logic

Sensor fixes:

* Fixed `ASensor::PostPhysTickInternal` using the physics substep delta instead of the correct tick-interval delta (`Sensor.h`/`Sensor.cpp`); affects every native sensor whose `sensor_tick` spans multiple physics substeps
* Set the demo lidar's `rotation_frequency` to 10 Hz and disabled point dropout (`autoware_demo.py`), matching the scan rate Autoware's sensor kit expects and eliminating the frame-to-frame jitter that was causing Autoware to perceive ghost cars moving
* Fixed a missing include in `World.cpp` (`client::Map`/`road::Controller`), a pre-existing transitive include through `TrafficLight.h` that the `TrafficSign` hoist above removed

* **Platform(s):** Ubuntu 26.04
* **Python version(s):** 3.12
* **Unreal Engine version(s):** 5.8

#### Possible Drawbacks

Traffic light and stop sign integration:

* Maps generated before this change need to be regenerated to include the updated traffic light and stop sign information.
* The traffic light ground-truth bridge must run alongside the CARLA-Autoware integration for Autoware to receive the simulated traffic light state.

Sensor fixes:

* The `PostPhysTick` timing fix changes the delta time every native sensor receives whenever its `sensor_tick` spans more than one physics substep, not just the lidar case that surfaced it. Any sensor implicitly relying on the old, incorrect substep-sized delta should be re-checked.

- - -
This change is [Reviewable](https://reviewable.io/reviews/carla-simulator/carla/9891)


JArmandoAnaya:
This PR can help improve the Carla-Autoware integration by fixing some issues encountered when testing traffic light detection with Autoware. @youtalk, it would be great if you have time to take a look and give some feedback about this refactoring.

Aperiss:
Live map generation crashes on towns with stop signs that have no OpenDRIVE id. On `Town03_Opt`:

```
File ".../generate_lanelet2_map.py", line 768, in <lambda>
    key=lambda s: int(s.get_opendrive_id()))
ValueError: invalid literal for int() with base 10: ''
```

`get_opendrive_id()` returns `''` for stop signs in several shipped towns:

| Map | Stop signs with empty id |
|---|---|
| Town03_Opt | 5 |
| Town05_Opt | 5 |
| Town06_Opt | 1 |
| Sf_Soma | 66 |
| Sunnyvale | 139 |

`_check_unique_opendrive_id` has the same problem for any traffic light or stop sign without a numeric id.

  **Suggestion:** skip id-less signs with a warning, or allocate them ids.

Aperiss:
The Sensor.cpp and Sensor.h fix is correct. On `ue58-dev` the demo lidar publishes alternating 180° half sweeps. The same wrong delta also feeds the IMU accelerometer and the radar whenever `sensor_tick` spans more than one frame.

With this PR, every sensor matches ground truth across 5 tick configurations: `sensor_tick` of 0.1, 0, and 0.15 at `dt=0.05`; a non-multiple `dt=1/30`; and substepping.

| Sensor | Result on this PR |
|---|---|
| Lidar / semantic lidar | full 360° per scan; points steady at 0.92 of budget |
| Radar | static-return speed 0.99–1.00× ego speed |
| IMU | forward accel 1.98–1.99 m/s² vs 2.0 commanded; z ≈ 9.83 |
| Camera / all sensors | intervals exactly `sensor_tick` |

- This however changes the output of every native sensor, not just the Autoware demo's. I'd split it into its own PR with a changelog entry, and mention the IMU and radar in it.

Aperiss (changes requested):
Please make the suggested changes to this PR.

Aperiss:
launch fails at startup on any map that has no traffic lights in it. On `ue58-dev` the same launch works; the car just doesn't react to lights.

This affects the prebuilt maps from `fetch_prebuilt_maps.sh` (Town01–07). They carry no traffic-light regulatory elements, and the README still documents them as a supported way to get a map ("Option A — prebuilt", `README.md:147`).

What happens:

1. `carla_traffic_light_state_source.py` reads the map at startup so it can pair each CARLA light with its regulatory element.
2. It finds none and raises, instead of treating "no lights" as "nothing to bridge".
3. `check_alive traffic_light_state_source` in `run_carla_autoware.sh` sees it exit and tears down the whole session: CARLA, Autoware and every other process.

Reproduced with `--town Town01_Opt --map-path <prebuilt Town01>`:

```
RuntimeError: no traffic-light regulatory elements found in .../lanelet2_map.osm
[run_carla_autoware] ERROR: 'traffic_light_state_source' exited immediately
```

**Suggestion:** a map without traffic lights should not be fatal. If the map has no traffic lights, log a warning and skip the bridge, so the stack still launches as before.

JArmandoAnaya:
Thank you @Aperiss, I will do the changes.

JArmandoAnaya:
Hello @Aperiss! I already implemented the requested changes.

